PR #1, opened 2024-02-28
Update to comm/message handling and enumerations.  Example handlings

billphipps:
Update to expand message handling groups and actions.  Changed name of message type to message kind.  Changed message groups to no longer care about upper level kind enumeration and only deal with locally enumerated actions.

wh_comm provides 16-bit "kind" in whCommHeader to specify the type of the packet.  wh_message splits the kind into a "group" (loosely related to the server component doing the work) and "action" (specific action the client is asking the server to perform).

wh_client sends the request (header and data) in native format (no byte swapping) to the server.  wh_server interprets the first header field (magic) to understand if the client is in a different endianness and swaps the header fields to local.  wh_server then invokes the corresponding group handler with the action, data, and size.

The server handler then translates the specific message using custom struct conversion functions (declared in wh_message_XXX.h) if necessary.  The handler performs the requested action to generate a response packet.  The handler translates the response packet if necessary and sends the response back to the client.

The aux field is now defined to be used for future async and session handling, currently not populated.

### DIFF
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -31,6 +31,7 @@
 #include "wolfhsm/nvm_remote.h"
 #endif
 
+#include "wolfhsm/wh_message.h"
 #include "wolfhsm/wh_message_comm.h"
 #include "wolfhsm/wh_client.h"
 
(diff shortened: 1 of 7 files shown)


wolfhsm/wh_comm.h
```diff
@@ -35,9 +30,9 @@
  * DATA_LEN bytes.
  */
 enum {
-    WOLFHSM_COMM_HEADER_LEN = 8,
-    WOLFHSM_COMM_DATA_LEN = 1280,
-    WOLFHSM_COMM_MTU = (WOLFHSM_COMM_HEADER_LEN + WOLFHSM_COMM_DATA_LEN)
+    WM_COMM_HEADER_LEN = 8,    /* whCommHeader */
```
bigbrett, line 33: Is this a typo in the prefix? Shouldn't it be `WH` not `WM`?:
```suggestion
    WH_COMM_HEADER_LEN = 8,    /* whCommHeader */
```
If so, you will need to go through and update all the references

wolfhsm/wh_comm.h
```diff
@@ -58,14 +53,26 @@ enum {
...
+    uint16_t kind;      /* Kind of packet.  Enumerated in message.h */
     uint16_t seq;       /* Sequence number. Incremented on request, copied for
                          * response. */
     uint16_t aux;       /* Session identifier for request or error indicator
                          * for response. */
-} whHeader;
+} whCommHeader;
 /* static_assert(sizeof_whHeader == WOLFHSM_COMM_HEADER_LEN,
-                 "Size of whHeader doesn't match WOLFHSM_COMM_HEADER_LEN") */
+                 "Size of whCommHeader doesn't match WOLFHSM_COMM_HEADER_LEN") */
```
bigbrett, line 63: I'm a big fan of `static_assert`, can we add something like this back in?

wolfhsm/wh_comm.h
```diff
@@ -120,8 +127,8 @@ typedef struct {
     uint16_t reqid;
     uint16_t seq;
     uint16_t size;
-    uint8_t packet[WOLFHSM_COMM_MTU];
-    whHeader* hdr;
+    uint8_t packet[WH_COMM_MTU];
+    whCommHeader* hdr;
     uint8_t* data;
     uint32_t client_id;
     uint32_t server_id;
```
bigbrett, line 134: nitpick, but could we change the struct name to `whCommClientCtx` or `whCommClientContext` to make it explicit what the type is meant to represent (and be consistent with naming of our other structs)?

wolfhsm/wh_comm.h
```diff
@@ -170,8 +177,8 @@ typedef struct {
     void* transport_context;
     const whTransportServerCb* transport_cb;
     uint16_t reqid;
-    uint8_t packet[WOLFHSM_COMM_MTU];
-    whHeader* hdr;
+    uint8_t packet[WH_COMM_MTU];
+    whCommHeader* hdr;
     uint8_t* data;
     uint32_t client_id;
     uint32_t server_id;
```
bigbrett, line 184: same as above, can we make the "context" explicit in the struct name?

src/wh_client.c
```diff
@@ -119,66 +120,117 @@ int wh_Client_Cleanup(whClient* c)
     return 0;
 }
 
-int wh_Client_EchoRequest(whClient* c, uint16_t size, const void* data)
+int wh_Client_SendRequest(whClient* c,
```
bigbrett, line 123: null check on `c` (only `c->comm` is `NULL` checked in `wh_CommClientSendRequest()`)

src/wh_client.c
```diff
@@ -119,66 +120,117 @@ int wh_Client_Cleanup(whClient* c)
...
+                WH_COMM_MAGIC_NATIVE, kind, &req_id,
+                data_size, data);
+    if (rc == 0) {
+        c->last_req_kind = kind;
+        c->last_req_id = req_id;
+    }
+    return rc;
+}
+
+int wh_Client_RecvResponse(whClient *c,
```
bigbrett, line 141: null check all pointer args (`data` is checked in `wh_CommClient_RecvResponse`)

src/wh_client.c
```diff
@@ -119,66 +120,117 @@ int wh_Client_Cleanup(whClient* c)
...
+    whMessageCommLenData msg = {0};
 
     if (    (c == NULL) ||
             ((size > 0) && (data == NULL)) ){
         return WH_ERROR_BADARGS;
     }
 
-    if (size > sizeof(msg.data)) size = sizeof(msg.data);
+    /* Populate the message */
+    if (size > sizeof(msg.data)) {
```
bigbrett, line 187: do we want to automagically trim the size on the TX path? or should it be `WH_ERROR_BADARGS`?

bigbrett (changes requested):
Great additions. A few minor fixes.

wolfhsm/wh_comm.h
billphipps: Sigh.  Fixed in next push.

wolfhsm/wh_comm.h
billphipps: Agreed.  _Static_assert() doesn't get in until C11 and so far I'm only using C99.  I think there's an equivalent in assert.h for GNU builds.  I'll take a peek.

wolfhsm/wh_comm.h
billphipps: Not a nit at all.  I'll correct in next push.

wolfhsm/wh_comm.h
billphipps: fixed

src/wh_client.c
billphipps: fixed.

src/wh_client.c
billphipps: fixed

bigbrett (approved):

PR #2, opened 2024-03-03
test improvements

bigbrett:
This PR contains changes, including:

- Reorganizes tests into module-specific files, with POSIX tests protected by configuration macros. This allows embedded targets to run a subset of the tests on device. 
- Fixes bug in POSIX TCP transport that causes build errors on Linux (`SO_NOSIGPIPE` is macOS/BSD only) by replacing `SO_NOSIGPIPE` functionality with Linux equivalent (use socket::send with `MSG_NOSIGNAL`  instead of write)
- Fixes bug on Linux where server address family is not set, leading to TCP connection failures on linux
- add a .clang-format file (feel free to modify or reject) 

I had originally gone down the rabbit hole of essentially writing a minimal test runner system but abandoned ship once I realized I was reinventing the wheel. I think we keep going with a more modular test setup like this for now, but eventually make the switch to a real unit test framework that gives us runner functionality, as well as other niceties


src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+
+int whFlashRamsim_Cleanup(void* context)
+{
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+    if (context == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    free(ctx->memory);
```
billphipps, line 60:
```suggestion
    if (ctx->memory != NULL) {
        free(ctx->memory);
        ctx->memory = NULL;
    }
```
stdlib free() can't handle a NULL argument.

port/posix/posix_transport_tcp.c
```diff
@@ -21,13 +22,21 @@
 #include "wolfhsm/wh_transport.h"
 #include "port/posix/posix_transport_tcp.h"
 
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__) \
```
billphipps, line 25: Barf.  Lemme see if send(..., MSG_NO_SIGNAL) works on MacOS.  Probably does and I'm fine with using send/recv instead of read/write.  The intent of to support Posix OS's so we don't need to hit BSD directly.  I should have verified Linux handles the sig pipe first.  My bad.  More to come.

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+    return WH_FLASH_RAMSIM_OK;
+}
+
+int whFlashRamsim_Program(void* context, uint32_t offset, uint32_t size,
+                             const uint8_t* data)
+{
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+
+    if (context == NULL || data == NULL) {
```
billphipps, line 71:
```suggestion
    if ((ctx == NULL) ||
        (ctx->memory == NULL) ||
        (ctx->pageSize != 0) ||
        (data == NULL) ) {
```

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+}
+
+int whFlashRamsim_Read(void* context, uint32_t offset, uint32_t size,
+                          uint8_t* data)
+{
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+    if (context == NULL || data == NULL) {
+        return WH_ERROR_BADARGS;
+    }
```
billphipps, line 99:
```suggestion
    if ((ctx == NULL) ||
        (ctx->memory == NULL) ||
        ((offset + size) > ctx->size) ||
        (data == NULL) ) {
```

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+    return WH_FLASH_RAMSIM_OK;
+}
+
+int whFlashRamsim_Erase(void* context, uint32_t offset, uint32_t size)
+{
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+    if (context == NULL) {
+        return WH_ERROR_BADARGS;
+    }
```
billphipps, line 111:
```suggestion
    if ((ctx == NULL) ||
        (ctx->memory == NULL) ||
        ((offset + size) > ctx->size) ) {
```

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+}
+
+int whFlashRamsim_Verify(void* context, uint32_t offset, uint32_t size,
+                            const uint8_t* data)
+{
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+    if (context == NULL || data == NULL) {
+        return WH_ERROR_BADARGS;
+    }
```
billphipps, line 130:
```suggestion
    if ((ctx == NULL) ||
        (ctx->memory == NULL) ||
        ((offset + size) > ctx->size) ||
        (data == NULL) ) {
```

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+}
+
+
+int whFlashRamsim_BlankCheck(void* context, uint32_t offset, uint32_t size)
+{
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+    if (context == NULL) {
+        return WH_ERROR_BADARGS;
+    }
```
billphipps, line 149:
```suggestion
    if ((ctx == NULL) ||
        (ctx->memory == NULL) ||
        ((offset + size) > ctx->size) ||
        (data == NULL) ) {
```

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+    return WH_FLASH_RAMSIM_OK;
+}
+
+
+uint32_t whFlashRamsim_PartitionSize(void* context)
+{
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+    if (context == NULL) {
+        return WH_ERROR_BADARGS;
```
billphipps, line 164:
```suggestion
        /* Invalid context.  Must return positive size, so 0*/
        return 0;
```

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+    if (context == NULL || data == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Ensure offset and size are within bounds and size is a multiple of page
+     * size */
+    if (offset + size > ctx->size || size % ctx->pageSize != 0) {
+        return WH_ERROR_BADARGS;
+    }
+
```
billphipps, line 80:
```suggestion
    if (size > 0 && ctx->writeLocked) {
        return WH_ERROR_LOCKED;
    }
```

billphipps (changes requested):
Looks excellent.  I added a few more checks.  I'll try this on MacOS and recommend to replace the "no sig pipe" with "msg_no_signal" to simplify the code.

billphipps:
The send(..., MSG_NO_SIGNAL) worked fine on MacOS.  I recommend we remove all of the no sig pipe and use send/recv for everything else.  Ok to do to this in a different PR if you want to get this one done first as it works both ways for me.  Much easier checkout and test and the output is far cleaner!

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+    ctx->erasedByte = cfg->erasedByte;
+
+    if (!ctx->memory) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Simulate starting from erased flash */
+    memset(ctx->memory, ctx->erasedByte, ctx->size);
+
+    return WH_FLASH_RAMSIM_OK;
```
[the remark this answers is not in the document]
bigbrett: Whoops, that actually should have been stripped out before I pushed this, I agree. 

> Maybe we would add a success enum WH_NO_ERROR
Introduced `WH_ERROR_OK` as "success"...keeps the `WH_ERROR_xxx` scheme intact.

src/wh_flash_ramsim.c
bigbrett: fixed

src/wh_flash_ramsim.c
bigbrett: fixed

src/wh_flash_ramsim.c
bigbrett: fixed

src/wh_flash_ramsim.c
bigbrett: fixed

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+    whFlashRamsimCtx* ctx = (whFlashRamsimCtx*)context;
+
+    if (context == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    if (offset % ctx->sectorSize != 0 || size % ctx->sectorSize != 0) {
+        return WH_ERROR_BADARGS;
+    }
+
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_flash_ramsim.c
bigbrett: fixed

src/wh_flash_ramsim.c
```diff
@@ -0,0 +1,203 @@
...
+    const whFlashRamsimCfg* cfg = (const whFlashRamsimCfg*)config;
+
+    if (context == NULL || config == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    ctx->size       = cfg->size;
+    ctx->sectorSize = cfg->sectorSize;
+    ctx->pageSize   = cfg->pageSize;
+    ctx->memory     = (uint8_t*)malloc(ctx->size);
```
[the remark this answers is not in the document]
bigbrett: yeah supporting static memory would be a good idea for a subsequent PR

billphipps (approved):
Looks good!

PR #3, opened 2024-03-09
Fix zeroed state variables on disk.

billphipps:
Sets non-zero values in the unused upper bytes of the state variables on the NVM when using nvm_flash subsystem.  This corrects the case when the base NVM uses 0's for Erased AND the EraseVerify/BlankCheck does NOT distinguish, such as the Infineon Tricore.

### DIFF
--- a/src/wh_nvm_flash.c
+++ b/src/wh_nvm_flash.c
@@ -19,7 +19,9 @@ enum {
     NF_COPY_OBJECT_BUFFER_LEN = 8 * WHFU_BYTES_PER_UNIT,
 };
 
-/* MSB's  of state variables */
+/* MSW of state variables (nfState) must be set to this pattern when written
+ * to flash to prevent hardware on certain chipsets from confusing zero values
+ * with erased flash */
 static const whFlashUnit BASE_STATE = 0x1234567800000000ull;
 
 /* On-flash layout of the state of an Object or Directory*/
--- a/src/wh_nvm_flash.c
+++ b/src/wh_nvm_flash.c
@@ -389,7 +389,7 @@ static int nfPartition_ProgramStart(whNvmFlashContext* context,
 static int nfPartition_ProgramCount(whNvmFlashContext* context,
         int partition, uint32_t count)
 {
-    whFlashUnit unit = BASE_STATE + count;
+    whFlashUnit unit = BASE_STATE | count;
 
     if ((context == NULL) || (context->cb == NULL)) {
         return WH_ERROR_BADARGS;
--- a/src/wh_nvm_flash.c
+++ b/src/wh_nvm_flash.c
@@ -89,6 +89,10 @@ static int nfPartition_ProgramStart(whNvmFlashContext* context, int partition,
 static int nfPartition_ProgramCount(whNvmFlashContext* context, int partition,
         uint32_t count);
 static int nfPartition_ProgramInit(whNvmFlashContext* context, int partition);
+static int nfPartition_CheckRangeUnits(whNvmFlashContext* context,
+                                       int partition,
+                                       uint32_t units_offset,
+                                       uint32_t units_count);
 
 static uint32_t nfObject_Offset(whNvmFlashContext* context, int partition,
         int object_index);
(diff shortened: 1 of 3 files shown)


bigbrett (dismissed):
@billphipps I added a few things:

- added the range check on NVM read/program to ensure target data is within the data section of the requested partition. 
- re-ordered read + getMetadata in test

 
Please merge if my changes (since your initial commit) look good to you.

src/wh_nvm_flash.c
```diff
@@ -380,7 +389,7 @@ static int nfPartition_ProgramStart(whNvmFlashContext* context,
 static int nfPartition_ProgramCount(whNvmFlashContext* context,
         int partition, uint32_t count)
 {
-    whFlashUnit unit = count;
+    whFlashUnit unit = BASE_STATE + count;
```
billphipps, line 392:
```suggestion
    whFlashUnit unit = BASE_STATE | count;
```

src/wh_nvm_flash.c
```diff
@@ -440,6 +449,33 @@ static int nfPartition_ProgramInit(whNvmFlashContext* context, int partition)
...
+    if (units_offset < partDataBase) {
+        return WH_ERROR_BADARGS;
+    }
+
+    if (units_offset + units_count > maxUnits) {
+        return WH_ERROR_BADARGS;
+    }
+
+    return WH_ERROR_OK;
+}
```
billphipps, line 477: Hrmm..  I think this function should operate in bytes instead of units.  Imagine trying to handle the edge case of accessing the last byte in the partition.  Also, the name of this function should include Data, since the starting offset is specific to the data portion of the partition.  This change also corrects the math error below.

```suggestion
/*
 * Checks that the range of bytes specified by byte_offset +  byte_count
 * is located inside the specified partition's data area
 */
static int nfPartition_CheckDataRange(whNvmFlashContext* context,
                                       int partition,
                                       uint32_t byte_offset,
                                       uint32_t byte_count)
{
    uint32_t maxOffset;
    uint32_t partDataBase;

    partDataBase = nfPartition_DataOffset(context, partition) * WHFU_BYTES_PER_UNIT;
    maxOffset = (nfPartition_Offset(context, partition)
               + context->partition_units) * WHFU_BYTES_PER_UNIT;

    if (byte_offset < partDataBase) {
        return WH_ERROR_BADARGS;
    }

    if (byte_offset + byte_count > maxOffset) {
        return WH_ERROR_BADARGS;
    }

    return WH_ERROR_OK;
}
```

src/wh_nvm_flash.c
```diff
@@ -86,6 +91,10 @@ static int nfPartition_ProgramStart(whNvmFlashContext* context, int partition,
 static int nfPartition_ProgramCount(whNvmFlashContext* context, int partition,
         uint32_t count);
 static int nfPartition_ProgramInit(whNvmFlashContext* context, int partition);
+static int nfPartition_CheckRangeUnits(whNvmFlashContext* context,
+                                       int partition,
+                                       uint32_t units_offset,
+                                       uint32_t units_count);
```
billphipps, line 97: Changed to match byte-based solution below.
```suggestion
static int nfPartition_CheckDataRange(whNvmFlashContext* context,
                                       int partition,
                                       uint32_t byte_offset,
                                       uint32_t byte_count);
```

src/wh_nvm_flash.c
```diff
@@ -510,6 +546,13 @@ static int nfObject_ProgramDataBytes(whNvmFlashContext* context, int partition,
 
     data_offset = nfPartition_DataOffset(context, partition) + offset;
 
+    /* Ensure we don't program outside of the active partition */
+    if (WH_ERROR_OK != nfPartition_CheckRangeUnits(context, partition,
+                                    data_offset,
+                                    WHFU_BYTES2UNITS(byte_count))) {
```
billphipps, line 552: Update to use bytes
```suggestion
    if (WH_ERROR_OK != nfPartition_CheckDataRange(context, partition,
                                    data_offset * WHFU_BYTES_PER_UNIT,
                                    byte_count)) {
```

src/wh_nvm_flash.c
```diff
@@ -578,6 +621,13 @@ static int nfObject_ReadDataBytes(whNvmFlashContext* context, int partition,
     start = context->directory.objects[object_index].state.start;
     startOffset = nfPartition_DataOffset(context, partition) + start;
 
+    /* Ensure we don't read off the end of the active partition */
+    if (WH_ERROR_OK != nfPartition_CheckRangeUnits(context, partition,
+                                    startOffset + WHFU_BYTES2UNITS(byte_offset),
+                                    WHFU_BYTES2UNITS(byte_count))) {
```
billphipps, line 627: Fix the math bug (BYTES2UNITS always rounds up) by making this byte based:
```suggestion
    if (WH_ERROR_OK != nfPartition_CheckDataRange(context, partition,
                                    startOffset * WHFU_BYTESPERUNIT + byte_offset,
                                   byte_count)) {
```

billphipps (commented):
Fixed my bug and recommend to change to byte-based range check.

bigbrett (approved):
incorporated your suggestions, re-tested on tricore, looks good.

PR #4, opened 2024-03-13
attempt 1, failing with no SO_NOSIGPIPE found and implicit declaratio…

jpbland1:
…n of pread


src/wh_server.c
```diff
@@ -4,18 +4,20 @@
...
+#include <wolfssl/wolfcrypt/settings.h>
+#include <wolfssl/wolfcrypt/types.h>
+#include <wolfssl/wolfcrypt/error-crypt.h>
+#include <wolfssl/wolfcrypt/wc_port.h>
+#include <wolfssl/wolfcrypt/sha256.h>
+#include <wolfssl/wolfcrypt/aes.h>
+#include <wolfssl/wolfcrypt/rsa.h>
+#include <wolfssl/wolfcrypt/cmac.h>
+#include <wolfssl/wolfcrypt/kdf.h>
+#include <wolfssl/wolfcrypt/cryptocb.h>
```
billphipps, line 20: Let's leave "" here

src/wh_server.c
```diff
@@ -28,8 +30,573 @@
 
 #include "wolfhsm/wh_message.h"
 #include "wolfhsm/wh_message_comm.h"
+#include "wolfhsm/wh_packet.h"
 #include "wolfhsm/wh_server.h"
 
+#ifdef WOLFHSM_SHE_EXTENSION
```
billphipps, line 36: Recommend to move all of the SHE stuff to a separate file.  Ensure the any state variables are members of the server context to avoid globals.

src/wh_server.c
```diff
@@ -28,8 +30,573 @@
...
+uint8_t hsmSheCmacKeyFound = 0;
+uint8_t hsmSheRamKeyPlain = 0;
+uint32_t hsmSheBlSize = 0;
+uint32_t hsmSheBlSizeReceived = 0;
+uint32_t hsmSheInitRng = 0;
+/* cmac is global since the bootloader update can be called multiple times */
+Cmac sheCmac[1];
+#endif
+
+#if 0
```
billphipps, line 69: Can you can leave out any dormant code for now since we won't be reviewing it.

src/wh_server.c
```diff
@@ -54,7 +621,7 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
 static int _wh_Server_HandleCommRequest(whServerContext* server,
         uint16_t magic, uint16_t action, uint16_t seq,
         uint16_t req_size, const void* req_packet,
-        uint16_t *out_resp_size, void* resp_packet)
```
billphipps, line 57: As a style point, I'll normally leave the * attached to an "output" parameter.  Ok to change if this is confusing.

src/wh_server.c
```diff
@@ -119,35 +686,1428 @@ static int _wh_Server_HandleNvmRequest(whServerContext* server,
     return rc;
 }
 
+static int _wh_Server_HandleCryptoRequest(whServerContext* server,
```
billphipps, line 689: Consider moving this crypto handler to a separate file to simplify management.

src/wh_server.c
```diff
@@ -119,35 +686,1428 @@ static int _wh_Server_HandleNvmRequest(whServerContext* server,
...
+    (void)out_resp_size;
+    (void)resp_packet;
+    return 0;
 }
 
-static int _wh_Server_HandleCryptoRequest(whServerContext* server,
-        uint16_t magic, uint16_t action, uint16_t seq,
-        uint16_t req_size, const void* req_packet,
-        uint16_t *out_resp_size, void* resp_packet)
+#ifdef WOLFHSM_SHE_EXTENSION
```
billphipps, line 1259: Move to a separate SHE source file

billphipps (changes requested):
The SO_NOSIGPIPE was removed previously and replaced using send/recv with MSG_NOSIGNAL flags in the Posix TCP transport. You'll probably need to pull and rebase to correct that. pread/pwrite are supposed to be in unistd.h . Make sure the tests build and run cleanly in your build environment before adding stuff.

Please move all of the crypto and SHE functions into separate files wh_crypto.h/.c and wh_she.h/.c .  Please move message definitions/translations into wh_message_crypto.h/.c and wh_message_she.h/.c

bigbrett (changes requested):
@jpbland1 Looks like you added this to an old version of the code, could you rebase onto main? I think a lot of your errors will go away. Tests have been split out too.

wolfhsm/wh_cryptocb.h
```diff
@@ -0,0 +1,35 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+#ifndef WOLFHSM_CRYPTOCB_H
+#define WOLFHSM_CRYPTOCB_H
+#include <wolfhsm/common.h>
```
billphipps, line 23:
```suggestion
#include "wolfhsm/common.h"
```

Also, make sure to move all of the necessary includes for this file into here.

wolfhsm/wh_cryptocb.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifdef __cplusplus
+    extern "C" {
+#endif
```
billphipps, line 27: Let's remove __cplusplus guards for now.  We use C99 initializers which are not supported by c++.

wolfhsm/wh_cryptocb.h
```diff
@@ -0,0 +1,35 @@
+/* cryptocb.h
+ *
+ * Copyright (C) 2006-2023 wolfSSL Inc.
+ *
+ * This file is part of wolfSSL.
+ *
+ * wolfSSL is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
```
billphipps, line 9: All of wolfHSM is actually under GPLv3, not 2.  We'll go back and fix all of the licensing once we are mostly feature complete.

wolfhsm/wh_packet.h
```diff
@@ -0,0 +1,456 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+#ifndef WOLFHSM_PACKET_H
+#define WOLFHSM_PACKET_H
+#include <wolfhsm/common.h>
```
billphipps, line 23: Same header include comment

wolfhsm/wh_packet.h
```diff
@@ -0,0 +1,456 @@
...
+{
+    uint32_t ok;
+} wh_Packet_key_erase_res;
+
+typedef struct WOLFHSM_PACK wh_Packet_version_exchange
+{
+    uint32_t version;
+} wh_Packet_version_exchange;
+
+#ifdef WOLFHSM_SHE_EXTENSION
```
billphipps, line 281: Let's move all of the SHE stuff to a separate .h/.c file

wolfhsm/wh_packet.h
billphipps, line 1: Recommend to change name to wh_message_crypto.h

src/wh_cryptocb.c
```diff
@@ -0,0 +1,474 @@
...
+#include <wolfssl/wolfcrypt/types.h>
+#include <wolfssl/wolfcrypt/error-crypt.h>
+#include <wolfssl/wolfcrypt/aes.h>
+#include <wolfssl/wolfcrypt/cmac.h>
+#include <wolfssl/wolfcrypt/cryptocb.h>
+#include <wolfhsm/wh_packet.h>
+#include <wolfhsm/common.h>
+#include <wolfhsm/wh_error.h>
+#include <wolfhsm/wh_client.h>
+#include <wolfhsm/wh_message.h>
```
billphipps, line 36: Recommend to use "" instead of <>

src/wh_cryptocb.c
```diff
@@ -0,0 +1,474 @@
...
+                    rawPacket);
+                sleep(1);
+            } while (ret == WH_ERROR_NOTREADY);
+        }
+        if (ret == 0) {
+            if (group != WH_MESSAGE_GROUP_CRYPTO || action != WC_ALGO_TYPE_RNG)
+                ret = packet->error;
+            /* read out */
+            else
+                XMEMCPY(info->rng.out, out, packet->rngRes.sz);
```
billphipps, line 410: Recommend to check rngRes.sz for sanity before memcpy'ing

src/wh_cryptocb.c
```diff
@@ -0,0 +1,474 @@
...
+        /* set sz */
+        packet->rngReq.sz = info->rng.sz;
+        /* write request */
+        ret = wh_Client_SendRequest(ctx, group, WC_ALGO_TYPE_RNG,
+            sizeof(packet->rngReq), rawPacket);
+        if (ret == 0) {
+            do {
+                ret = wh_Client_RecvResponse(ctx, &group, &action, &dataSz,
+                    rawPacket);
+                sleep(1);
```
billphipps, line 402: This is not portable.  Let's replace this with XUSLEEP(1000) macro that the user can replace.  Default implementation simply returns

test/user_settings.h
billphipps, line 1: Grumble.  Not sure we should have a user_settings here.  Lemme think about it.

test/wh_test_crypto.c
billphipps, line 1: This should be changed to use the memtransport so we don't need networking for test.  Leave for now and I'll put together a client server test that you can directly reuse.

wolfhsm/common.h
billphipps, line 1: This file should be removed with its contents going in other spots.  consider wh_common.h

billphipps (changes requested):
Lots of integration nits, but we're getting there!

test/wh_test.c
```diff
@@ -587,13 +586,20 @@ static void* _whClientTask(void *cf)
...
         return NULL;
-    }
 
     ret = wh_Client_Init(client, config);
     printf("wh_Client_Init:%d\n", ret);
 
+    /* register the wolfHSM crypto callback */
+    if (wc_CryptoCb_RegisterDevice(myDevId, wolfHSM_CryptoCb, ctx) != 0)
+        return NULL;
+
```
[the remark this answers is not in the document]
jpbland1: done

test/wh_test.c
```diff
@@ -576,6 +574,7 @@ void wh_CommClientServer_TcpThreadTest(void)
 
 static void* _whClientTask(void *cf)
 {
+    int myDevId = 1;
```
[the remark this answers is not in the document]
jpbland1: fixed

wolfhsm/wh_server.h
```diff
@@ -3,6 +3,11 @@
 
 #include "wolfhsm/wh_comm.h"
 
+#ifndef WOLFSSL_USER_SETTINGS
+    #include <wolfssl/options.h>
+#endif
+#include <wolfssl/wolfcrypt/settings.h>
+
```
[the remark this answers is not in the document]
jpbland1: done

wolfhsm/wh_server.h
```diff
@@ -11,9 +16,9 @@
 /* Context structure to maintain the state of an HSM server */
 typedef struct whServerContext_t {
     whCommServer comm[1];
+    WC_RNG rng[1];
```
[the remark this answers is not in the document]
jpbland1: done

src/wh_server.c
```diff
@@ -4,18 +4,20 @@
 #include <stdlib.h>  /* For NULL */
 #include <string.h>  /* For memset, memcpy */
 
-#if 0
-/* wolfCrypt */
+#include <arpa/inet.h>
```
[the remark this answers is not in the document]
jpbland1: yes for a SHE function it must use big endian

src/wh_server.c
jpbland1: done

src/wh_server.c
jpbland1: no, want to get this all in the repo ASAP, it will come out as we get it working

src/wh_server.c
jpbland1: `(void*)myVal` or `(void *)myVal`? makes no sense putting part of the type on the variable name

wolfhsm/wh_cryptocb.h
jpbland1: done

wolfhsm/wh_packet.h
jpbland1: done

wolfhsm/wh_server.h
```diff
@@ -2,18 +2,51 @@
...
+    uint8_t confDigest[WOLFHSM_DIGEST_STUB];
+#ifdef WOLFHSM_SHE_EXTENSION
+    uint32_t count;
+#endif
+} NvmMetaData;
+
+typedef struct CacheSlot {
+    uint8_t commited;
+    NvmMetaData meta[1];
+    uint8_t buffer[WOLFHSM_NVM_MAX_OBJECT_SIZE];
```
billphipps, line 34: This is not going to work on small RAM devices.  Max object size is 64kB.  The intent would be to dynamically load whatever content is needed during the request and unload it afterwards.  The NVM Read function allows partial reads at an offset so only the pertinent part of the object data must be loaded in RAM at any given time.  Let's work on an alternate cache design that doesn't necessarily rely on fully preallocating RAM storage.  Ok for now, but this is a design issue.

billphipps (approved):
Still have a number of organizational/design comments, but we need to get this code put in now so we can continue to make progress.

src/wh_server.c
```diff
@@ -28,8 +30,613 @@
...
+#endif
+
+static int hsmGetUniqueId(whServerContext* server)
+{
+    int i;
+    int ret = 0;
+    uint16_t id;
+    /* try every index, unless both cache and nvm are full one will work */
+    for (id = 1; id < WOLFHSM_NUM_RAMKEYS + 1; id++) {
+        /* check against chache keys */
```
bigbrett, line 436:
```suggestion
        /* check against cache keys */
```

wolfhsm/wh_server.h
```diff
@@ -2,18 +2,51 @@
...
+typedef struct NvmMetaData {
+    uint16_t id;
+    uint32_t flags;
+    uint16_t len;
+    uint8_t label[WOLFHSM_NVM_LABEL_LEN];
+    uint8_t confDigest[WOLFHSM_DIGEST_STUB];
+#ifdef WOLFHSM_SHE_EXTENSION
+    uint32_t count;
+#endif
+} NvmMetaData;
```
bigbrett, line 29: Why are we defining a server-specific meta data struct here? Isn't [whNvmMetaData ](https://github.com/wolfSSL/wolfHSM/blob/main/wolfhsm/wh_common.h#L76)already defined in wh_common.h? I don't think we want multiple clones of this floating around

wolfhsm/wh_server.h
```diff
@@ -2,18 +2,51 @@
...
+    uint8_t commited;
+    NvmMetaData meta[1];
+    uint8_t buffer[WOLFHSM_NVM_MAX_OBJECT_SIZE];
+} CacheSlot;
+
+typedef struct {
+    curve25519_key curve25519Private[1];
+    curve25519_key curve25519Public[1];
+    WC_RNG rng[1]; 
+} crypto_context;
```
bigbrett, line 41: nit: we should probably follow the naming convention for public structs
```suggestion
}  whCryptoContext;
```

wolfhsm/wh_server.h
```diff
@@ -2,18 +2,51 @@
...
+#ifdef WOLFHSM_SHE_EXTENSION
+    uint32_t count;
+#endif
+} NvmMetaData;
+
+typedef struct CacheSlot {
+    uint8_t commited;
+    NvmMetaData meta[1];
+    uint8_t buffer[WOLFHSM_NVM_MAX_OBJECT_SIZE];
+} CacheSlot;
```
bigbrett, line 35: nit: we should probably follow naming convention for public structs
```suggestion
} whCacheSlot;
```

wolfhsm/wh_server.h
bigbrett, line 41: just musing (no action needed) but I wonder if this makes sense to define in its own file (e.g. `whCryptoContext.h`) since it will eventually grow to hold the state of every crypto algorithm we want to support

src/wh_server.c
```diff
@@ -28,8 +30,613 @@
...
+    return ret;
+}
+
+static int hsmCacheKey(whServerContext* server, NvmMetaData* meta, uint8_t* in)
+{
+    int i;
+    int foundIndex = -1;
+    /* make sure id is valid */
+    if (meta->id == 0)
+        return BAD_FUNC_ARG;
```
bigbrett, line 473: This comment is a catch-all for the whole file: I think we should be using wolfHSM error codes, not wolfSSL error codes. I think we only want to use wolfCrypt error codes to process result of wolfCrypt  functions

src/wh_server.c
```diff
@@ -28,8 +30,613 @@
...
+    for (i = 0; i < WOLFHSM_KEYSLOT_COUNT; i++) {
+        /* check for duplicate id, shouldn't cache a key in nvm */
+        if (ctx->nvmMetaCache[i].id == meta->id)
+            ctx->cache[foundIndex].commited = 1;
+    }
+#endif
+    /* write key if slot found */
+    XMEMCPY((uint8_t*)server->cache[foundIndex].buffer, in, meta->len);
+    XMEMCPY((uint8_t*)server->cache[foundIndex].meta, (uint8_t*)meta,
+        sizeof(NvmMetaData));
```
bigbrett, line 508: probably want `sizeof(server->cache[foundIndex].meta)`, especially since in my other comment I point out that we want to change the type to `whNvmMetaData`

src/wh_server.c
```diff
@@ -28,8 +30,613 @@
...
+    /* use empty string if we couldn't find the master ecu key */
+    if (ret != 0 && searchId == WOLFHSM_SHE_MASTER_ECU_KEY_ID) {
+        XMEMSET(out, 0, WOLFHSM_SHE_KEY_SZ);
+        meta->len = WOLFHSM_SHE_KEY_SZ;
+        meta->id = searchId;
+        ret = 0;
+    }
+#endif
+#endif
+    return ret;
```
bigbrett, line 590: is it possible to not get a hit on the key in the cache? It looks like this unconditionally returns success.

src/wh_server.c
```diff
@@ -28,8 +30,613 @@
...
+    }
+    if (ret > 0) {
+        XMEMSET(meta, 0, sizeof(NvmMetaData));
+        meta->len = privSz + pubSz;
+        meta->id = ret;
+        ret = hsmCacheKey(server, meta, keyBuf);
+    }
+    if (ret == 0)
+        ret = meta->id;
+    return ret;
```
bigbrett, line 616: I think we are mixing up return codes and the unique ID. It should probably return an error code and set the ID as an output argument. We need to be able to return an error code in case the wolfcrypt calls fail etc.

src/wh_server.c
```diff
@@ -119,35 +728,1492 @@ static int _wh_Server_HandleNvmRequest(whServerContext* server,
...
+            }
+            /* set the assigned id */
+            wc_curve25519_free(server->crypto->curve25519Private);
+            if (ret > 0) {
+                packet->pkCurve25519kgRes.keyId = ret;
+                *size = sizeof(packet->pkCurve25519kgRes);
+            }
+            else
+                ret = BAD_FUNC_ARG;
+            break;
```
bigbrett, line 1133: on success this will return the key ID not a valid return code. We need to refactor hsmCacheKey to return a proper error code, or at the very least, fix all cases of mixed up return code/key IDs in this function

wolfhsm/common.h
```diff
@@ -0,0 +1,226 @@
+
+/* common.h
```
bigbrett, line 2: We already have `wh_common.h` I don't think we should add an additional common file

bigbrett (changes requested):
Good start, but I think we need to clean up a lot before it can be merged in. I would really like to see all the dead code blocks removed. You can have a local branch that has them not removed, but we should be adding that code in piecemeal, rather than throwing it all in and commenting/ifdefing it out.

src/wh_server.c
jpbland1: errors are negative, all keyIds are positive and are limited to uint_16 so there's no overlap

src/wh_server.c
jpbland1: the dead code handled the miss case, updated since nvm isn't working

wolfhsm/common.h
jpbland1: fixed

src/wh_cryptocb.c
```diff
@@ -0,0 +1,531 @@
...
+            ret = hsmClientXferPacket(ctx);
+            if (ret == 0 && packet->type == WOLFHSM_ERROR)
+                ret = packet->error;
+            break;
+#endif
+        case WC_PK_TYPE_CURVE25519_KEYGEN:
+            packet->pkCurve25519kgReq.sz = info->pk.curve25519kg.size;
+            /* write request */
+            ret = wh_Client_SendRequest(ctx, group,
+                WC_PK_TYPE_CURVE25519_KEYGEN, sizeof(packet->rngReq),
```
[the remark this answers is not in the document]
jpbland1: yes, fixed

src/wh_cryptocb.c
```diff
@@ -0,0 +1,531 @@
...
+        if (ret == 0) {
+            do {
+                ret = wh_Client_RecvResponse(ctx, &group, &action, &dataSz,
+                    rawPacket);
+                sleep(1);
+            } while (ret == WH_ERROR_NOTREADY);
+        }
+        if (ret == 0) {
+            if (group != WH_MESSAGE_GROUP_CRYPTO || action != WC_ALGO_TYPE_RNG)
+                ret = packet->error;
```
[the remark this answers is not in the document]
jpbland1: my mistake, fixed and set error

src/wh_cryptocb.c
```diff
@@ -400,12 +82,15 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
             }
```
bigbrett, line 82: Still need to remove the `sleep()`s from this function. Line 398 for example (sorry, GH won't let me comment on that line in this review since it wasn't changed)

bigbrett (dismissed):
Nice, some good fixes! Did a quick pass with some things that might have been overlooked or not addressed from prior feedback. Still want to dive a little deeper but wanted to give a first pass. I resolved all issues you fixed, but there are still some remaining from the last review.

src/wh_server_crypto.c
```diff
@@ -0,0 +1,337 @@
+/* System libraries */
+#include <stdint.h>
+#include <stdlib.h>  /* For NULL */
+#include <string.h>  /* For memset, memcpy */
+
+#include <arpa/inet.h>
```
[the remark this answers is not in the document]
jpbland1: I think it's the she functions that need HTONS, I think bill had a note about this but I'll remove for now

src/wh_server_crypto.c
```diff
@@ -0,0 +1,337 @@
...
+    if (ret == 0)
+        ret = wc_curve25519_import_public(keyBuf, pubSz, key);
+    /* only import private if what we got back holds 2 keys */
+    if (ret == 0 && meta->len == CURVE25519_KEYSIZE * 2)
+        ret = wc_curve25519_import_private(keyBuf + pubSz, privSz, key);
+    return ret;
+}
+
+int _wh_Server_HandleCryptoRequest(whServerContext* server,
+    uint16_t* action, uint8_t* data, uint16_t* size)
```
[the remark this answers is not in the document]
jpbland1: Action doesn't tell you what error occurred, it tells the cryptocb layer that the action it wanted wasn't finished and then packet->error gives the specific error code

test/wh_test_crypto.c
```diff
@@ -189,36 +194,35 @@ static void _whClientServerThreadTest(whClientConfig* c_conf,
...
+                 .transport_cb      = tscb,
+                 .transport_context = (void*)tmsc,
+                 .transport_config  = (void*)tmcf,
+                 .server_id         = 5678,
     }};
-    whServerConfig s_conf[1] = {{
-            .comm = cs_conf,
-            /*.nvm = NULL, */
+    whServerConfig                  s_conf[1] = {{
+       .comm = cs_conf,
```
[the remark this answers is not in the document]
jpbland1: renamed to mem

test/wh_test_crypto.c
```diff
@@ -86,10 +89,12 @@ static void* _whClientTask(void *cf)
         printf("Failed to wc_curve25519_init_ex %d\n", ret);
         goto exit;
     }
+    printf("pre 1 wc_curve25519_make_key %d\n", ret);
     if ((ret = wc_curve25519_make_key(rng, CURVE25519_KEYSIZE, curve25519PrivateKey)) != 0) {
         printf("Failed to wc_curve25519_make_key %d\n", ret);
         goto exit;
     }
+    printf("pre 2 wc_curve25519_make_key %d\n", ret);
```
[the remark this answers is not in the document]
jpbland1: done

test/wh_test.c
```diff
@@ -18,8 +18,8 @@ int whTest_Unit(void)
     /* Comm tests */
     WH_TEST_ASSERT(0 == whTest_Comm());
     WH_TEST_ASSERT(0 == whTest_Crypto());
-    WH_TEST_ASSERT(0 == whTest_Flash_RamSim());
-    WH_TEST_ASSERT(0 == whTest_NvmFlash());
+    //WH_TEST_ASSERT(0 == whTest_Flash_RamSim());
+    //WH_TEST_ASSERT(0 == whTest_NvmFlash());
```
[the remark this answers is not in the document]
jpbland1: done

bigbrett:
@jpbland1 gonna give this a shot on the tricore

src/wh_cryptocb.c
billphipps, line 1: consider changing the name of this to wh_client_cryptocb.c as this is the cryptocb for the HSM client.

src/wh_server_crypto.c
```diff
@@ -0,0 +1,339 @@
...
+        ret = BAD_FUNC_ARG;
+    /* check the length, meta->len is the return size */
+    if (ret == 0 && ctx->nvmMetaCache[i].len > outLen)
+        ret = BUFFER_E;
+    /* read key */
+    if (ret == 0) {
+        ret = hal_flash_read(addr + WOLFHSM_HEADER_SIZE + keyOffset, out,
+            ctx->nvmMetaCache[i].len);
+    }
+    /* calculate key digest */
```
billphipps, line 157: Recommend to make integrity/auth of keys to be optional and not required at this point.  No changes for now.

src/wh_server_crypto.c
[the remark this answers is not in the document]
billphipps: The intent of the Group/Action was to identify the kind of request/response, with the expectation that a request of a specific group+action will result in a single response of that same group+action, even under error cases.  There's a check in comm_client that ensures that the response group+action matches the last request group+action.  The action should NOT be changed on the response under normal and error circumstances.  I haven't completed the async cancel handling yet that would allow an additional request or an unexpected response to be sent and handled correctly.  These would use fixed group+action values to make processing simple.

src/wh_server_she.c
```diff
@@ -0,0 +1,907 @@
...
+#include "wolfssl/wolfcrypt/aes.h"
+#include "wolfssl/wolfcrypt/rsa.h"
+#include "wolfssl/wolfcrypt/cmac.h"
+#include "wolfssl/wolfcrypt/kdf.h"
+#include "wolfssl/wolfcrypt/cryptocb.h"
+
+#include "wolfhsm/wh_server.h"
+#include "wolfhsm/wh_packet.h"
+
+#if 0
```
billphipps, line 25: Maybe hold off on including this file until we are ready to use it?

test/user_settings.h
billphipps, line 1: Still not excited about having a user_settings file here when we could put these defines into the makefile instead.  Leave for now.

billphipps (dismissed):
Still a number of bug/integration changes will be necessary, but we want to get this pushed in.  Please keep track of the requests for name changes, moving of functions to separate files, and the dependence on POSIX features.  I think many of the small things we can pick up quickly in future PR's, so lets get this in.

src/wh_cryptocb.c
```diff
@@ -0,0 +1,159 @@
...
+                } while (ret == WH_ERROR_NOTREADY);
+            }
+            if (ret == 0) {
+                if (group != WH_MESSAGE_GROUP_CRYPTO ||
+                    action != WC_ALGO_TYPE_PK)
+                    ret = packet->error;
+                /* read out */
+                else {
+                    info->pk.curve25519kg.key->devCtx =
+                        (void*)0 + packet->pkCurve25519kgRes.keyId;
```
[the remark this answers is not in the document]
jpbland1: typecasting between two different sized types throws an error so I did this instead, super ugly but everything else I tried threw an error

src/wh_cryptocb.c
```diff
@@ -0,0 +1,159 @@
...
+    uint8_t* iv;
+    uint8_t* in;
+    uint8_t* authIn;
+    uint8_t* authTag;
+    uint8_t* sig;
+    uint8_t* hash;
+#endif
+    int ret = CRYPTOCB_UNAVAILABLE;
+    whClientContext* ctx = inCtx;
+    uint8_t rawPacket[WH_COMM_MTU];
```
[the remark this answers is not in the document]
jpbland1: done

src/wh_server.c
```diff
@@ -124,29 +110,105 @@ static int _wh_Server_HandleKeyRequest(whServerContext* server,
...
-
-static int _wh_Server_HandleCryptoRequest(whServerContext* server,
-        uint16_t magic, uint16_t action, uint16_t seq,
-        uint16_t req_size, const void* req_packet,
-        uint16_t *out_resp_size, void* resp_packet)
-{
-    int rc = 0;
-    switch (action) {
-    /* TODO: Add crypto message handling here */
+#if 0
```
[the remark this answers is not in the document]
jpbland1: done

src/wh_cryptocb.c
jpbland1: done

src/wh_server.c
billphipps: wolfCrypt errors should be returned from cryptocb.c, but wolfHSM everywhere else.  Our goal was to have very few error codes to make autonomous handling simple.  Broke or not broke.

src/wh_cryptocb.c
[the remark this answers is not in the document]
billphipps: intptr_t is your friend.  :)

billphipps:
Here's my list of things that should be addressed in the future crypto increments:
1. Update licensing and copyright to GPLv3 and 2024.  This should be handled as a single global PR, once we have our filenames/structures finalized.
2. remove C++ guards.  This is C99 and we use things like struct x = {0}; for compiler-enforced zeroing of structs, especially on the stack, so we can avoid XMEMSET(0)'s
3. Update filenames and type names to follow the typical convention. 
4. Resolve ntoh() and byteswapping in a utility header.  Provide a config to support use of arpa/inet or byte swap, otherwise provide dumb, unoptimized versions.
5. make sure to validate reasonableness of request/response size/offset values before using them.
6. Eliminate posix dependencies lie sleep()/usleep and replace with XSLEEP XUSLEEP in a utility header.
7. Try to eliminate user_settings.h in the test folder and put defines into the Makefile just for feature enablement
8. RAM usage of the key cache is a problem.  Active memory usage in the server for keys should be the usable keys, not the unparsed DER's
9. Use wolfHSM error codes everywhere except cryptocb implementations.  BADARGS should only be used when local function arguments are not right.  Use ABORTED if a problem is detected after the function is started.
10. add optional static_assert on packet sizes to ensure they are less (including reasonable payloads) than the DATA_LEN
11. Add support for large or purely DMA-based crypto calls.  These should be separately configured so it is clear to the user when DMA is used and when memcpy to a buffer is used. 
12.

jpbland1:
I cant' reply to it but I've switched the action to be unchanged and added an rc field on the packet that explicitly gives the client the return code every time

src/wh_server.c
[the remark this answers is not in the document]
billphipps: Added to todo list

wolfhsm/wh_packet.h
billphipps: Added to todo list

src/wh_cryptocb.c
billphipps: Added to todo list

src/wh_cryptocb.c
billphipps: Added to todo list

test/user_settings.h
billphipps: Added to todo list

wolfhsm/wh_server.h
billphipps: Added to todo list

wolfhsm/wh_packet.h
```diff
@@ -0,0 +1,484 @@
...
+        wh_Packet_she_load_key_req sheLoadKeyReq;
+        wh_Packet_she_load_key_res sheLoadKeyRes;
+        wh_Packet_she_export_ram_key_res sheExportRamKeyRes;
+        wh_Packet_she_init_rng_res sheInitRngRes;
+        wh_Packet_she_rnd_res sheRndRes;
+        wh_Packet_she_extend_seed_req sheExtendSeedReq;
+        wh_Packet_she_extend_seed_res sheExtendSeedRes;
+#endif
+    };
+} whPacket;
```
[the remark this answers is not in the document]
billphipps: added to todo list

src/wh_cryptocb.c
```diff
@@ -0,0 +1,531 @@
...
+            packet->pkCurve25519kgReq.sz = info->pk.curve25519kg.size;
+            /* write request */
+            ret = wh_Client_SendRequest(ctx, group,
+                WC_PK_TYPE_CURVE25519_KEYGEN, sizeof(packet->rngReq),
+                rawPacket);
+            if (ret == 0) {
+                do {
+                    ret = wh_Client_RecvResponse(ctx, &group, &action, &dataSz,
+                        rawPacket);
+                    sleep(1);
```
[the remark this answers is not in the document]
billphipps: added to todo list

wolfhsm/wh_server.h
billphipps: added to todo list

src/wh_server.c
billphipps: Added to todo list

src/wh_cryptocb.c
```diff
@@ -0,0 +1,159 @@
...
+        default:
+            ret = CRYPTOCB_UNAVAILABLE;
+            break;
+        }
+        break;
+    case WC_ALGO_TYPE_RNG:
+        /* out is after the fixed size fields */
+        out = (uint8_t*)(&packet->rngRes + 1);
+        /* set sz */
+        packet->rngReq.sz = info->rng.sz;
```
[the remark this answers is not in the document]
billphipps: Added to todo list

billphipps (approved):
Make sure to start chewing on the todo list.

PR #5, opened 2024-03-15
Adding client/server NVM.  Missing some client functions.

billphipps:
Added client/server test sequential. Added all NVM messages and server handling.  Still adding client functions, but this is enough to show round-trip processing.  A few renames so I'll update wolfHSM-examples too.

Still owe:
1. multithread test when POSIX is defined -- Deferring to later
2. remaining client functions-- DONE
3. additional error/edge case handling. -- Addd more.  Will add as we go


src/wh_server_nvm.c
```diff
@@ -0,0 +1,311 @@
...
+        uint16_t hdr_len = sizeof(req);
+        whNvmMetadata meta = {0};
+        const uint8_t* data = (const uint8_t*)req_packet + hdr_len;
+        whMessageNvm_SimpleResponse resp = {0};
+
+        /* Convert request struct */
+        wh_MessageNvm_TranslateAddObjectRequest(magic,
+                (whMessageNvm_AddObjectRequest*)req_packet, &req);
+
+        if(req_size == (hdr_len + req.len)) {
```
bigbrett, line 133: Should we be checking `req_size` for every case? It looks like this is the only place we even use it. If we never are going to use it then we should probably remove it as an argument.

src/wh_server_nvm.c
```diff
@@ -0,0 +1,311 @@
...
+        wh_MessageNvm_TranslateDestroyObjectsRequest(magic,
+                (whMessageNvm_DestroyObjectsRequest*)req_packet, &req);
+
+        if (req.list_count <= WH_MESSAGE_NVM_MAX_DESTROY_OBJECTS_COUNT) {
+            /* Process the DestroyObjects action */
+            resp.rc = wh_Nvm_DestroyObjects(server->nvm,
+                    req.list_count, req.list);
+        } else {
+            /* Problem in transport or request */
+            resp.rc = WH_ERROR_ABORTED;
```
bigbrett, line 166: Just a thought: should this be a WH_ERROR_BADARGS? Or is that just meant for function args?

src/wh_nvm_flash.c
```diff
@@ -1076,91 +1091,115 @@ int wh_NvmFlash_AddObject(void* c, whNvmMetadata *meta,
...
+                                    .count = context->state.count,
+                                };
 
     /* Go through the current directory and mark the listed id's as bad */
-    int list_entry = 0;
-    int entry = 0;
-    int ret = 0;
     for (list_entry = 0; list_entry < list_count; list_entry++) {
-        /* Mark all matching ids as bad.  Should only be 1. */
+        /* Mark all matching entry as bad.  Should only be 1. */
```
bigbrett, line 1123:
```suggestion
        /* Mark all matching entrys as bad.  Should only be 1. */
```

src/wh_nvm.c
```diff
@@ -0,0 +1,142 @@
...
+    if (    (context == NULL) ||
+            (context->cb == NULL) ) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* No callback? Return ABORTED */
+    if (context->cb->AddObject == NULL) {
+        return WH_ERROR_ABORTED;
+    }
+    return context->cb->AddObject(context->context, meta, data_len, data);
```
bigbrett, line 76: Just at thought: If we aren't going to NULL check `data` or check `data_len` then we should make sure to note in the callback interface in `wh_nvm.h` what arguments the callbacks are responsible for checking (E.g. at the "NVM flash layer") and what will be checked by the NVM layer before dispatching the callback. As we start to add more back-ends, then documenting these things will be important.

test/wh_test_clientserver.c
```diff
@@ -0,0 +1,501 @@
...
+    WH_TEST_RETURN_ON_FAIL(
+            wh_Client_NvmGetAvailableResponse(client,
+                    &avail_size, &avail_objects,
+                    &reclaim_size, &reclaim_objects));
+#if defined(WH_CFG_TEST_VERBOSE)
+        printf("Client NvmGetAvailableResponse:%d, avail_size:%d avail_objects:%d, reclaim_size:%d reclaim_objects:%d\n",
+            ret, avail_size, avail_objects, reclaim_size, reclaim_objects);
+#endif
+
+#if 0
```
bigbrett, line 191: is this dead code going to be removed in your next commit w/ client NVM stuff? If not we should remove

bigbrett (changes requested):
Looks great. A few things I flagged but nothing major.

src/wh_server.c
```diff
@@ -21,29 +21,35 @@
...
-        /* All good */
+
+    if (    (config->nvm_config->cb != NULL) &&
+            (config->nvm_config->cb->Init != NULL)) {
+
+        rc = config->nvm_config->cb->Init(config->nvm_config->context, config->nvm_config->config);
+    }
+    if (rc == 0) {
+        server->nvm->cb = config->nvm_config->cb;
+        server->nvm->context = config->nvm_config->context;
```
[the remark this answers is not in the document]
billphipps: Yeah, I re-added the helper layer in whNvm.c as it made the code much more readable because there had to be a number of parameter validations performed before invoking the cb.  We should probably do the same for comms as well to keep the pattern.

src/wh_server.c
```diff
@@ -21,29 +21,35 @@
...
+
+        rc = wh_CommServer_Init(server->comm, config->comm_config);
+        if (rc == 0) {
+            /* All good */
+        }
```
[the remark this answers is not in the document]
billphipps: Not insane at all.  There's probably no downside to providing a NULL CommServer implementation so that this would still return success and any function call would return aborted (or not ready).  I think it is vital to error out of the server init if any of the components fail to initialize, but the user is free to put in really dumb implementations if they want to.  I can imagine a server build that does not support external comms an is only there to support HSM core debugging/benchmarking or for failure analysis.  I wanted to have "skeleton" implementations for each of the components, so let's use those as the NULL instances that always return reasonably to support unit testing.

src/wh_server.c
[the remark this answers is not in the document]
bigbrett: cool. Meant to delete this comment and replace with the one two below, but clearly forgot

src/wh_nvm.c
```diff
@@ -0,0 +1,142 @@
...
+            (context->cb == NULL) ) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* No callback? Return ABORTED */
+    if (context->cb->Init == NULL) {
+        return WH_ERROR_ABORTED;
+    }
+    return context->cb->Init(context->context, config);
+}
```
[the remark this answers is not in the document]
billphipps: When this is invoked, the context that is passed in is actually the one from the config, rather than the one that is in the upper level object.  We invoke init on this using the config to prove it's going to work.  So the binding that you are talking about has to happen at the next level up because we are not given the "real" context here, only the one from the config.  Maybe I should crank out a picture to better document this because if we are getting confused, then our customers will be too.

wolfhsm/wh_nvm.h
```diff
@@ -76,49 +68,47 @@ typedef struct {
 
     /* Read the data of the object starting at the byte offset */
     int (*Read)(void* context, whNvmId id, whNvmSize offset,
-            whNvmSize data_len, uint8_t* out_data);
+            whNvmSize data_len, uint8_t* data);
 } whNvmCb;
 
-#if 0
+#if 1
```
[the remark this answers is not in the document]
billphipps: Done

wolfhsm/wh_server.h
```diff
@@ -33,6 +24,12 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config);
  */
 int wh_Server_HandleRequestMessage(whServerContext* server);
 
+/* Handle an NVM request and generate a response */
+int wh_Server_HandleNvmRequest(whServerContext* server,
+        uint16_t magic, uint16_t action, uint16_t seq,
+        uint16_t req_size, const void* req_packet,
+        uint16_t *out_resp_size, void* resp_packet);
```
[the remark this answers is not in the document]
billphipps: Good point.  Lemme think about the "right" way of hiding this function.

src/wh_nvm.c
billphipps: Invoking AddObject with NULL data and 0 data length is absolutely ok.  The lower level nvm handler should be the one verify "goodness" of parameters and they are all required to do that.  This helper library is only stopping obvious insanity.

src/wh_server_nvm.c
```diff
@@ -0,0 +1,311 @@
...
+#include "wolfhsm/wh_nvm.h"
+
+#include "wolfhsm/wh_message.h"
+#include "wolfhsm/wh_message_nvm.h"
+#include "wolfhsm/wh_server.h"
+
+int wh_Server_HandleNvmRequest(whServerContext* server,
+        uint16_t magic, uint16_t action, uint16_t seq,
+        uint16_t req_size, const void* req_packet,
+        uint16_t *out_resp_size, void* resp_packet)
```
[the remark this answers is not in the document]
billphipps: Gaah!  I missed handling the return codes.  I'll add those checks.  Good catch.

src/wh_server_nvm.c
```diff
@@ -0,0 +1,311 @@
...
+{
+    int rc = 0;
+    switch (action) {
+
+    case WH_MESSAGE_NVM_ACTION_INIT:
```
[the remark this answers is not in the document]
billphipps: Originally, I had expected the NVM Init client call to provide some identity/authentication so that the server understands who is on the other side of the connection and could use that to help mask which objects it has access to.  Cleanup was expected to inform the server that no further authenticated access is necessary.  Just a placeholder for now.

src/wh_server_nvm.c
billphipps: Yes.  This is another miss.  req_size should always be checked to ensure the received packet is of the right size.  I'll add those checks.

src/wh_server_nvm.c
billphipps: Should be aborted.  BADARGS means that the function was called with something wrong and fixing the parameters will correct it.  In this case, we end up seeing a malformed request message so we must abort.  It is likely that the caller has a bad argument passed into the request function, but this should have been caught locally and returned then.

test/wh_test_clientserver.c
billphipps: Going to be turned on.

src/wh_nvm_flash.c
billphipps: done

src/wh_server_nvm.c
```diff
@@ -61,14 +81,19 @@ int wh_Server_HandleNvmRequest(whServerContext* server,
         whMessageNvm_ListRequest req = {0};
         whMessageNvm_ListResponse resp = {0};
 
-        /* Convert request struct */
-        wh_MessageNvm_TranslateListRequest(magic,
-                (whMessageNvm_ListRequest*)req_packet, &req);
+        if (req_size == 0) {
```
bigbrett, line 84: copy-pasta error?

```suggestion
        if (req_size == sizeof(req)) {
```

test/wh_test_clientserver.c
```diff
@@ -168,66 +169,82 @@ int whTest_ClientServerSequential(void)
...
-                wh_Client_NvmGetAvailableResponse(client,
+                wh_Client_NvmGetAvailableResponse(client, &server_rc,
                         &avail_size, &avail_objects,
                         &reclaim_size, &reclaim_objects));
         #if defined(WH_CFG_TEST_VERBOSE)
-                printf("Client NvmGetAvailableResponse:%d, avail_size:%d avail_objects:%d, reclaim_size:%d reclaim_objects:%d\n",
-                    ret, avail_size, avail_objects, reclaim_size, reclaim_objects);
+                printf("Client NvmGetAvailableResponse:%d, server_rc:%d, avail_size:%d avail_objects:%d, reclaim_size:%d reclaim_objects:%d\n",
+                    ret, server_rc, avail_size, avail_objects, reclaim_size, reclaim_objects);
         #endif
```
bigbrett, line 233: We should probably make this portion of the test self-checking (e.g. how we did the `addObjectWithReadbackCheck` functions for the NVM flash layer) - that way the test will fail if the objects we expected to read back aren't the ones we wrote (or don't exist). Currently this requires us to read the output of printfs to sanity check everything except return values. Very easy for a broken implementation to return success but not do what we think it is doing.

bigbrett (changes requested):

wolfhsm/wh_server.h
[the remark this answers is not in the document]
bigbrett: I think it is fine to just have this in `wh_server.c` given that it is "internal"

src/wh_server.c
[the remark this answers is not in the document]
bigbrett: OK so we want all components to be required, (e.g. NVM or comms cant be null) but tell customers to register dummy handlers if they don't need? I think that is fine. Def don't need to do now, but if that is the case, maybe we should probably just provide a dummy handler for each layer to make that easier for them to just register if they don't want a component.

src/wh_server.c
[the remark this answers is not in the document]
bigbrett: I realize now this comment should have selected the line:
```
        rc = config->nvm_config->cb->Init(config->nvm_config->context, config->nvm_config->config);
```
not the context/callback settings. 


But now that we have the helper layer, should we use that here, vs invoking callbacks directly?

src/wh_nvm.c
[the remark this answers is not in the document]
bigbrett: Yes please, some documentation on this would be great. It is the paradigm we are using for a lot of the layers so would not be wasted effort to diagram it

bigbrett (changes requested):
Nice. Haven't fully dove into wh_client_nvm.c yet, just did a quick pass, but I think we need to add self-checking tests before merging. We can't be parsing printfs to determine if the test passed or failed. Otherwise, 10000 foot view looks great.

src/wh_server.c
```diff
@@ -3,40 +3,93 @@
...
     return rc;
 }
 
+int wh_Server_Cleanup(whServerContext* server)
+{
+    if (server ==NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    (void)wh_CommServer_Cleanup(server->comm);
```
jpbland1, line 85: should we be ignoring the return codes here? if cleanup fails the client won't know if they tried to init again

jpbland1 (approved):
looks good and ran on my setup, just the ignored return codes I noticed

bigbrett (approved):
Looks great, works on tricore and POSIX. Merging with the assumption we will address readback tests later on

PR #8, opened 2024-03-22
operation forklift: bare metal fixes

bigbrett:
- Fixes some incompatibilities of recently merged code with bare-metal devices (both build and runtime errors)
- Fixes memory leak w/ server RNG 
- some formatting/housekeeping

### DIFF
--- a/src/wh_server.c
+++ b/src/wh_server.c
@@ -19,8 +19,10 @@
 /* Server API's */
 #include "wolfhsm/wh_server.h"
 #include "wolfhsm/wh_server_crypto.h"
-#include "wolfhsm/wh_server_she.h"
 #include "wolfhsm/wh_server_internal.h"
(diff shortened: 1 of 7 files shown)


src/wh_server.c
```diff
@@ -84,6 +85,7 @@ int wh_Server_Cleanup(whServerContext* server)
 
     (void)wh_CommServer_Cleanup(server->comm);
     (void)wh_Nvm_Cleanup(server->nvm);
+    (void)wc_FreeRng(server->crypto->rng);
```
billphipps, line 88: All of the crypto component init and shutdown should be self contained.  No changes necessary.

src/wh_server_she.c
```diff
@@ -1,3 +1,5 @@
+#if 0
```
billphipps, line 1: if def wolfhsm_she_extension?

PR #9, opened 2024-03-22
add rsa, currently fails because rsa keygen and export

jpbland1:
both require malloc and current build settings require WOLFSSL_NO_MALLOC, may need to change for rsa builds

### DIFF
--- a/src/wh_client_cryptocb.c
+++ b/src/wh_client_cryptocb.c
@@ -87,7 +87,7 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
                     ret = packet->rc;
                 else {
                     info->pk.rsakg.key->devCtx =
-                        (void*)0 + packet->pkRsakgRes.keyId;
+                        (void*)((intptr_t)packet->pkRsakgRes.keyId);
                 }
             }
             break;
@@ -99,7 +99,7 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
             packet->pkRsaReq.opType = info->pk.rsa.type;
             /* set keyId */
             packet->pkRsaReq.keyId =
-                *((uint32_t*)(&info->pk.rsa.key->devCtx));
+                (intptr_t)(info->pk.rsa.key->devCtx);
             /* set inLen */
             packet->pkRsaReq.inLen = info->pk.rsa.inLen;
             /* set outLen */
@@ -133,7 +133,7 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
         case WC_PK_TYPE_RSA_GET_SIZE:
             /* set keyId */
             packet->pkRsaGetSizeReq.keyId =
-                *((uint32_t*)(&info->pk.rsa_get_size.key->devCtx));
+                (intptr_t)(info->pk.rsa_get_size.key->devCtx);
             /* write request */
             ret = wh_Client_SendRequest(ctx, group,
                 WC_ALGO_TYPE_PK,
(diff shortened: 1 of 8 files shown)


src/wh_server_crypto.c
```diff
@@ -81,6 +139,88 @@ int wh_Server_HandleCryptoRequest(whServerContext* server,
...
+                        packet->pkRsaRes.outLen = field;
+                        *size = WOLFHSM_PACKET_STUB_SIZE +
+                            sizeof(packet->pkRsaRes) + field;
+                    }
+                    break;
+            }
+            break;
+        case WC_PK_TYPE_RSA_GET_SIZE:
+            /* init rsa key */
+            ret = wc_InitRsaKey_ex(server->crypto->rsa, NULL, INVALID_DEVID);
```
billphipps, line 204: Not sure INVALID_DEVID is correct here, but server->dev_id isn't correct either.  Based on the keyId, we'll need to figure out if the key exists in RAM, NVM, or is a HW-only key.  No changes here, but we'll need to resolve the HW-only case soon.  We already have a backlog entry for this.

src/wh_server_crypto.c
```diff
@@ -81,6 +139,88 @@ int wh_Server_HandleCryptoRequest(whServerContext* server,
     case WC_ALGO_TYPE_PK:
         switch (packet->pkAnyReq.type)
         {
+#ifndef NO_RSA
+        case WC_PK_TYPE_RSA_KEYGEN:
```
billphipps, line 143: Need to add some per-action packet size validations.  I'll add a backlog entry.  Ok for now.

src/wh_client_cryptocb.c
```diff
@@ -66,6 +66,96 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
+                    ret = wh_Client_RecvResponse(ctx, &group, &action, &dataSz,
+                        rawPacket);
+                } while (ret == WH_ERROR_NOTREADY);
+            }
+            if (ret == 0) {
+                if (packet->rc != 0)
+                    ret = packet->rc;
+                else {
+                    info->pk.rsakg.key->devCtx =
+                        (void*)0 + packet->pkRsakgRes.keyId;
```
billphipps, line 90: void* arithmetic is generally frowned upon.  Can you switch this to typecast to uintptr_t and then to void*?  That is the best way to keep analyzers happy and then intent of putting an integer into a pointer member.

billphipps (dismissed):
Looks good.  A few nits but I think we should get this in so I can throw it at the SPC compiler, which always seems to find complaints.

bigbrett (commented):
Question about one thing, but good to merge if I am misunderstanding how it should work. A clearer comment addressing the source of my confusion would be great. As always, I'd like to see a few more test cases here giving us more coverage for things like the key not being in cache but needing to be loaded, as well as key already in cache, etc.

src/wh_server_keystore.c
```diff
@@ -89,6 +117,52 @@ int hsmCacheKey(whServerContext* server, whNvmMetadata* meta, uint8_t* in)
...
+    if (server == NULL || keyId == WOLFHSM_ID_ERASED)
+        return WH_ERROR_BADARGS;
+    for (i = 0; i < WOLFHSM_NUM_RAMKEYS; i++) {
+        /* check for empty slot or rewrite slot */
+        if ((foundIndex == -1 &&
+            server->cache[i].meta->id == WOLFHSM_ID_ERASED) ||
+            server->cache[i].meta->id == keyId) {
+            foundIndex = i;
+            if (server->cache[i].meta->id == keyId)
+                return i;
```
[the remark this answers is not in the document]
jpbland1: no, that `if` case happens when the key is found to be already in the cache. The `if` statement this one is nested in checks for keyId match or empty, and will continue checking even if it finds an empty spot just in case there's an empty spot before the key.

bigbrett (approved):

PR #10, opened 2024-03-22
Server-side crypto callback support

bigbrett:
- Adds support for server-side crypto callbacks
- Some formatting/housekeeping cleanups

Tested full stack on tricore and it works for current crypto tests on `main` (e.g. RNG and Curve25519) !
 (` hardware crypto <--> server cryptocb <--> wolfHSM server (HSM) <--| IPC comms |--> wolfHSM client (tricore)`)

### DIFF
--- a/src/wh_server.c
+++ b/src/wh_server.c
@@ -70,15 +70,19 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
     server->crypto->devId = config->devId;
     if (config->cryptocb != NULL) {
         /* register the crypto callback with wolSSL */
-        rc = wc_CryptoCb_RegisterDevice(config->devId, config->cryptocb, NULL);
+        rc = wc_CryptoCb_RegisterDevice(server->crypto->devId,
+                                        config->cryptocb,
+                                        NULL);
         if (rc != 0) {
             (void)wh_Server_Cleanup(server);
             return WH_ERROR_ABORTED;
         }
     }
-#endif /* WOLF_CRYPTO_CB */
+#else
+    server->crypto->devId = INVALID_DEVID;
+#endif
 
-    rc = wc_InitRng_ex(server->crypto->rng, NULL, config->devId);
+    rc = wc_InitRng_ex(server->crypto->rng, NULL, server->crypto->devId);
     if (rc != 0) {
         (void)wh_Server_Cleanup(server);
         return WH_ERROR_ABORTED;
--- a/src/wh_server.c
+++ b/src/wh_server.c
@@ -53,26 +53,49 @@ static int _wh_Server_HandleCustomRequest(whServerContext* server,
 int wh_Server_Init(whServerContext* server, whServerConfig* config)
 {
     int rc = 0;
+
     if ((server == NULL) || (config == NULL)) {
         return WH_ERROR_BADARGS;
     }
 
     memset(server, 0, sizeof(*server));
-    if (
-        ((rc = wolfCrypt_Init()) == 0) &&
-        ((rc = wc_InitRng_ex(server->crypto->rng, NULL, INVALID_DEVID)) == 0)) {
-        rc = wh_Nvm_Init(server->nvm, config->nvm_config);
-        if (rc == 0) {
-            server->nvm->cb = config->nvm_config->cb;
-            server->nvm->context = config->nvm_config->context;
 
-            rc = wh_CommServer_Init(server->comm, config->comm_config);
-            if (rc == 0) {
-                /* All good */
-            }
+    rc = wolfCrypt_Init();
+    if (rc != 0) {
+        (void)wh_Server_Cleanup(server);
+        return WH_ERROR_ABORTED;
+    }
+
+#if defined(WOLF_CRYPTO_CB)
+    server->crypto->devId = config->devId;
+    if (config->cryptocb != NULL) {
+        /* register the crypto callback with wolSSL */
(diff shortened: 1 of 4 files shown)


src/wh_server.c
```diff
@@ -53,26 +53,53 @@ static int _wh_Server_HandleCustomRequest(whServerContext* server,
...
-        rc = wh_Nvm_Init(server->nvm, config->nvm_config);
-        if (rc == 0) {
-            server->nvm->cb = config->nvm_config->cb;
-            server->nvm->context = config->nvm_config->context;
 
-            rc = wh_CommServer_Init(server->comm, config->comm_config);
-            if (rc == 0) {
-                /* All good */
-            }
+    rc = wolfCrypt_Init();
```
billphipps, line 63: Let's be a bit more explicit during initialization and tear down to make sure it is always safe to call Server_Cleanup:
1. Add int wc_inited to server_context.  0 means not inited yet, 1 = inited successfully and needs to be cleaned up in ServerCleanup.
2. Use context.dev_id == 0 to mean cryptocb not registered yet, == INVALID_DEVID for no cryptocb, and otherwise cryptocb registered.  Deregister the cryptocb in server_cleanup before wolfCrypt_Cleanup if not 0.  Note that invoking register/deregister will trigger events to be sent to the cryptocb handler to allow hw initialization late if the dev context is NULL when WOLF_CRYPTO_CB_CMD is set, which it should be.  see https://github.com/wolfSSL/wolfssl/blob/master/wolfcrypt/src/cryptocb.c#L334 for more info.
3. Not sure why we are init'ing the RNG, but if we need one per server context, then add a context member to track if the InitRNG was successful so we can FreeRNG in server_cleanup if necessary.
4. NVM Init now updates the nvm_context on init and correctly unsets it on init failure so no additional state is necessary here
5. We should update CommServer_init to ensure it can correctly cleanup on failure.

src/wh_server.c
```diff
@@ -53,26 +53,53 @@ static int _wh_Server_HandleCustomRequest(whServerContext* server,
...
+        return WH_ERROR_ABORTED;
+    }
+
+    rc = wh_Nvm_Init(server->nvm, config->nvm_config);
+    if (rc != 0) {
+        (void)wh_Server_Cleanup(server);
+        return WH_ERROR_ABORTED;
+    }
+    server->nvm->cb      = config->nvm_config->cb;
+    server->nvm->context = config->nvm_config->context;
```
billphipps, line 97: This should have already been moved into Nvm_Init.  I should have removed these lines.  my bad.

wolfhsm/wh_server.h
```diff
@@ -39,6 +41,10 @@ typedef struct whServerContext_t {
 typedef struct whServerConfig_t {
     whCommServerConfig* comm_config;
     whNvmConfig* nvm_config;
+#if defined WOLF_CRYPTO_CB /* TODO: should we be relying on wolfSSL defines? */
```
billphipps, line 44: devId should ALWAYS be defined and set to invalid_dev_id when no cryptocb is available.  dev_id=0 means server is not initialized yet.

For now, let's reuse the wolfCrypt defines to ensure nothing is out of sync.

wolfhsm/wh_server.h
```diff
@@ -23,6 +24,7 @@ typedef struct CacheSlot {
 } CacheSlot;
 
 typedef struct {
+    int devId;
     curve25519_key curve25519Private[1];
     curve25519_key curve25519Public[1];
     WC_RNG rng[1];
```
billphipps, line 30: I don't understand why we have these in the server context.  Leave for now, but we should be sensitive to having extra space here when it may only be needed temporarily.  If these are used mutually exclusively, then consider a union.

billphipps (changes requested):
Looks good.  Some updates are needed to handle cleanup when failing during init.  We have the same problem in the client-side too.

src/wh_server.c
jpbland1: on 2, I'm pretty sure 0 is a valid devId, we only reserve -2 as an invalid id. I think `wh_Server_Init` should take an explicit argument for devId the same way wolfCrypt does it and just require the caller to specify if they're using a devId or not to get around the memset 0 of the crypto struct.

jpbland1 (changes requested):

wolfhsm/wh_server.h
bigbrett: Keep in mind this is in the config struct - devId is always defined in the server context, since we will always need to pass it into the internally invoked wolfCrypt functions. Here, if we don't have support for crypto callbacks (e.g. `WOLF_CRYPTO_CB` isn't defined) then it doesn't exist in the config struct, and is always stored as `INVALID_DEVID` when passed to wolfCrypt APIs internally (which are then ignored since we haven't compiled in support for CCBs). 

I agree with @jpbland1 that we don't want a secret meaning to devId==0, as that is a perfectly valid devId. Instead I added an additional flag for devId initialization.

wolfhsm/wh_server.h
bigbrett: I agree, though they might have been put there to persist state across cryptoCb invocations? Could use a pointer and a serparate context struct through, or the cryptoCb context for that

src/wh_server.c
bigbrett: @billphipps can you clarify what you mean on 5? Action items for this PR of for a subsequent one? Since this PR just adds support for server CCBs I don't want to touch too much stuff

src/wh_server.c
bigbrett: done

PR #11, opened 2024-03-25
Spc integration fixes

billphipps:
A number of small changes to make the HighTec compiler happy again. I suspect additional updates as we go.

Bigger change: Pulled out the nvm_init and wolfcrypt_init/register/rng_init from server_init.  These must be done prior to server_init and this breaks when supporting multiple server instances when we have multiple clients.


src/wh_client_cryptocb.c
```diff
@@ -84,7 +85,7 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
                 /* read out */
                 else {
                     info->pk.curve25519kg.key->devCtx =
-                        (void*)0 + packet->pkCurve25519kgRes.keyId;
```
jpbland1, line 87: I think I found a better way to do this that makes all sizes the same when I added ecc
```suggestion
                        XMEMCPY((void*)&info->pk.curve25519kg.key->devCtx, (void*)&packet->pkCurve25519kgRes.keyId, sizeof(packet->pkCurve25519kgRes.keyId));
```

billphipps: I couldn't find the definition for WOLFHSM_PACKET_STUB_SIZE in any of your PR's.  Can you point that to me?

test/wh_test_clientserver.c
```diff
@@ -199,13 +199,23 @@ int whTest_ClientServerSequential(void)
...
        .comm_config = cs_conf,
-       .nvm_config = n_conf,
+       .nvm = nvm,
+       .crypto = crypto,
     }};
     whServerContext                server[1] = {0};
 
+    wolfCrypt_Init();
+    wc_InitRng_ex(crypto->rng, NULL, crypto->devId);
+    wh_Nvm_Init(nvm, n_conf);
```
bigbrett, line 217: check return values for these three functions
```suggestion
    WH_TEST_RETURN_ON_FAIL(wolfCrypt_Init());
    WH_TEST_RETURN_ON_FAIL(wc_InitRng_ex(crypto->rng, NULL, crypto->devId));
    WH_TEST_RETURN_ON_FAIL(wh_Nvm_Init(nvm, n_conf));
```

src/wh_server.c
```diff
@@ -54,42 +54,15 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
     }
 
     memset(server, 0, sizeof(*server));
+    server->crypto = config->crypto;
+    server->nvm = config->nvm;
```
bigbrett, line 58: How should we handle a `NULL` NVM? Disallow? 

Right now, this could be NULL and thankfully it only returns failure to the client since the lower-level NVM handlers NULL check, but I absolutely can see this potentially resulting in a NULL dereference somewhere at some point if we don't properly enforce. 

Also, note that the tests still all pass with NULL `server->nvm` since they don't actually check any return values :(

bigbrett (changes requested):
Looks great. pointed out some potential issues with NULL server->crypto and NULL server->nvm cases. Let me know your thoughts on whether this should be allowed

test/wh_test_crypto.c
```diff
@@ -269,14 +269,29 @@ static void wh_ClientServer_MemThreadTest(void)
...
        .comm_config = cs_conf,
-       .nvm_config = n_conf,
+       .nvm = nvm,
+       .crypto = crypto,
     }};
 
+    wh_Nvm_Init(nvm, n_conf);
+
+    wolfCrypt_Init();
+    wc_InitRng_ex(crypto->rng, NULL, crypto->devId);
```
bigbrett, line 288: check return values

billphipps, line 288:
```suggestion
    WH_TEST_RETURN_ON_FAIL(wh_Nvm_Init(nvm, n_conf));

    WH_TEST_RETURN_ON_FAIL(wolfCrypt_Init());
    WH_TEST_RETURN_ON_FAIL(wc_InitRng_ex(crypto->rng, NULL, crypto->devId));
```

src/wh_client_cryptocb.c
bigbrett: @billphipps [defined here](https://github.com/wolfSSL/wolfHSM/blob/b92437a951a813572f2649a9020cce64fe61cade/wolfhsm/wh_packet.h#L36)

src/wh_server.c
billphipps: Returning WH_ERROR_BADARGS is probably the correct thing as this captures the problem of a dynamic allocation that fails as well.

bigbrett (approved):

PR #18, opened 2024-04-04
Added readback/self-checking NVM tests to whTest_ClientServerSequential

bigbrett:
Test will now fail if server NVM isn't working as expected from the client perspective.

Note: clang-format caused a lot of whitespace changes so recommend you turn off the whitespace diff when looking at the PR in github (Files changed -> settings button -> check hide whitespace)


jpbland1 (approved):
Reviewed and tested, working fine

PR #21, opened 2024-04-11
server DMA address handling

bigbrett:
- Adds support for custom platform-specific DMA callbacks, allowing for injecting operations like cache invalidation, address remapping, etc. before/after accessing client memory
- Adds basic support for memory "allow lists", enabling each server object to maintain a list of allowed regions of system memory that the client is able to read/write from
- Adds the above features to the NVM message handling layer 
- Tests for the above features that run on the POSIX sim and tricore

### A few specific areas that I'd like feedback on:
- data types / address manipulation logic when going between 32/64-bit client addresses and native server addresses
- what should be defined in `wh_server.h` vs `wh_server_dma.h`? Do we want all public server APIs in `wh_server.h` and everything else in the other module? We also have a circular dependency between the two, but two separate modules is nice to reduce file size. Happy to relocate stuff wherever, so just let me know if you have strong opinions.
- [hidden "allow list" check ](https://github.com/bigbrett/wolfHSM/blob/192fa0176b7f7db9b5dbc542cbe88f7698fed97c/src/wh_server_dma.c#L109)inside `wh_Server_DmaProcessClientAddress{32,64}` vs pulling it out and having that be an explicit separate step. Right now this function does the work and the allowlist checks are private. I'm leaning towards pushing another commit that removes the allow list check inside this function such that it only invokes the DMA callback, and making address screening accessible via public function. Having all allowlist checks in one place might be wise. 
- Use of `goto`s in [handling the NVM DMA message requests](https://github.com/bigbrett/wolfHSM/blob/192fa0176b7f7db9b5dbc542cbe88f7698fed97c/src/wh_server_nvm.c#L304) - There are a large number of sequential function calls that each produce a return code that should be returned, preventing further processing from taking place. I tried to write this using nested `if (resp.rc == WH_ERROR_OK)` blocks, but it got way too deeply nested and ugly. I think the `goto` solution is the most concise, but I know `goto`s are controversial and sometimes frowned upon, so lets discuss if this section of code is a problem.

### DIFF
--- a/src/wh_server_dma.c
+++ b/src/wh_server_dma.c
@@ -92,7 +92,8 @@ int wh_Server_DmaProcessClientAddress32(whServerContext* server,
         *serverPtr = (void*)((uintptr_t)clientAddr);
     }
 
-    /* if the server has a allowlist registered, check address against it */
+    /* if the server has a allowlist registered, check transformed address
+     * against it */
     if (server->dmaAddrAllowList != NULL) {
         rc = _checkMemOperAgainstAllowList(server->dmaAddrAllowList, oper,
                                            *serverPtr, len);
@@ -139,13 +140,24 @@ int whServerDma_CopyFromClient32(struct whServerContext_t* server,
 {
     int rc = WH_ERROR_OK;
 
+    void* transformedAddr = NULL;
+
     /* TODO: should len be checked against UINTxxMax? Should it be size_t? */
     if (NULL == server || NULL == serverPtr || 0 == len) {
         return WH_ERROR_BADARGS;
     }
 
+    /* Check the server address against the allow list */
+    rc = _checkMemOperAgainstAllowList(server->dmaAddrAllowList,
+                                       WH_DMA_OPER_CLIENT_READ_PRE, serverPtr,
+                                       len);
+    if (rc != WH_ERROR_OK) {
+        return rc;
+    }
+
+    /* Process the client address pre-read */
     rc =
-        wh_Server_DmaProcessClientAddress32(server, clientAddr, &serverPtr, len,
+        wh_Server_DmaProcessClientAddress32(server, clientAddr, &transformedAddr, len,
                                             WH_DMA_OPER_CLIENT_READ_PRE, flags);
     if (rc != WH_ERROR_OK) {
         return rc;
@@ -153,9 +165,10 @@ int whServerDma_CopyFromClient32(struct whServerContext_t* server,
 
     /* Perform the actual copy */
     /* TODO should we add a flag to force client word-sized reads? */
-    memcpy(serverPtr, (void*)((uintptr_t)clientAddr), len);
+    memcpy(serverPtr, transformedAddr, len);
 
-    rc = wh_Server_DmaProcessClientAddress32(server, clientAddr, &serverPtr,
+    /* Process the client address post-read */
+    rc = wh_Server_DmaProcessClientAddress32(server, clientAddr, &transformedAddr,
                                              len, WH_DMA_OPER_CLIENT_READ_POST,
                                              flags);
 
@@ -169,25 +182,37 @@ int whServerDma_CopyFromClient64(struct whServerContext_t* server,
 {
     int rc = WH_ERROR_OK;
 
+    void *transformedAddr = NULL;
+
     /* TODO: should len be checked against UINTxxMax? Should it be size_t? */
     if (NULL == server || NULL == serverPtr || 0 == len) {
         return WH_ERROR_BADARGS;
     }
 
-    rc =
-        wh_Server_DmaProcessClientAddress64(server, clientAddr, &serverPtr, len,
-                                            WH_DMA_OPER_CLIENT_READ_PRE, flags);
+    /* Check the server address against the allow list */
+    rc = _checkMemOperAgainstAllowList(server->dmaAddrAllowList,
+                                       WH_DMA_OPER_CLIENT_READ_PRE, serverPtr,
+                                       len);
+    if (rc != WH_ERROR_OK) {
+        return rc;
+    }
+
+    /* Process the client address pre-read */
+    rc = wh_Server_DmaProcessClientAddress64(
+        server, clientAddr, &transformedAddr, len, WH_DMA_OPER_CLIENT_READ_PRE,
+        flags);
     if (rc != WH_ERROR_OK) {
         return rc;
     }
 
     /* Perform the actual copy */
     /* TODO should we add a flag to force client word-sized reads? */
-    memcpy(serverPtr, (void*)((uintptr_t)clientAddr), len);
+    memcpy(serverPtr, transformedAddr, len);
 
-    rc = wh_Server_DmaProcessClientAddress64(server, clientAddr, &serverPtr,
-                                             len, WH_DMA_OPER_CLIENT_READ_POST,
-                                             flags);
+    /* process the client address post-read */
+    rc = wh_Server_DmaProcessClientAddress64(
+        server, clientAddr, &transformedAddr, len, WH_DMA_OPER_CLIENT_READ_POST,
+        flags);
 
     return rc;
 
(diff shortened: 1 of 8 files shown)


src/wh_server_dma.c
```diff
@@ -0,0 +1,307 @@
...
+{
+    uintptr_t startAddr = (uintptr_t)addr;
+    uintptr_t endAddr   = startAddr + size;
+
+    /* Check if the address range is fully within a allowlist entry */
+    for (int i = 0; i < WH_DMA_ADDR_ALLOWLIST_SIZE; i++) {
+        uintptr_t allowlistStartAddr = (uintptr_t)allowList[i].addr;
+        uintptr_t allowlistEndAddr   = allowlistStartAddr + allowList[i].size;
+
+        if (startAddr >= allowlistStartAddr && endAddr <= allowlistEndAddr) {
```
billphipps, line 21: If allowList[].size == 0, then you shouldn't check for an allow.  Currently, a size==0 will allow addr==allowList[].addr.  I  also think an incoming size of 0 should always fail with WH_ERROR_BADARGS

src/wh_server_dma.c
```diff
@@ -0,0 +1,307 @@
...
+{
+    int rc = WH_ERROR_OK;
+
+    /* If a read/write operation is requested, check the transformed address
+     * against the appropriate allowlist
+     *
+     * TODO: do we need to allowlist check on POST in case there are subsequent
+     * memory operations for some reason?
+     */
+    if (oper == WH_DMA_OPER_CLIENT_READ_PRE) {
```
billphipps, line 41: Consider switch() with default to return WH_ERROR_BADARGS if the oper is unknown.

src/wh_server_dma.c
```diff
@@ -0,0 +1,307 @@
...
+    else if (oper == WH_DMA_OPER_CLIENT_WRITE_PRE) {
+        rc = _checkAddrAgainstAllowList(allowList->writeList, addr, size);
+    }
+
+    return rc;
+}
+
+int wh_Server_DmaRegisterCb32(whServerContext* server, whServerDmaClientMem32Cb cb)
+{
+    if (NULL == server || NULL == cb) {
```
billphipps, line 53: Should cb==NULL be used to use a dummy/default CB that allows everything and does no translation?

src/wh_server_dma.c
```diff
@@ -0,0 +1,307 @@
...
+
+    server->dma.cb64 = cb;
+
+    return WH_ERROR_OK;
+}
+
+int wh_Server_DmaRegisterAllowList(whServerContext*          server,
+                                   const whServerDmaAddrAllowList* allowlist)
+{
+    if (NULL == server || NULL == allowlist) {
```
billphipps, line 76: consider using allowlist == NULL to bypass address checking

src/wh_server_dma.c
```diff
@@ -0,0 +1,307 @@
...
+    /* Process the client address pre-read */
+    rc = wh_Server_DmaProcessClientAddress32(
+        server, clientAddr, &transformedAddr, len, WH_DMA_OPER_CLIENT_READ_PRE,
+        flags);
+    if (rc != WH_ERROR_OK) {
+        return rc;
+    }
+
+    /* Perform the actual copy */
+    /* TODO: should we add a flag to force client word-sized reads? */
```
billphipps, line 176: Nope.  if the client is asking for unaligned reads, then they can wait.

src/wh_server_nvm.c
```diff
@@ -262,20 +262,46 @@ int wh_Server_HandleNvmRequest(whServerContext* server,
...
-            /* TODO: Add hostaddr conversion/checking */
-            metadata =(void*)((intptr_t)req.metadata_hostaddr);
-            data = (void*)((intptr_t)req.data_hostaddr);
-            /* TODO: Add data_len checking */
+            /* perform platform-specific host address processing */
+            resp.rc = wh_Server_DmaProcessClientAddress32(
+                server, req.metadata_hostaddr, &metadata, sizeof(whNvmMetadata),
+                WH_DMA_OPER_CLIENT_READ_PRE, (whServerDmaFlags){0});
+            if (resp.rc != WH_ERROR_OK) {
+                goto transRespAddObjDma32;
```
billphipps, line 270: Yeah, not loving the goto.  I'd sooner refactor to add separate DMA handling functions to keep this function reasonably small.

wolfhsm/wh_server_dma.h
```diff
@@ -0,0 +1,107 @@
+#ifndef WH_SERVER_DMA_H_
+#define WH_SERVER_DMA_H_
+
+#include <stdint.h>
+#include <stddef.h>
+
+#include "wolfhsm/wh_server.h"
+
+#define WH_DMA_ADDR_ALLOWLIST_SIZE (10)
```
billphipps, line 9: Consider using NUM/COUNT instead of SIZE.  SIZE/LEN makes me think its the number of bytes, not elements.

billphipps (approved):
Looks great!  A few minor recommended tweaks, but ok as is.
1. Fine with void* and size_t as the base types.
2. Recommend exposing the address check functions publicly
3. Recommend that we break off the DMA handling functions in the server separately to keep things more manageable.
4. Not a fan of the goto's as these will trigger static analyzer.  Recommend to add backlog to rework these.

src/wh_server_dma.c
bigbrett: Right now the switch would have empty cases for READ_POST and WRITE_POST, as we don't do anything with those. Instead, I added `isOperValid()` as a check to this function and return failure if an invalid operation is requested. LMK what you think

src/wh_server_dma.c
bigbrett: It already works like this, albeit without calling a "default callback" (it just doesn't call a callback if NULL), and I don't allow you to explicitly register a NULL callback using this function. When initializing the server config struct, if the callback is left NULL that is exactly how it will behave. I think if we ever have logic that a default callback would need to do on every target that doesn't register one, we can definitely add it in, but right now it just does nothing.

wolfhsm/wh_server_dma.h
bigbrett: fixed

src/wh_server_dma.c
bigbrett: Fixed, and added test that NULL callbacks don't break things

wolfhsm/wh_server_dma.h
```diff
@@ -0,0 +1,114 @@
...
+#define WH_SERVER_DMA_H_
+
+#include <stdint.h>
+#include <stddef.h>
+
+#include "wolfhsm/wh_server.h"
+
+#define WH_DMA_ADDR_ALLOWLIST_COUNT (10)
+
+struct whServerContext_t;
```
billphipps, line 11: Isn't this already in wh_server.h?  Would prefer the typedef'ed one without struct and _t for all parameters below as well.

billphipps (approved):
Excellent!  Minor nit on using the non-typedefed name of struct whServerContext_t.  Feel free to do a find/replace if you want to or leave it as is.   Let me know when you want to merge

jpbland1 (approved):
Reviewed and tested, looks good

wolfhsm/wh_server_dma.h
bigbrett: AFAICT there needs to be a forward declaration in `wh_server_dma.h` too because of the bidirectional dependency between the two files. Specifically, `wh_server.h` includes `wh_server_dma.h` before defining the server context (which it must, as the server context has a member of type `whServerDmaContext`). So if that forward declaration goes away then you get undefined type errors in `wh_server_dma.h`. I played around with it a bit and the only other way I could get it to compile cleanly was to typedef `whServerContext` before including `wh_server_dma.h` (which is gross). I figured forward declaring it in `wh_server_dma.h` wasn't a big issue because there is already a bidirectional dependency between the two modules - `wh_server_dma.h` isn't an independent API, but rather just extends the server API but in a separate file for clarity. I think the real solution from an encapsulation perspective would just be to have this all in `wh_server.{c,h}`. Let me know if you think there is a better way to do it and I can revise.

PR #22, opened 2024-04-12
use key type and key user as part of the internal keyId

jpbland1:
splitting the key handle up into 3 parts internally, while presenting a one byte key id to the user, allows the same external or user facing key to be identical for each user value and between SHE and crypto keys. algorithm types were added as flags instead of being a part of the key handle and the user param was added to the client so a user can split operations between 16 clients. may not fully implement multiple client handling but seems good for keys

### DIFF
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -43,7 +43,6 @@ int wh_Client_Init(whClientContext* c, const whClientConfig* config)
     }
 
     memset(c, 0, sizeof(*c));
-    c->user = config->user;
 
     if (    ((rc = wh_CommClient_Init(c->comm, config->comm)) == 0) &&
 #ifndef WOLFHSM_NO_CRYPTO
@@ -85,8 +84,8 @@ int wh_Client_SendRequest(whClientContext* c,
     if (c == NULL) {
         return WH_ERROR_BADARGS;
     }
-    rc = wh_CommClient_SendRequest(c->comm, WH_COMM_MAGIC_NATIVE, kind, c->user,
-        &req_id, data_size, data);
+    rc = wh_CommClient_SendRequest(c->comm, WH_COMM_MAGIC_NATIVE, kind, &req_id,
+        data_size, data);
     if (rc == 0) {
         c->last_req_kind = kind;
         c->last_req_id = req_id;
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -657,4 +657,9 @@ void wh_Client_SetKeyCurve25519(curve25519_key* key, whNvmId keyId)
 {
     XMEMCPY(key->devCtx, (void*)&keyId, sizeof(keyId));
 }
+
+void wh_Client_SetKeyRsa(RsaKey* key, whNvmId keyId)
+{
+    XMEMCPY(key->devCtx, (void*)&keyId, sizeof(keyId));
+}
 #endif  /* WOLFHSM_NO_CRYPTO */
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -43,6 +43,7 @@ int wh_Client_Init(whClientContext* c, const whClientConfig* config)
     }
 
     memset(c, 0, sizeof(*c));
+    c->user = config->user;
 
     if (    ((rc = wh_CommClient_Init(c->comm, config->comm)) == 0) &&
 #ifndef WOLFHSM_NO_CRYPTO
@@ -84,10 +85,8 @@ int wh_Client_SendRequest(whClientContext* c,
     if (c == NULL) {
         return WH_ERROR_BADARGS;
     }
-
-    rc = wh_CommClient_SendRequest(c->comm,
-                WH_COMM_MAGIC_NATIVE, kind, &req_id,
-                data_size, data);
+    rc = wh_CommClient_SendRequest(c->comm, WH_COMM_MAGIC_NATIVE, kind, c->user,
+        &req_id, data_size, data);
     if (rc == 0) {
         c->last_req_kind = kind;
         c->last_req_id = req_id;
@@ -465,6 +464,7 @@ int wh_Client_KeyCacheRequest_ex(whClientContext* c, uint32_t flags,
     uint8_t* packIn = (uint8_t*)(&packet->keyCacheReq + 1);
     if (c == NULL || in == NULL || inSz == 0)
         return WH_ERROR_BADARGS;
+    packet->keyCacheReq.id = keyId;
     packet->keyCacheReq.flags = flags;
     packet->keyCacheReq.sz = inSz;
     if (label == NULL)
@@ -489,7 +489,7 @@ int wh_Client_KeyCacheRequest(whClientContext* c, uint32_t flags,
     uint8_t* label, uint32_t labelSz, uint8_t* in, uint32_t inSz)
 {
     return wh_Client_KeyCacheRequest_ex(c, flags, label, labelSz, in, inSz,
-        WOLFHSM_ID_ERASED);
+        WOLFHSM_KEYID_ERASED);
 }
 
 int wh_Client_KeyCacheResponse(whClientContext* c, uint16_t* keyId)
@@ -514,7 +514,7 @@ int wh_Client_KeyCacheResponse(whClientContext* c, uint16_t* keyId)
 int wh_Client_KeyEvictRequest(whClientContext* c, uint16_t keyId)
 {
     whPacket packet[1] = {0};
-    if (c == NULL || keyId == WOLFHSM_ID_ERASED)
+    if (c == NULL || keyId == WOLFHSM_KEYID_ERASED)
         return WH_ERROR_BADARGS;
     /* set the keyId */
     packet->keyEvictReq.id = keyId;
@@ -544,7 +544,7 @@ int wh_Client_KeyEvictResponse(whClientContext* c)
 int wh_Client_KeyExportRequest(whClientContext* c, uint16_t keyId)
 {
     whPacket packet[1] = {0};
-    if (c == NULL || keyId == WOLFHSM_ID_ERASED)
+    if (c == NULL || keyId == WOLFHSM_KEYID_ERASED)
         return WH_ERROR_BADARGS;
     /* set keyId */
     packet->keyExportReq.id = keyId;
@@ -597,7 +597,7 @@ int wh_Client_KeyExportResponse(whClientContext* c, uint8_t* label,
 int wh_Client_KeyCommitRequest(whClientContext* c, whNvmId keyId)
 {
     whPacket packet[1] = {0};
-    if (c == NULL || keyId == WOLFHSM_ID_ERASED)
+    if (c == NULL || keyId == WOLFHSM_KEYID_ERASED)
         return WH_ERROR_BADARGS;
     /* set keyId */
     packet->keyCommitReq.id = keyId;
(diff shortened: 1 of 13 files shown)


src/wh_server_keystore.c
```diff
@@ -241,24 +246,21 @@ int wh_Server_HandleKeyRequest(whServerContext* server,
...
+        /* get a new id if one wasn't provided, user has already been applied */
+        if (packet->keyCacheReq.id == WOLFHSM_KEYID_ERASED)
+            ret = hsmGetUniqueId(server, &meta->id);
         /* write the key */
         if (ret == 0)
             ret = hsmCacheKey(server, meta, in);
         if (ret == 0) {
-            packet->keyCacheRes.id = meta->id;
+            /* remove the user, client may set type */
+            packet->keyCacheRes.id = (meta->id & (~WOLFHSM_KEYUSER_MASK));
```
bigbrett, line 257: Should we remove type as well, given that it isn't set by the client? From the client perspective it would be weird to get back an ID that is different than the one you sent, in case a client were to use that ID for bookkeeping reasons.

bigbrett (dismissed):
A few things I'd like to see (here or in next PR):
- Helper functions/macros  for keyId bit manipulation/masking, as they are used lots of places in the code and easy to screw up
- Expanded testing (check all return values and invariants for one - I already pointed out the bitmasking bug with the keyIds earlier that wasn't caught by tests). Perhaps using other keyIds and different keys, ensuring they changes are propagated through

billphipps:
This is definitely getting close to what I was hoping.  Instead of hitting each file with repeated comments, I'm going to identify a few design misses that should be pretty quick to correct and some items that need to be made more clear before we move too much farther down the road.

Stuff that is close but must be corrected:
1. The client_id (called user in your code) is set in whClientConfig as a value between 0-0xF.  client_id==0 means anonymous/public/not set.  This will replace the client_id in whCommClient and whNvm, which will now need these to be passed down from the upper level.
2. The client_id is intended to be used as part of the whNvmId and server-side key Id to identify the user that owns the resource.  The client was expected to send the WH_MESSAGE_COMM_ACTION_INIT action after connecting so that the server context would store this as the client_id to be passed/masked with every request.  I see that I did not implement the client/server functions for this.  My bad.
3. The header of each message MUST remain at 8 bytes, so the "user" field will need to be removed and the server will simply keep the id for each of the components as a member like server->client_id, again set by the client after connecting.
4.  The client's perspective of the key identifier should simply be an identifier from 0x1-0xFE, where we reserve 0x00 and 0xFF as Invalid.  These id's are NOT unique per key type, so you can't have AES key 0x1 and RSA private key 0x1.  These are fully user-defined and are unique (to the server) per client_id.
5. The client's perspective of an NVM object is similar, in that each NVM object has a unique ID per client_id.  This means that a client's key id of 0x1 is distinct from a client's NvmId of 0x1.  The client will be able to see the NVM objects that store the key values when invoking the List command.
6. The NVM metadata design expected the client_id to be masked/passed/used as part of the whNvmAccess field (think of user/group/global Unix permissions), the object-type specific flags to be used in the whNvmFlags field, and the whNvmId to uniquely identify the owner, object number, and object type (key, counter, cert, SHE NVM data, ordinary data).

Here's some examples:
1. Client 0x4 wants to add an owned NVM object data[67] as id 0x55, with label "My Object".
whClient_Init();
whClient_SetClientId(0x4);  // Doesn't exist yet.  Use WH_MESSAGE_COMM_ACTION_INIT
char* label ="My Object";
int remote_rc = 0;
whClient_NvmAddObject(c, 0x55, WH_ACCESS_ANY, WH_FLAGS_ANY, 
                                               strlen(label), label, 67, data, &remote_rc);
The server gets this request and constructs a whNvmId = 0x4055, which is the client_id==0x4, object_type==0x0 (ordinary data), and num==0x55.  Note the server will deny any id's with the top nibble set as WH_ERROR_BADARGS.  The second nibble is allowed to be set here as long as it is a valid object type.  We should add explicit whClient_AddDataObject() that the user should be using to avoid any object-type snafus.

2. Client 0x5 wants to add an owned AES key at key[32] as key id 0x33 to NVM, with label "My Key".
whClient_Init();
whClient_SetClientId(0x5);
char* label = "My Key";
int remote_rc = 0;
/* Put key into RAM.  This function doesn't exist yet */
whClient_ImportKey(c, 0x33, WH_ACCESS_ANY, WH_FLAGS_KEY_AES,
                                          strlen(label), label, 32, key, &remote_rc);
Note: the type of the key is encoded in the flags metadata, rather than the ID.  In the key cache, the server will construct a whNvmId as 0x5133 with flags of WH_FLAGS_KEY_AES, as if it were already in the NVM.
/* Store the key into NVM */
whClient_CommitKey(c, 0x33, &remote_rc);
The server gets this request and constructs a whNvmId==0x5133, which is the client_id==0x5, object_type==0x1 (NVM key), and num== 0x33.  The flags holds the NVM key kind which is set by WH_FLAGS_KEY_AES.

3. Client 0x5 wants to Read an owned AES key with ID 0x33 from NVM back into RAM and use it with an AES context
whClient_Init();
whClient_SetClientId(0x5);
int remote_rc = 0;
whClient_ReadKey(c, 0x33, WH_ACCESS_ANY, WH_FLAGS_KEY_AES, &remote_rc);
The server gets this request and constructs a whNvmId==0x5133.  It reads the metadata and checks that the flags==WH_FLAGS_KEY_AES.  If not, it returns WH_ABORTED.  Note the client could have used WH_FLAGS_ANY to match any key type with the matching ID.
Aes aes = {0};
wc_AesInit_ex(aes, NULL, WH_CLIENT_DEVID);
whClient_SetAesKey(c, aes, 0x33);
The client modifies the aes->devCtx to encode it 0x33.  The server will construct the whNvmId and flags during the server side crypto message handling to match the unique id within the key cache of 0x5133.  The client is not aware of this encoding.

4. Client 0x3 wants to directly update the unowned SHE NVM key[16] as WH_SHE_KEY_4, rather than using the SHE update key protocol.  Note that all SHE keys are 16 bytes. 
whClient_Init();
whClient_SetClientId(0x0);  //Note that this is an unowned key so multiple clients can use it
int remote_rc = 0;
/* Put key into RAM.  Note WH_SHE_KEY_4 is hard-coded as WH_NVM_TYPE_SHE | WH_SHE_KEYID_4 = 0x0204.*/
whClient_ImportKey(c, WH_SHE_KEY_4, WH_ACCESS_ANY, WH_FLAGS_SHE_NVM,
                                          0, NULL, 16, key, &remote_rc);
In the key cache, the server will construct a whNvmId as 0x0204 with object type of 0x2 (rather than 0x1) and flags of WH_FLAGS_SHE_NVM.
/* Store the key into NVM */
whClient_CommitKey(c, WH_SHE_KEY_4, &remote_rc);
The server gets this request and constructs a whNvmId==0x0204, which is the client_id==0x0, object_type==0x2 (SHE key), and num== 0x04.  The flags holds the NVM key kind which is set by WH_FLAGS_SHE_NVM.

Hopefully this will better show how the whNvmId and whNvmFlags were intended to be used.  Also, it should point out that there should be a distinct whKeyId type passed in and out of the key functions that don't need the client_id encoded in the top nibble because it is the server that actually handles that.  

FLAGS_KEY_* should include: AES, RSA_PRIVATE, RSA_PUBLIC, ECC_PRIVATE, ECC_PUBLIC, HMAC.  I haven't worked through how we should tag KDF'ed keys yet, but we'll get there.

Can you update your code to put the key types into the flags field and switch the user to client_id and make it a member of the server context?  I can craft a quickie that provides the whClient_SetClientId() functions if you'd like.

src/wh_client.c
```diff
@@ -658,4 +657,9 @@ void wh_Client_SetKeyCurve25519(curve25519_key* key, whNvmId keyId)
 {
     XMEMCPY(key->devCtx, (void*)&keyId, sizeof(keyId));
 }
+
+void wh_Client_SetKeyRsa(RsaKey* key, whNvmId keyId)
+{
+    XMEMCPY(key->devCtx, (void*)&keyId, sizeof(keyId));
```
billphipps, line 663: This will likely not set the entire devCtx value since whNvmId is probably a smaller size than void*.  Recommend to use uintptr_t to let the compiler promote the unsigned integer up to the proper size.

Like: key->devCtx = (void*) ((uintptr_t)keyId);

src/wh_server_crypto.c
```diff
@@ -56,6 +57,7 @@ static int hsmLoadKeyRsa(whServerContext* server, RsaKey* key, whKeyId keyId)
     int slotIdx = 0;
     uint32_t idx = 0;
     uint32_t size;
+    keyId |= (WOLFHSM_KEYTYPE_CRYPTO | (server->comm->client_id << 8));
```
billphipps, line 60: Probably should be a macro or function as I suspect we'll need this a lot.

src/wh_server_keystore.c
```diff
@@ -14,37 +14,40 @@
...
             break;
     }
     /* unlikely but cover the case where we've run out of ids */
-    if (id >= WOLFHSM_KEYID_MASK)
+    if (id > WOLFHSM_KEYID_MASK)
         ret = WH_ERROR_NOSPACE;
     /* ultimately, return found id */
     if (ret == 0)
-        ret = (id | WOLFHSM_KEYID_CRYPTO);
+        *outId |= buildId;
```
billphipps, line 50: Either check the parameter for NULL at the beginning and return WH_ERROR_BADARGS or gate it here.

src/wh_server_keystore.c
```diff
@@ -14,37 +14,40 @@
...
         for (i = 0; i < WOLFHSM_NUM_RAMKEYS; i++) {
-            if ((id | WOLFHSM_KEYID_CRYPTO) == server->cache[i].meta->id)
+            if (buildId == server->cache[i].meta->id)
                 break;
         }
         /* try again if match */
         if (i < WOLFHSM_NUM_RAMKEYS)
             continue;
         /* if keyId exists */
         ret = wh_Nvm_List(server->nvm, WOLFHSM_NVM_ACCESS_ANY,
```
billphipps, line 38: Recommend to simply use wh_Nvm_GetMetadata() != WH_ERROR_NOTFOUND since you are only looking for an exact match.

test/wh_test_crypto.c
```diff
@@ -116,7 +116,83 @@ int whTest_CryptoClientConfig(whClientConfig* config)
         WH_ERROR_PRINT("KEY CACHE/EXPORT FAILED TO MATCH\n");
         goto exit;
     }
-    /* test evict */
+    /* test cache with duplicate keyId for a different user */
+    WH_TEST_RETURN_ON_FAIL(wh_Client_CommClose(client));
+    client->comm->client_id = 2;
```
billphipps, line 121: This shouldn't work this way.  The server should be applying the client id that it stored after the client invoked wh_Client_CommInit(c);   I understand this is for testing only, but I don't see how comm->client_id is being used at all in the code.  Are we sure this is working the way you expected?

test/wh_test_crypto.c
```diff
@@ -340,6 +417,16 @@ int whTest_CryptoServerConfig(whServerConfig* config)
             break;
         }
         wh_Server_GetConnected(server, &am_connected);
+        /* keep alive for 2 user changes */
+        if (am_connected != WH_COMM_CONNECTED && userChange < 2) {
+            if (userChange == 0)
+                server->comm->client_id = 2;
```
billphipps, line 423: I would remove this and simply use the wh_Client_CommInit function to update the server-side state.

billphipps (approved):
Looks really good!  A few nits, but I'd like to merge it as is and we'll fix a few things in the next round.  Only bug I see is not using the client_id correctly, but I'll fix as I go.

PR #23, opened 2024-04-18
Add connect/disconnect, reduce copy, simplify headers

billphipps:
A number of changes:
1. Adds server-side external connect/disconnect handlers
2. Adds server transport callbacks that can trigger a connect or disconnect from the transport layer
3. Adds CommInit message to exchange client and server id's
4. Adds CommClose message which will trigger server to disconnect
5. Updated tests to now use CommClose to close servers async
6. Consolidated wh_transport.h into wh_comm.h to avoid circular includes
7. Consolidated wh_server_dma.h into wh_server.h to avoid circular includes
8. Eliminated declaring counter variables within for loops
9. Eliminated // comments.  Now compiles clean for C90
10. Reordered/typed whServerContext and whClientContext to ensure packet buffer is 64-bit aligned
11. Exposed data ptr into internal buffer for CommServer and CommClient to eliminate extra copies and space
12. Reworked server handler to use internal buffer

Nearly all of these changes are necessary to make the Bernini and Aurix ports compile clean and align packet buffers correctly as well as use optimized memcpy.


wolfhsm/wh_comm.h
```diff
@@ -32,7 +32,8 @@
 enum {
     WH_COMM_HEADER_LEN = 8,    /* whCommHeader */
     WH_COMM_DATA_LEN = 1280,
-    WH_COMM_MTU = (WH_COMM_HEADER_LEN + WH_COMM_DATA_LEN)
+    WH_COMM_MTU = (WH_COMM_HEADER_LEN + WH_COMM_DATA_LEN),
+    WH_COMM_MTU_8 = (WH_COMM_MTU + 7) / 8,
```
jpbland1, line 36: I think we should call it `WH_COMM_MTU_BITS`, wolfSSL defines things as bits when referring to sizes in bits: `#define SP_INT_BITS 4096`

wolfhsm/wh_comm.h
```diff
@@ -97,17 +145,17 @@ typedef struct {
...
-    uint8_t packet[WH_COMM_MTU];
     whCommHeader* hdr;
     uint8_t* data;
     uint32_t client_id;
     uint32_t server_id;
     int initialized;
+    uint16_t reqid;
+    uint16_t seq;
+    uint16_t size;
 } whCommClient;
```
jpbland1, line 159: I think we should combine the client and server comm types and functions since they're so similar, really seems like a waste of code size

wolfhsm/wh_server.h
```diff
@@ -66,22 +126,53 @@ typedef struct whServerConfig_t {
...
+    whServerDmaContext dma;
+    int connected;
+};
+
+
+/** Public server context functions */
+
 /* Initialize the comms and crypto cache components.
- * Note: NVM and Crypto must be initialized prior to Server Init
+ * Note: NVM and Crypto components must be initialized prior to Server Init
```
jpbland1, line 145: Still pretty vauge, maybe `hardware drivers`

test/wh_test_crypto.c
```diff
@@ -253,6 +253,9 @@ int whTest_CryptoClientConfig(whClientConfig* config)
         WH_ERROR_PRINT("CURVE25519 shared secrets don't match\n");
     }
 
+    /* Tell server to close */
+    wh_Client_CommClose(client);
```
jpbland1, line 257: Should this be called inside of wh_Client_Cleanup? Users should be able to call it manually to reset just the connection but it seems convenient to also add it to cleanup

jpbland1 (changes requested):

wolfhsm/wh_comm.h
billphipps: Good idea.  At the end of the day, the callbacks are distinct, but there's probably a common struct and accessors that we can combine to handle the buffering and comm states, even if there are extra items that aren't used in the server side.  That would simplify the code and reduce cut-n-paste errors.  The send and recv funcs are nearly identical, except for that actual callback to the transport.  Lemme take a peek at this, as I was wanting to split the comms into wh_comm.h, wh_commclient.h, and wh_commserver.h, as well as the C files to improve isolation.  I'll add this to the backlog.

wolfhsm/wh_server.h
billphipps: Server has multiple components: comm, nvm, crypto, customcb, and eventually SHE and PKCS11, so these aren't necessarily drivers.  We need to ensure that if these pointers are NULL or not initialized, that the server gracefully handles this case.  Added to backlog.

wolfhsm/wh_comm.h
billphipps: Hah!  This is actually the rounded-up number of 8-byte uint64_t's needed to hold an MTU packet.  How about I change this to WH_COMM_MTU_64?

test/wh_test_crypto.c
billphipps: This is an async request that we are just busy spinning waiting for it to complete.  wh_Client_Cleanup is expected to synchronously clean up local resources, which could also have the transport zero-out/cancel any requests.  So, this must be outside of the cleanup call.

src/wh_server.c
```diff
@@ -84,16 +85,73 @@ int wh_Server_Cleanup(whServerContext* server)
...
+{
+    if (server == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    server->connected = connected;
+    return WH_ERROR_OK;
+}
+
+int wh_Server_SetConnectedCb(void* s, int connected)
```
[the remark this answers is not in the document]
billphipps: I had a common grumble on this.  The problem is that shared-memory platforms need to use some mechanism outside of the transport to indicate connected/disconnected because it's not necessary safe to use the shared-memory before it is configured.  So, the server/client context is the one that needs to maintain this upper-level connectedness, but the lower-level may (in some circumstances) be in the best position to detect connects and disconnects.  Notice in the Posix TCP transport implementation, we ALWAYS set connected initially because the connection sequence is handled during the Recv function, which will only be called if the upper level thinks its connected.  Chicken/egg.  The disconnect is automatically set by the send/recv functions when it notices the socket is no longer connected.  I did not wire in the client side yet as I'm not convinced we actually need this on the client side because the server won't disconnect on purpose and the client is the one the instantiates the connection/shared-memory so there's never an issue here.  

As for multiple connections, recall that each client/server context handles only 1 connection at a time.  The server app instantiations multiple server contexts, which share NVM and other singleton components.

src/wh_comm.c
```diff
@@ -276,6 +301,15 @@ int wh_CommServer_SendResponse(whCommServer* context,
     return rc;
 }
 
+uint8_t* wh_CommServer_GetDataPtr(whCommServer* context)
```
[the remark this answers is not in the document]
billphipps: Added to backlog to refactor to share common buffering/state code.

wolfhsm/wh_comm.h
jpbland1: ah gotcha yeah thats much less confusing, I'd also add a comment

test/wh_test_crypto.c
```diff
@@ -273,77 +276,28 @@ int whTest_CryptoServerConfig(whServerConfig* config)
...
         return WH_ERROR_BADARGS;
     }
 
     WH_TEST_RETURN_ON_FAIL(wh_Server_Init(server, config));
 
-    /* handle client rng */
-    do {
+    for (   wh_Server_SetConnected(server, am_connected);
+            am_connected == WH_COMM_CONNECTED;
+            wh_Server_GetConnected(server, &am_connected) ){
```
bigbrett, line 289: this is fun

test/wh_test_clientserver.c
```diff
@@ -1235,6 +1236,7 @@ int whTest_ClientCfg(whClientConfig* clientCfg)
     WH_TEST_ASSERT_RETURN(server_rc == WH_ERROR_OK);
     WH_TEST_ASSERT_RETURN(avail_objects == NF_OBJECT_COUNT);
 
+    wh_Client_CommClose(client);
```
bigbrett, line 1239:
```suggestion
    WH_TEST_RETURN_ON_FAIL(wh_Client_CommClose(client));
```

bigbrett (dismissed):

wolfhsm/wh_comm.h
```diff
@@ -73,18 +74,65 @@ enum {
...
+
+/* Provide a callback to invoke when the transport can detect a connect or a
+ * disconnect */
+typedef int (*whCommSetConnectedCb)(void* context, int connected);
+
+/* Status of whether a client is connected or not */
+enum {
+    WH_COMM_DISCONNECTED = 0,
+    WH_COMM_CONNECTED = 1,
+};
```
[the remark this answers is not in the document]
billphipps: Concur.  I was trying to keep the callbacks generic and untyped, but there's no reason not to force the type to an enum.  Fixed in next push.

test/wh_test_crypto.c
billphipps: Prolly too fancy.  I'll switch it to a while structure. Fixed in next push

test/wh_test_clientserver.c
billphipps: Fixed in next push.

wolfhsm/wh_comm.h
billphipps: went with WH_COMM_MTU_U64_COUNT with a comment.  Fixed in next push.

jpbland1 (approved):
Looks good

PR #25, opened 2024-04-18
No crypto and padding

billphipps:
1. Support for doing a no crypto build and enabling -Wpadded on all of wolfHSM.
2. Adds additional workflow action to run this build
3. Corrects some wolfCrypt conditional compile misses.  Note WOLFSSL_KEY_GEN is required on server for now.
4. Reorders structures and adds additional padding bytes.

### DIFF
--- a/src/wh_server.c
+++ b/src/wh_server.c
@@ -47,9 +47,10 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
     }
 
     memset(server, 0, sizeof(*server));
-    server->crypto = config->crypto;
     server->nvm = config->nvm;
 
+#ifndef WOLFHSM_NO_CRYPTO
+    server->crypto = config->crypto;
     if (server->crypto != NULL) {
 #if defined(WOLF_CRYPTO_CB)
         server->crypto->devId = config->devId;
@@ -57,6 +58,7 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
         server->crypto->devId = INVALID_DEVID;
 #endif
     }
+#endif
 
     rc = wh_CommServer_Init(server->comm, config->comm_config,
             wh_Server_SetConnectedCb, (void*)server);
@@ -254,6 +256,7 @@ int wh_Server_HandleRequestMessage(whServerContext* server)
                     size, data, &size, data);
         break;
 
+        #ifndef WOLFHSM_NO_CRYPTO
         case WH_MESSAGE_GROUP_KEY:
             rc = wh_Server_HandleKeyRequest(server, magic, action, seq,
                     data, &size);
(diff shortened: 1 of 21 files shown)


src/wh_server.c
```diff
@@ -254,6 +256,7 @@ int wh_Server_HandleRequestMessage(whServerContext* server)
                     size, data, &size, data);
         break;
 
+        #ifndef WOLFHSM_NO_CRYPTO
```
bigbrett, line 259: indentation doesn't match the `#endif`

test/Makefile
```diff
@@ -47,13 +47,19 @@ CFLAGS += -fsanitize=address
...
 endif
 
+# Support a NO CRYPTO build
+ifeq ($(NOCRYPTO),1)
+CFLAGS += -DWOLFHSM_NO_CRYPTO -Wpadded
+endif
+
 # wolfHSM-specific defines
-CFLAGS += -DWH_CONFIG
+CFLAGS += -DWH_CONFIG -DWOLFHSM_NO_CRYPTO
```
bigbrett, line 56: I assume this should only be defined for the NOCRYPTO option right? 

```suggestion
CFLAGS += -DWH_CONFIG
```

wolfhsm/wh_common.h
```diff
@@ -80,7 +80,7 @@ typedef struct {
 
 
 /** Manifest storage  */
-
+#if 0
```
bigbrett, line 83: did you mean to leave this in?

bigbrett (changes requested):
Looks good outside of a few potential oversights. So we need to be padding ALL structs now, not just "wire" structs? Or is that only because you want `-Wpadding` on?

src/wh_server.c
billphipps: Fixed.

test/Makefile
billphipps: Gaah!! Fixed.

wolfhsm/wh_common.h
billphipps: This was NE's first struct that we talked about.  I'll remove it.

billphipps:
The -Wpadded feature is double-edged as it checks for internal padding in the struct (which we should fix by rearranging members and adding growth/padding fields) as well as the overall struct size in terms of struct alignment in case this was put into an array.  Internal padding should absolutely be corrected.  External padding should be explicit when the struct is intended to be in an array or as a member of another struct.  I'm good with leaving external padding just to 32-bits, but I'd rather make everything 64-bit sized, meaning it will pass -Werror -Wpadded NO_CRYPT (because there's a LOT of work to make that fully aligned).

bigbrett (approved):

PR #27, opened 2024-05-02
She extension update

jpbland1:
added she functions, tested working with test vectors and the custom uid. decided to use the normal nvm functions to set she keys as it is simple. I think this should be the default way of programming the master ecu key, secret key etc. currently doesn't use the crypto struct for aes and cmac because those PRs are not merged yet. missing the load plain key command and test for export ram key, the load plain key command from my first draft needs to be re written due to how the new keyId system works.


jpbland1:
the asan test is failing due to a stack-buffer-overflow in the nvm code. I changed the metadata len to 36 after I added the count parameter which is a uint32_t, 4 bytes. does something else need to be updated when that const changes that I'm not aware of?

test/wh_test_she.c
```diff
@@ -0,0 +1,385 @@
...
+    /* store the secret key */
+    if ((ret = wh_Client_NvmAddObject(client, MAKE_WOLFHSM_KEYID(WOLFHSM_KEYTYPE_SHE, client->comm->client_id, WOLFHSM_SHE_SECRET_KEY_ID), 0, 0, 0, NULL, sizeof(secretKey), secretKey, &outRc)) != 0 || outRc != 0) {
+        WH_ERROR_PRINT("Failed to wh_Client_NvmAddObjectRequest %d\n", ret);
+        goto exit;
+    }
+    /* store the prng seed */
+    if ((ret = wh_Client_NvmAddObject(client, MAKE_WOLFHSM_KEYID(WOLFHSM_KEYTYPE_SHE, client->comm->client_id, WOLFHSM_SHE_PRNG_SEED_ID), 0, 0, 0, NULL, sizeof(prngSeed), prngSeed, &outRc)) != 0 || outRc != 0) {
+        WH_ERROR_PRINT("Failed to wh_Client_NvmAddObjectRequest %d\n", ret);
+        goto exit;
+    }
```
bigbrett, line 155: can we add convenience functions for these under the client SHE API?

test/wh_test_she.c
```diff
@@ -0,0 +1,385 @@
...
+    /* set the she uid */
+    if ((ret = wh_Client_SheSetUid(client, sheUid, sizeof(sheUid))) != 0) {
+        WH_ERROR_PRINT("Failed to wh_Client_SheSetUid %d\n", ret);
+        goto exit;
+    }
+    /* verify bootloader */
+    if ((ret = wh_Client_SheSecureBoot(client, bootloader, bootloaderSz)) != 0) {
```
bigbrett, line 162: can we add a negative test to know that the validation works and wont validate everything?

test/wh_test_she.c
```diff
@@ -0,0 +1,385 @@
...
+#include "port/posix/posix_transport_tcp.h"
+#include "port/posix/posix_flash_file.h"
+#endif
+
+#if defined(WH_CFG_TEST_POSIX)
+#include <unistd.h> /* For sleep */
+#include <pthread.h> /* For pthread_create/cancel/join/_t */
+#include "port/posix/posix_transport_tcp.h"
+#include "port/posix/posix_flash_file.h"
+#endif
```
bigbrett, line 42: can we condense into just one #if defined() block?

bigbrett (changes requested):
Overall it looks great, though I haven't dug into the SHE spec to verify adherence. A few nits, including adding a few more negative tests to ensure things fail when they should fail. But yeah, first pass LGTM

src/wh_server_she.c
```diff
@@ -119,173 +127,90 @@ static int wh_AesMp16(WOLFHSM_CTX* ctx, byte* in, word32 inSz,
...
+        action != WOLFHSM_SHE_SECURE_BOOT_FINISH &&
+        action != WOLFHSM_SHE_GET_STATUS &&
+        action != WH_SHE_SET_UID)) ||
+        (action != WH_SHE_SET_UID && hsmSheUidSet == 0)) {
+        packet->rc = WH_SHE_ERC_SEQUENCE_ERROR;
+        *size = WOLFHSM_PACKET_STUB_SIZE + sizeof(packet->rc);
+        return 0;
+    }
+
+    switch (action)
```
bigbrett, line 187: I think we should have handler functions for each individual case - this function is getting pretty gnarly

src/wh_client_she.c
```diff
@@ -241,6 +250,51 @@ int wh_Client_SheLoadKey(whClientContext* c, uint8_t* messageOne,
     return ret;
 }
 
+int wh_Client_SheLoadPlainKeyRequest(whClientContext* c, uint8_t* key,
+    uint32_t keySz)
+{
+    int ret;
+    whPacket* packet;
+    if (c == NULL || key == NULL || keySz < WOLFHSM_SHE_KEY_SZ)
```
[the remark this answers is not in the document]
jpbland1: they might just be using some random buffer and calling sizeof on it, just needs to be large enough

src/wh_client_she.c
```diff
@@ -439,3 +490,249 @@ int wh_Client_SheExtendSeed(whClientContext* c, uint8_t* entropy,
...
+    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ)
+        return WH_ERROR_BADARGS;
+    packet = (whPacket*)wh_CommClient_GetDataPtr(c->comm);
+    /* out is after fixed sized fields */
+    packOut = (uint8_t*)(&packet->sheEncEcbRes + 1);
+    ret = wh_Client_RecvResponse(c, &group, &action, &dataSz, (uint8_t*)packet);
+    if (ret == 0) {
+        if (packet->rc != WOLFHSM_SHE_ERC_NO_ERROR)
+            ret = packet->rc;
+        else if (sz < packet->sheEncEcbRes.sz)
```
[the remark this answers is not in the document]
jpbland1: I just based it on how we do aes in wolfCrypt, see https://www.wolfssl.com/doxygen/group__AES.html#gaed1e38cd30d917165183fc68dd4b218b, usually the same buffer is used and overwritten for the output anyways.

src/wh_she_common.c
```diff
@@ -0,0 +1,217 @@
...
+ *
+ */
+/* System libraries */
+
+#include <stdint.h>
+#include <stdlib.h>  /* For NULL */
+#include <string.h>  /* For memset, memcpy */
+
+/* TODO replace with our own to host function */
+#include <arpa/inet.h>
```
bigbrett, line 30: We can't merge this until you fix this TODO. We can't have `arpa/inet.h` on any of our real targets

test/Makefile
```diff
@@ -108,6 +108,7 @@ ifeq ($(SHE),1)
 SRC_C += \
             $(WOLFHSM_DIR)/src/wh_client_she.c \
             $(WOLFHSM_DIR)/src/wh_server_she.c \
+            $(WOLFHSM_DIR)/src/wh_she_common.c \
```
bigbrett, line 111: trailing slash, while okay now, might lead to a cryptic Make error later on

src/wh_she_common.c
```diff
@@ -0,0 +1,217 @@
...
+    /* free aes for protection */
+    wc_AesFree(aes);
+    return ret;
+}
+
+int wh_SheGenerateLoadableKey(uint8_t keyId,
+    uint8_t authKeyId, uint32_t count, uint32_t flags, uint8_t* uid,
+    uint8_t* key, uint8_t* authKey, uint8_t* messageOne, uint8_t* messageTwo,
+    uint8_t* messageThree, uint8_t* messageFour, uint8_t* messageFive)
+{
```
bigbrett, line 101: Null check args since this is a public function

test/wh_test_she.c
```diff
@@ -82,10 +82,10 @@ int whTest_SheClientConfig(whClientConfig* config)
         0x3d, 0x7e, 0x11, 0x73, 0x93, 0x17, 0x2a};
     uint8_t zeros[WOLFHSM_SHE_BOOT_MAC_PREFIX_LEN] = {0};
     uint8_t bootloader[512];
-    uint8_t keyAndDigest[32] = {0};
+    uint8_t bootMacDigest[16] = {0};
```
bigbrett, line 85: do we have macros for these array sizes?

bigbrett (approved):
As always, would like to see some more expanded testing for these features. At least once negative test case would be good to ensure things fail when we need them to fail. Would go a long way towards being more confident that further code modifications don't introduce regerssions

bigbrett (changes requested):
sorry, prior approval was premature - only blocker is removing `arpa/inet.h`. Can't merge that in

src/wh_client_she.c
```diff
@@ -439,3 +490,249 @@ int wh_Client_SheExtendSeed(whClientContext* c, uint8_t* entropy,
...
     return ret;
 }
+
+int wh_Client_SheEncEcbRequest(whClientContext* c, uint8_t keyId, uint8_t* in,
+    uint32_t sz)
+{
+    int ret;
+    uint8_t* packIn;
+    whPacket* packet;
+    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ)
```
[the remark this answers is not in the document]
bigbrett: I still think there are cases here where we need to validate the size. The size the user passes is directly passed to `memcpy` without any upper bound checking, so while it might be considered abuse of the function, it is still a potential buffer overflow. Yes it is on the client side and so not any real security implications, but still, I don't think we should allow the client to overflow a buffer in one of our public APIs if we can avoid it. We should look at all the size arguments passed to these APIs and ensure they aren't used directly without validation.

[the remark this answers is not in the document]
jpbland1: That's not how it's done in any of our crypto for aes, the ecb encrypt and decrypt functions don't even check minimum length:
```
        for (i = 0; i < sz; i += AES_BLOCK_SIZE) {
            ret = wc_AesEncryptDirect(aes, out, in);
            if (ret != 0)
                break;
            in += AES_BLOCK_SIZE;
            out += AES_BLOCK_SIZE;
        }
```
when the data is longer than the packet buffer, assuming we're not using shared memory, we'll need a different scheme that lets the client send multiple packets for a single command, and that's not in the scope of this change

bigbrett (approved):
Good on my end. @billphipps please review

src/wh_server_she.c
```diff
@@ -61,34 +57,65 @@ enum WOLFHSM_SHE_SB_STATE {
...
+static uint8_t hsmSheRamKeyPlain = 0;
+static uint8_t hsmSheUidSet = 0;
+static uint32_t hsmSheBlSize = 0;
+static uint32_t hsmSheBlSizeReceived = 0;
+static uint32_t hsmSheRndInited = 0;
 /* cmac is global since the bootloader update can be called multiple times */
 Cmac sheCmac[1];
+Aes sheAes[1];
+
+static int isLittleEndian() {
```
billphipps, line 72: Consider using #ifdef __ BYTE_ORDER__ to see if it is already defined and simplify this test to a compile-time test.

src/wh_server_she.c
```diff
@@ -61,34 +57,65 @@ enum WOLFHSM_SHE_SB_STATE {
...
+    unsigned int x = 1; /* 0x00000001 */
+    char *c = (char*)&x;
+    return (int)*c;
+}
+
+/* Converts a 32-bit value from host to network byte order */
+static uint32_t htonl(uint32_t hostlong) {
+    if (isLittleEndian()) {
+        return ((hostlong & 0x000000FF) << 24) | ((hostlong & 0x0000FF00) << 8)
+            | ((hostlong & 0x00FF0000) >> 8) | ((hostlong & 0xFF000000) >> 24);
```
billphipps, line 82: Consider using the equally ugly functions in wh_comm.h.  We need to move to a utility library at some point.  low priority,

src/wh_server_she.c
```diff
@@ -61,34 +57,65 @@ enum WOLFHSM_SHE_SB_STATE {
...
+static uint8_t hsmSheSbState = WOLFHSM_SHE_SB_INIT;
+static uint8_t hsmSheCmacKeyFound = 0;
+static uint8_t hsmSheRamKeyPlain = 0;
+static uint8_t hsmSheUidSet = 0;
+static uint32_t hsmSheBlSize = 0;
+static uint32_t hsmSheBlSizeReceived = 0;
+static uint32_t hsmSheRndInited = 0;
 /* cmac is global since the bootloader update can be called multiple times */
 Cmac sheCmac[1];
+Aes sheAes[1];
```
billphipps, line 70: Collect all required she server state into a single struct that is allocated/provided at the same time as the server.  Note that this state is per server instance and NOT global, but the server configs could use the same SHE instance to provide the same state, if desired.

src/wh_server_she.c
```diff
@@ -16,43 +16,39 @@
...
 #include "wolfssl/wolfcrypt/cmac.h"
-#include "wolfssl/wolfcrypt/kdf.h"
-#include "wolfssl/wolfcrypt/cryptocb.h"
 
 #include "wolfhsm/wh_server.h"
+#include "wolfhsm/wh_server_keystore.h"
+#include "wolfhsm/wh_message.h"
 #include "wolfhsm/wh_packet.h"
+#include "wolfhsm/wh_error.h"
+#ifdef WOLFHSM_SHE_EXTENSION
```
billphipps, line 40: Shouldn't this entire file be protected with this macro?

src/wh_server_keystore.c
```diff
@@ -197,8 +200,9 @@ int hsmReadKey(whServerContext* server, whKeyId keyId, whNvmMetadata* outMeta,
     int i;
     whNvmMetadata meta[1] = {0};
     /* make sure id is valid */
-    if (server == NULL || (keyId & WOLFHSM_KEYID_MASK) == WOLFHSM_KEYID_ERASED
-        || outSz == NULL) {
+    if (server == NULL || ((keyId & WOLFHSM_KEYID_MASK) == WOLFHSM_KEYID_ERASED
+        && (keyId & WOLFHSM_KEYTYPE_MASK) != WOLFHSM_KEYTYPE_SHE) ||
```
billphipps, line 204: Does this need to be protected by the compile time ifdef for she_extension?

src/wh_server_keystore.c
```diff
@@ -239,11 +243,18 @@ int hsmReadKey(whServerContext* server, whKeyId keyId, whNvmMetadata* outMeta,
...
+        *outSz = WOLFHSM_SHE_KEY_SZ;
+        if (outMeta != NULL) {
+            /* need empty flags and corect lenth and id */
+            XMEMSET(outMeta, 0, sizeof(meta));
+            meta->len = WOLFHSM_SHE_KEY_SZ;
+            meta->id = keyId;
+        }
         ret = 0;
     }
 #endif
```
billphipps, line 260: Shouldn't this be handled in the SHE version of read key and not in the generic one?

src/wh_client_she.c
```diff
@@ -0,0 +1,879 @@
...
+    return ret;
+}
+
+int wh_Client_SheDecCbcRequest(whClientContext* c, uint8_t keyId, uint8_t* iv,
+    uint32_t ivSz, uint8_t* in, uint32_t sz)
+{
+    int ret;
+    uint8_t* packIn;
+    whPacket* packet;
+    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ || iv == NULL ||
```
billphipps, line 706: Not sure I understand why the input size is limited to the key size which is the same as the block size.  Does the user have to repeatedly call this function with each block and make sure the iv is pointed to the previous cipher text?

billphipps (commented):
Looks good overall.  A few things we need to get correct, but I'll merge it now if you want to.
1. AES CBC and CMAC sizes larger than a single block must be supported.  recommend to use DMA for these and simply pass a pointer when available.  Otherwise, add a client-side loop to perform the iterative process.
2. Collect SHE state into a single struct and associate that with a server context.  The same state can be used by multiple contexts, or not.  It's not necessarily global.
3. Separate SHE stuff into different files/headers to minimize logic spread into common functions.  SHE should appear to be an add-on or helper to use the base HSM library.  I expect to code up the PKCS11 as a layer on top, hopefully with no internal logic aside from ID segregation.

A few nits and questions throughout.  I still don't see a consistent whKeyId vs whNvmId use yet, but we'll get there.  Excellent job on the test code!  We may want to have a fixed known answer test in there before you generate everything randomly.

src/wh_she_common.c
```diff
@@ -0,0 +1,222 @@
...
+        /* add WOLFHSM_SHE_KEY_UPDATE_MAC_C to the input */
+        XMEMCPY(kdfInput + WOLFHSM_SHE_KEY_SZ, WOLFHSM_SHE_KEY_UPDATE_MAC_C,
+            sizeof(WOLFHSM_SHE_KEY_UPDATE_MAC_C));
+        /* generate K4 */
+        ret = wh_AesMp16(kdfInput,
+            WOLFHSM_SHE_KEY_SZ + sizeof(WOLFHSM_SHE_KEY_UPDATE_MAC_C), tmpKey);
+    }
+    /* cmac messageFour using K4 as the cmac key */
+    if (ret == 0) {
+        ret = wc_InitCmac_ex(cmac, tmpKey, WOLFHSM_SHE_KEY_SZ, WC_CMAC_AES,
```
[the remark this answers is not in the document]
jpbland1: oneshot adds it's own cmac struct to the stack, want to re-use one copy and eventually get she using the crypto struct once everything is merged in

src/wh_client_she.c
jpbland1: < WOLFHSM_SHE_KEY_SZ, must be at least one block size, multiple blocks are fine

src/wh_server_keystore.c
jpbland1: multiple places can read the unset ecu key, they don't call their own she read function they make a she keyId and then call the normal read function, putting it here means no duplicated code

src/wh_server_keystore.c
jpbland1: I don't think so, the she keyId stuff is not compile guarded since it's reserved in case they update to use she key later but still want the same keystore

src/wh_she_common.c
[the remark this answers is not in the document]
jpbland1: wc_AesCmacGenerate_ex takes a struct, will switch to using that

wolfhsm/wh_server_she.h
```diff
@@ -16,12 +16,27 @@
...
+    WOLFHSM_SHE_SECURE_BOOT_INIT,
+    WOLFHSM_SHE_SECURE_BOOT_UPDATE,
+    WOLFHSM_SHE_SECURE_BOOT_FINISH,
+    WOLFHSM_SHE_GET_STATUS,
+    WOLFHSM_SHE_LOAD_KEY,
+    WOLFHSM_SHE_EXPORT_RAM_KEY,
+    WOLFHSM_SHE_INIT_RNG,
+    WOLFHSM_SHE_RND,
+    WOLFHSM_SHE_EXTEND_SEED,
+};
```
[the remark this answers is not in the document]
jpbland1: yes these are redundant/wrong

jpbland1:
Adressed most of bills comments, reorganized the structs and changed the way the "non-loadable" she keys are loaded into the hsm

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+    whPacket* packet;
+    if (c == NULL || entropy == NULL || entropySz != WOLFHSM_SHE_KEY_SZ)
+        return WH_ERROR_BADARGS;
+    packet = (whPacket*)wh_CommClient_GetDataPtr(c->comm);
+    /* set entropy */
+    memcpy(packet->sheExtendSeedReq.entropy, entropy,
+        sizeof(packet->sheExtendSeedReq.entropy));
+    /* send init rng req */
+    ret = wh_Client_SendRequest(c, WH_MESSAGE_GROUP_SHE, WH_SHE_EXTEND_SEED,
+        WOLFHSM_PACKET_STUB_SIZE, (uint8_t*)packet);
```
billphipps, line 488:
```suggestion
        WOLFHSM_PACKET_STUB_SIZE + sizeof(packet->sheExtendSeedReq), (uint8_t*)packet);
```

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+    return ret;
+}
+
+int wh_Client_SheEncEcbRequest(whClientContext* c, uint8_t keyId, uint8_t* in,
+    uint32_t sz)
+{
+    int ret;
+    uint8_t* packIn;
+    whPacket* packet;
+    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ)
```
billphipps, line 527: Better error out if it can't be placed in a single packet.
```suggestion
    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ || sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheEncEcbReq))
```

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+
+int wh_Client_SheEncEcbResponse(whClientContext* c, uint8_t* out, uint32_t sz)
+{
+    int ret;
+    uint16_t group;
+    uint16_t action;
+    uint16_t dataSz;
+    uint8_t* packOut;
+    whPacket* packet;
+    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ)
```
billphipps, line 551:
```suggestion
    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ|| sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheEncEcbRes))
```

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+    return ret;
+}
+
+int wh_Client_SheEncCbcRequest(whClientContext* c, uint8_t keyId, uint8_t* iv,
+    uint32_t ivSz, uint8_t* in, uint32_t sz)
+{
+    int ret;
+    uint8_t* packIn;
+    whPacket* packet;
+    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ || iv == NULL ||
```
billphipps, line 587:
```suggestion
    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ || || sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheEncCbcReq || iv == NULL ||
```

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+
+int wh_Client_SheEncCbcResponse(whClientContext* c, uint8_t* out, uint32_t sz)
+{
+    int ret;
+    uint16_t group;
+    uint16_t action;
+    uint16_t dataSz;
+    uint8_t* packOut;
+    whPacket* packet;
+    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ)
```
billphipps, line 614:
```suggestion
    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ || sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheEncCbcRes))
```

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+    return ret;
+}
+
+int wh_Client_SheDecEcbRequest(whClientContext* c, uint8_t keyId, uint8_t* in,
+    uint32_t sz)
+{
+    int ret;
+    uint8_t* packIn;
+    whPacket* packet;
+    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ)
```
billphipps, line 650:
```suggestion
    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ || sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheDecEcbReq))
```

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+
+int wh_Client_SheDecEcbResponse(whClientContext* c, uint8_t* out, uint32_t sz)
+{
+    int ret;
+    uint16_t group;
+    uint16_t action;
+    uint16_t dataSz;
+    uint8_t* packOut;
+    whPacket* packet;
+    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ)
```
billphipps, line 674:
```suggestion
    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ || sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheDecEcbRes))
```

src/wh_client_she.c
billphipps: Yeah, dumb me!

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+    return ret;
+}
+
+int wh_Client_SheDecCbcRequest(whClientContext* c, uint8_t keyId, uint8_t* iv,
+    uint32_t ivSz, uint8_t* in, uint32_t sz)
+{
+    int ret;
+    uint8_t* packIn;
+    whPacket* packet;
+    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ || iv == NULL ||
```
billphipps, line 710:
```suggestion
    if (c == NULL || in == NULL || sz < WOLFHSM_SHE_KEY_SZ ||  || sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheDecCbcReq) || iv == NULL ||
```

src/wh_client_she.c
```diff
@@ -0,0 +1,883 @@
...
+
+int wh_Client_SheDecCbcResponse(whClientContext* c, uint8_t* out, uint32_t sz)
+{
+    int ret;
+    uint16_t group;
+    uint16_t action;
+    uint16_t dataSz;
+    uint8_t* packOut;
+    whPacket* packet;
+    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ)
```
billphipps, line 737:
```suggestion
    if (c == NULL || out == NULL || sz < WOLFHSM_SHE_KEY_SZ || sz > (WH_COMM_DATA_LEN - WOLFHSM_PACKET_STUB_SIZE - sizeof(packet->sheDecCbcRes))
```

billphipps (approved):
A few suggestions to handle overrunning the packet size.  Looks good enough for now.

PR #32, opened 2024-05-24
guard around if HAVE_ECC is used, for example build

JacobBarthelmeh:


wolfhsm/wh_server.h
```diff
@@ -58,8 +58,10 @@ typedef struct {
     int devId;
     Aes aes[1];
     RsaKey rsa[1];
+#ifdef HAVE_ECC
```
dgarske, line 61: When are we building without ECC? Maybe the Aes (NO_AES), RsaKey (NO_RSA) and curve_25519 (HAVE_CURVE25519) should also be added?

dgarske (changes requested):

dgarske (approved):

PR #37, opened 2024-05-29
wolfHSM example variables fix / Macro guard checks

jackctj117:
Added HAVE_ECC macro guards in wh_client_cryptocb.c and wh_client_cryptocb.c for a couple variables in order to fix unused variables warnings in wolfHSM-examples.

### DIFF
--- a/test/wh_test_crypto.c
+++ b/test/wh_test_crypto.c
@@ -74,12 +74,8 @@ int whTest_CryptoClientConfig(whClientConfig* config)
     int res = 0;
     /* wolfcrypt */
     WC_RNG rng[1];
-#ifndef NO_AES
     Aes aes[1];
-#endif
-#ifndef NO_RSA
     RsaKey rsa[1];
-#endif
     ecc_key eccPrivate[1];
     ecc_key eccPublic[1];
     curve25519_key curve25519PrivateKey[1];
@@ -91,16 +87,12 @@ int whTest_CryptoClientConfig(whClientConfig* config)
     uint8_t keyEnd[16];
     uint8_t labelStart[WOLFHSM_NVM_LABEL_LEN];
     uint8_t labelEnd[WOLFHSM_NVM_LABEL_LEN];
-#ifndef NO_AES
     uint8_t iv[AES_IV_SIZE];
-#endif
     char plainText[16];
     char cipherText[256];
     char finalText[256];
     uint8_t authIn[16];
-#ifndef NO_AES
     uint8_t authTag[16];
-#endif
     uint8_t sharedOne[CURVE25519_KEYSIZE];
     uint8_t sharedTwo[CURVE25519_KEYSIZE];
     uint8_t knownCmacKey[] = {0x2b, 0x7e, 0x15, 0x16, 0x28, 0xae, 0xd2, 0xa6,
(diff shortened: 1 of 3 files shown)


bigbrett:
@jackctj117 reassign me when this is ready for review after the additions we talked about on slack

jpbland1:
@jackctj117 I've refactored `wh_server_crypto.c` and it addresses this issue, I would say just do the client code and test code but take a take a look at my PR https://github.com/wolfSSL/wolfHSM/pull/37 and make comments if you see any other issues in your testing, I'll incorporate them

jackctj117:
@jpbland1 Sounds good, I just finished up looking at `NO_AES` and believe I caught everything in the other files, I will push it to the PR soon. Do you want me to still pus my `wh_server_crypto.c`?

jpbland1:
@jackctj117 No I moved all that code into separate functions with compile guards and it's going to cause a merge conflict

bigbrett (changes requested):
@jackctj117 my bad, `HAVE_AES` isn't the macro you want to use (that doesn't exist in wolfCrypt). It is actually `NO_AES` that turns off AES. Same with RSA: `NO_RSA`. 

See how it is guarded in other places in the functions you touched.

src/wh_client_cryptocb.c
```diff
@@ -90,10 +94,12 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
     {
     case WC_ALGO_TYPE_CIPHER:
         /* set type */
+#ifdef HAVE_AES
         packet->cipherAnyReq.type = info->cipher.type;
         /* set enc */
         packet->cipherAnyReq.enc = info->cipher.enc;
         switch (info->cipher.type)
```
bigbrett, line 101: you should not be macroing out the switch construct like this, it make it difficult to read. Just protect the calls to the wolfCrypt APIs

src/wh_client_cryptocb.c
```diff
@@ -244,7 +250,9 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
             break;
 #endif /* HAVE_AESGCM */
 #endif /* NO_AES */
+#ifdef HAVE_AES
```
bigbrett, line 253: don't protect the control flow constructs, see above

bigbrett (changes requested):

src/wh_client_cryptocb.c
jackctj117: When using the macros guard it should be `#ifndef NO_AES`?

src/wh_client_cryptocb.c
jackctj117: What cipher options are there other than AES, because when no AES is defined cipher will not work. So should I be worrying about this switch statement not working when AES isn't defined?

bigbrett: wolfCrypt has other ciphers that it supports. That said, I'm actually realizing that we don't need these macros here, as this cryptoCallback is invoked internally by wolfCrypt functions, which will not actually be compiled if they have built with `NO_AES` and friends. So since these are just sending requests to the server, we might be A-OK leaving this file as is. Let me discuss this with some other folks and I'll get back to you

bigbrett: Yeah @jackctj117 you don't need to add protection to this file.

jackctj117: Ok sounds good I'll remove it from the PR.

test/wh_test_crypto.c
```diff
@@ -346,6 +374,8 @@ int whTest_CryptoClientConfig(whClientConfig* config)
         printf("Failed to wc_RsaPrivateDecrypt %d\n", ret);
         goto exit;
     }
+#endif
+#ifndef NO_RSA
```
bigbrett, line 378: These should be unified

test/wh_test_crypto.c
```diff
@@ -367,6 +397,7 @@ int whTest_CryptoClientConfig(whClientConfig* config)
         printf("RSA SUCCESS\n");
     else
         printf("RSA FAILED TO MATCH\n");
+#endif
```
bigbrett, line 400: Let's add closing comments for long (more than 10 lines) macro blocks. Makes it easier to keep track of, especially when nested.

```suggestion
#endif /* !NO_RSA */
```

bigbrett (changes requested):
@jackctj117 this will need a rebase, as John's latest PR that just merged will probably break some of this

jackctj117:
I have fixed merge conflicts and consolidated the macros.

PR #38, opened 2024-05-30
static check for structure unionization

bigbrett:
- Adds a compile-time check for wolfCrypt's `HAVE_ANONYMOUS_INLINE_AGGREGATES` macro to ensure the `wc_CryptoInfo` structure used in our client crypto callback is unionized. This is required as we assume it is in a few different places
- Renames client crypto callback header file to match the source file

### DIFF
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -33,7 +33,7 @@
 #include "wolfhsm/wh_comm.h"
 
 #ifndef WOLFHSM_NO_CRYPTO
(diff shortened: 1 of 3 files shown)


jpbland1 (approved):
Change looks good, if the broken `#include "wolfhsm/wh_cryptocb.h"` didn't cause any issues it's probably because we're not actually using anything from cryptocb in the client function code

PR #39, opened 2024-06-05
add nvm counters

jpbland1:


src/wh_server_counter.c
```diff
@@ -0,0 +1,104 @@
...
+    switch (action)
+    {
+    case WH_COUNTER_RESET:
+        /* write 0 to nvm with the supplied id and user_id */
+        counter = 0;
+        memset((void*)meta, 0, sizeof(meta));
+        meta->id = MAKE_WOLFHSM_KEYID(WOLFHSM_KEYTYPE_COUNTER,
+            server->comm->client_id, packet->counterResetReq.keyId);
+        meta->len = sizeof(counter);
+        ret = wh_Nvm_AddObject(server->nvm, meta, meta->len,
```
billphipps, line 49: I thought we were going to store the counter value in the label so we didn't need any data segment?

src/wh_server_counter.c
```diff
@@ -0,0 +1,104 @@
...
+        if (ret == 0) {
+            ret = wh_Nvm_Read(server->nvm, meta->id, 0, sizeof(counter),
+                (uint8_t*)&counter);
+        }
+        /* increment and write the counter back */
+        if (ret == 0) {
+            counter++;
+            /* check if counter rolled over */
+            if (counter == 0)
+                ret = WH_ERROR_NOSPACE;
```
billphipps, line 71: This error is incorrect and conflicts with the real possibility of not having any free entries in the directory (which the user should immediately attempt to reclaim space using DestroyObjects).  The counter should be defined as saturating so if it is going to roll over, it simply stays at the highest value and does not return an error.  Since you return the value, the user can check

src/wh_server_counter.c
```diff
@@ -0,0 +1,104 @@
...
+        }
+        break;
+    case WH_COUNTER_INCREMENT:
+        /* read the counter's metadata */
+        ret = wh_Nvm_GetMetadata(server->nvm, MAKE_WOLFHSM_KEYID(
+            WOLFHSM_KEYTYPE_COUNTER, server->comm->client_id,
+            packet->counterIncrementReq.keyId), meta);
+        /* read the counter value */
+        if (ret == 0) {
+            ret = wh_Nvm_Read(server->nvm, meta->id, 0, sizeof(counter),
```
billphipps, line 63: If the counter is left in the metadata label, then you won't have to Nvm_Read here.  Just get the metadata and decode the label

src/wh_server_counter.c
```diff
@@ -0,0 +1,104 @@
...
+    whPacket* packet = (whPacket*)data;
+    whNvmMetadata meta[1] = {0};
+
+    if (server == NULL || server->nvm == NULL || data == NULL || size == NULL)
+        return WH_ERROR_BADARGS;
+
+    switch (action)
+    {
+    case WH_COUNTER_RESET:
+        /* write 0 to nvm with the supplied id and user_id */
```
billphipps, line 43: There's no way to remove a counter from NVM.  Recommend to add a function to destroy a counter.

src/wh_server_counter.c
```diff
@@ -0,0 +1,104 @@
...
+        counter = 0;
```
billphipps, line 44: Consider letting the user specify the starting value.

test/wh_test_crypto.c
```diff
@@ -562,6 +561,32 @@ int whTest_CryptoClientConfig(whClientConfig* config)
...
         goto exit;
     }
+    printf("CMAC SUCCESS\n");
+    /* test counters */
+    keyId = 1;
+    if ((ret = wh_Client_CounterReset(client, keyId, &counter)) != 0 || counter != 0) {
+        WH_ERROR_PRINT("Failed to wh_Client_CounterReset %d\n", ret);
+        goto exit;
+    }
+    if ((ret = wh_Client_CounterIncrement(client, keyId, &counter)) != 0 || counter != 1) {
```
billphipps, line 571: Make sure to add a loop that invokes Increment more times than entries in the directory (currently 32), so this call will return with WH_ERROR_NOSPACE.  You should add a wh_Client_NvmDestroyObjects(client, 0,NULL,&out_rc) to reclaim the space and retry the CounterIncrement.  Just noticed the DestroyObjects API is incorrect.

billphipps (changes requested):
Looks good.  a few changes:
1. Please move the counter value into the metadata label. 
2. Consider automatically handling the reclaimable space error when incrementing.

wolfhsm/wh_packet.h
```diff
@@ -307,6 +307,36 @@ typedef struct WOLFHSM_PACK wh_Packet_version_exchange
...
+typedef struct WOLFHSM_PACK wh_Packet_counter_read_req
+{
+    uint16_t keyId;
+} wh_Packet_counter_read_req;
+
+typedef struct WOLFHSM_PACK wh_Packet_counter_read_res
+{
+    uint32_t counter;
+} wh_Packet_counter_read_res;
+
```
bigbrett, line 339: hmm, I'm noticing we have a split paradigm developing here - you have been putting messages into the `whPacket` union in this file, like you do for crypto, whereas Bill and I have been adding new message definitions and the necessary marshalling/data transformations in their own `wh_message_xxx.{c,h}` files. I think we should standardize on doing it one way. 

Related, I see there aren't any endian transformations for these messages. This will break if we ever have platforms with mixed HSM and host endianness. The reason Bill and I implemented the intermediate message layer is to allow a distinct module that handles the endian transformations for a given message data type, between host format and HSM format. 

I'm also now realizing we don't do that for any of the crypto...I think that is something we will need to add at some point.

I suppose we could punt it for now, though I'd really prefer adding the appropriate transformation support to this module before merging, and just punting the crypto. And separating this from crypto in its entirety, as it is a completely different feature/module

bigbrett (changes requested):
Nice. A few comments and things to address. Also, I would really really like to see tests do a little more exercising of the functionality, testing edge cases, etc. rather than just incrementing a counter once and resetting it. Doing some incrementation in a loop, demonstrating proper handling of rollover/saturation, etc would be really good. Tests that could catch regressions to the logic are awesome.

wolfhsm/wh_packet.h
jpbland1: I think everything should be put in packet.h and all packets should be defined in the single unionionized struct that I've been using so we don't have to serialize and deserialize anything and can use a single piece of memory for the packet by just casting. We will need to add networkize and hostize functions that swap the endianness in a single switch statement based on message group and action. frankly adding new files for all this just makes more technical debt

jpbland1:
Updated based on you guys feedback, I had to modify nvm to allow NULL data entries for our metadata only idea, need to confirm this doesn't have any unintended consequences

src/wh_client.c
```diff
@@ -733,6 +733,203 @@ int wh_Client_KeyErase(whClientContext* c, whNvmId keyId)
...
+    if (ret == 0) {
+        if (packet->rc != 0)
+            ret = packet->rc;
+        else
+            *counter = packet->counterInitRes.counter;
+    }
+    return ret;
+}
+
+int wh_Client_CounterInit(whClientContext* c, whNvmId keyId, uint32_t* counter)
```
billphipps, line 771: Consider splitting the input counter and the output counter parameters and allow null for output counter.

src/wh_client.c
```diff
@@ -733,6 +733,203 @@ int wh_Client_KeyErase(whClientContext* c, whNvmId keyId)
...
+}
+
+int wh_Client_CounterIncrementResponse(whClientContext* c, uint32_t* counter)
+{
+    uint16_t group;
+    uint16_t action;
+    uint16_t size;
+    int ret;
+    whPacket* packet;
+    if (c == NULL || counter == NULL)
```
billphipps, line 819: Recommend to allow NULL output counter

src/wh_client.c
```diff
@@ -733,6 +733,203 @@ int wh_Client_KeyErase(whClientContext* c, whNvmId keyId)
...
+        return WH_ERROR_BADARGS;
+    packet = (whPacket*)wh_CommClient_GetDataPtr(c->comm);
+    /* set keyId */
+    packet->counterDestroyReq.keyId = keyId;
+    return wh_Client_SendRequest(c, WH_MESSAGE_GROUP_COUNTER,
+        WH_COUNTER_DESTROY, WOLFHSM_PACKET_STUB_SIZE +
+        sizeof(packet->counterReadReq), (uint8_t*)packet);
+}
+
+int wh_Client_CounterDestroyResponse(whClientContext* c)
```
billphipps, line 903: If there could be ambiguity in the return value between local and remote, then you should provide the local return value as the return to this function and use and output parameter for the remote return value.

src/wh_server_counter.c
```diff
@@ -0,0 +1,112 @@
...
+    switch (action)
+    {
+    case WH_COUNTER_INIT:
+        /* write 0 to nvm with the supplied id and user_id */
+        memset((void*)meta, 0, sizeof(meta));
+        meta->id = MAKE_WOLFHSM_KEYID(WOLFHSM_KEYTYPE_COUNTER,
+            server->comm->client_id, packet->counterInitReq.keyId);
+        /* use the label buffer to hold the counter value */
+        *(uint32_t*)meta->label = packet->counterInitReq.counter;
+        ret = wh_Nvm_AddObject(server->nvm, meta, 0, NULL);
```
billphipps, line 49: Shouldn't Init return an error if the object already exists?  Otherwise, this simply overwrites the current value.

src/wh_server_counter.c
```diff
@@ -0,0 +1,112 @@
...
+        /* if we ran out of space try cleaning up duplicate entries and then retry */
```
billphipps, line 50:
```suggestion
        /* if we ran out of space try reclaiming space and then retry */
```

src/wh_server_counter.c
```diff
@@ -0,0 +1,112 @@
...
+        ret = wh_Nvm_GetMetadata(server->nvm, MAKE_WOLFHSM_KEYID(
+            WOLFHSM_KEYTYPE_COUNTER, server->comm->client_id,
+            packet->counterIncrementReq.keyId), meta);
+        /* increment and write the counter back */
+        if (ret == 0) {
+            (*(uint32_t*)meta->label)++;
+            /* set counter to uint32_t max if it rolled over */
+            if (*(uint32_t*)meta->label == 0)
+                *(uint32_t*)meta->label = 0xffffffff;
+            ret = wh_Nvm_AddObject(server->nvm, meta, 0, NULL);
```
billphipps, line 72: No need to AddObject if the count is saturated.

src/wh_server_counter.c
```diff
@@ -0,0 +1,112 @@
...
+        /* write 0 to nvm with the supplied id and user_id */
+        memset((void*)meta, 0, sizeof(meta));
+        meta->id = MAKE_WOLFHSM_KEYID(WOLFHSM_KEYTYPE_COUNTER,
+            server->comm->client_id, packet->counterInitReq.keyId);
+        /* use the label buffer to hold the counter value */
+        *(uint32_t*)meta->label = packet->counterInitReq.counter;
+        ret = wh_Nvm_AddObject(server->nvm, meta, 0, NULL);
+        /* if we ran out of space try cleaning up duplicate entries and then retry */
+        if (ret == WH_ERROR_NOSPACE) {
+            ret = wh_Nvm_DestroyObjects(server->nvm, 0, NULL);
```
billphipps, line 52: Should check to see if there is any space that could be reclaimed GetAvailable() before invoking the painful DestroyObjects().

src/wh_server_counter.c
```diff
@@ -0,0 +1,112 @@
...
+            WOLFHSM_KEYTYPE_COUNTER, server->comm->client_id,
+            packet->counterReadReq.keyId), meta);
+        /* return counter to the caller */
+        if (ret == 0) {
+            packet->counterReadRes.counter = *(uint32_t*)meta->label;
+            *size = WOLFHSM_PACKET_STUB_SIZE + sizeof(packet->counterReadRes);
+        }
+        break;
+    case WH_COUNTER_DESTROY:
+        keyId = packet->counterDestroyReq.keyId;
```
billphipps, line 99: Should this be MAKE_WOLFHSM_KEYID(WOLFHSM_KEYTYPE_COUNTER, ... ) ?

wolfhsm/wh_client.h
```diff
@@ -617,6 +617,82 @@ int wh_Client_AesCmacVerify(Cmac* cmac, const byte* check, word32 checkSz,
...
+    uint32_t counter);
+int wh_Client_CounterInitResponse(whClientContext* c, uint32_t* counter);
+/**
+ * @brief Creates and initializes a counter with the value set in counter.
+ *
+ * This function creates/resets a counter with the supplied keyId and gives it
+ * the value stored in counter at the start of the call.
+ *
+ * @param[in] c Pointer to the whClientContext structure.
+ * @param[in] keyId Key ID to be associated with the CMAC key.
```
billphipps, line 631: Copy pasta with the CMAC.  Consider naming the parameter CounterId instead of KeyId.

billphipps (changes requested):
Maybe a bug in the destroy function?  Also, the overall intent of these counters was to provide a basis to implement the PKCS11 monotonic counters (https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/csprd01/pkcs11-base-v3.0-csprd01.html#_Toc10472588), so let's make sure that we can simulate the same behavior using the exposed API and internals.

wolfhsm/wh_packet.h
billphipps: Counters should be supported even when crypto is not.  We'll need to move things around in the future, but we'll stay the course for now.  Providing separate translation units means fewer conditional compilation flags, so the giant union isn't always the easiest way to manage bulk data.  Each individual kind of message will be typecast anyway, and should be processed in separate functions, so the union only exists at the main dispatch.

Again, nothing to fix at this point, but when we get a chance, we really should focus on eliminating "packed" by reorganizing the structs and using fixed size types in message fields.  Then, the only processing per field is for endianness.

bigbrett: Well I think we will always need to serialize/deserialize. To clarify, when I say serialize/deserialize I'm referring to the same thing as you are when you say networkize/hostize. I suppose "marshalling" is a better word to use. Going from a standard "wire format" to a native host format. So if we want to support platforms with different byte orders, there isn't really any way we can get around this operation. And it goes beyond endianness too, it includes packing/padding of structs, which can be annoyingly different for different compilers/toolchains/architectures.

And the marshaling is already is done based on the message group and action like you are suggesting, but it just is handled one level down (e.g. NVM messages are deserialized in `wh_Server_HandleNvmRequest`, Comm messages are deserialized in `wh_Server_HandleCommRequest`, etc.) instead of before the message dispatch in `wh_Server_HandleRequest()`. Not sure why doing it this way results in more technical debt. Some might say more files mean less code in a given context, making things clearer, and there is more of a clear separation of concerns. Not at all trying to die on this hill, I just don't think it is necessarily a problem. What is a problem IMO, is not doing any proper marshalling of data and manually reordering/padding the structs to try and avoid issues.

That said, I've long been thinking about using a ser/de library (or making our own simple one) so wolfHSM itself doesn't need to worry about this stuff. Something like protobuf/flatbuffers or the like. Since using packed structs on different architectures is pretty dangerous, and right now we are just manually packing/aligning structs which seems like not the best way.

Ideally we have a struct representation of each data type that is a normal native struct able to be used on both sides, and we can just call `myStruct_deSerialize(&myLocalStruct, &commBuffer)` or `myStruct_serialize(&myLocalStruct, &commBuffer)`  and the data would be properly marshalled without worrying about endianness, padding, alignment, etc.

Add it to the backlog, I suppose...

jpbland1:
rebased, need to discuss the issue with `wh_Nvm_DestroyObjects` the way I'm using it

src/wh_nvm.c
```diff
@@ -87,6 +87,36 @@ int wh_Nvm_GetAvailable(whNvmContext* context,
             out_reclaim_size, out_reclaim_objects);
 }
 
+int wh_Nvm_AddObjectPessimistic(whNvmContext* context, whNvmMetadata *meta,
```
billphipps, line 90: Let's try a more objective name.  How about:
```suggestion
int wh_Nvm_AddObjectWithReclaim(whNvmContext* context, whNvmMetadata *meta,
```

src/wh_nvm.c
```diff
@@ -87,6 +87,36 @@ int wh_Nvm_GetAvailable(whNvmContext* context,
...
+    const uint8_t* data, whNvmSize dataLen)
```
billphipps, line 91: Keep the same argument order
```suggestion
    whNvmSize dataLen, const uint8_t* data)
```

src/wh_nvm.c
```diff
@@ -87,6 +87,36 @@ int wh_Nvm_GetAvailable(whNvmContext* context,
...
+    int ret;
+    uint32_t availableSize;
+    uint32_t reclaimSize;
+    if (context == NULL)
+        return WH_ERROR_BADARGS;
+    /* check if we have room for the object */
+    ret = wh_Nvm_GetAvailable(context, &availableSize, &availableObjects,
+        &reclaimSize, &reclaimObjects);
+    if (ret == 0) {
+        if (availableSize >= WOLFHSM_NVM_METADATA_LEN + dataLen &&
```
billphipps, line 104: Metadata len is not included in the available data space
```suggestion
        if (availableSize >=  dataLen &&
```

src/wh_nvm.c
```diff
@@ -87,6 +87,36 @@ int wh_Nvm_GetAvailable(whNvmContext* context,
...
+            availableObjects > 0) {
+            return wh_Nvm_AddObject(context, meta, dataLen, data);
```
billphipps, line 106: Prefer single return
```suggestion
            ret = wh_Nvm_AddObject(context, meta, dataLen, data);
```

src/wh_nvm.c
```diff
@@ -87,6 +87,36 @@ int wh_Nvm_GetAvailable(whNvmContext* context,
...
+        }
+        /* if there's no room, try to reclaim space, otherwise return error */
+        else if (availableSize + reclaimSize >= WOLFHSM_NVM_METADATA_LEN +
+            dataLen && reclaimObjects > 0) {
+            ret = wh_Nvm_DestroyObjects(context, 0, NULL);
+            if (ret == 0)
+                return wh_Nvm_AddObject(context, meta, dataLen, data);
```
billphipps, line 113:
```suggestion
                ret = wh_Nvm_AddObject(context, meta, dataLen, data);
```

src/wh_nvm.c
```diff
@@ -87,6 +87,36 @@ int wh_Nvm_GetAvailable(whNvmContext* context,
...
+    ret = wh_Nvm_GetAvailable(context, &availableSize, &availableObjects,
+        &reclaimSize, &reclaimObjects);
+    if (ret == 0) {
+        if (availableSize >= WOLFHSM_NVM_METADATA_LEN + dataLen &&
+            availableObjects > 0) {
+            return wh_Nvm_AddObject(context, meta, dataLen, data);
+        }
+        /* if there's no room, try to reclaim space, otherwise return error */
+        else if (availableSize + reclaimSize >= WOLFHSM_NVM_METADATA_LEN +
+            dataLen && reclaimObjects > 0) {
```
billphipps, line 110:
```suggestion
        else if (availableSize + reclaimSize >= dataLen && reclaimObjects > 0) {
```

src/wh_server_counter.c
```diff
@@ -0,0 +1,104 @@
...
+    whNvmMetadata meta[1] = {0};
+
+    if (server == NULL || server->nvm == NULL || data == NULL || size == NULL)
+        return WH_ERROR_BADARGS;
+
+    switch (action)
+    {
+    case WH_COUNTER_INIT:
+        /* write 0 to nvm with the supplied id and user_id */
+        memset((void*)meta, 0, sizeof(meta));
```
billphipps, line 44: Already zeroed above.
```suggestion
```

src/wh_server_counter.c
```diff
@@ -0,0 +1,104 @@
...
+#include "wolfhsm/wh_server_counter.h"
+#include "wolfhsm/wh_message.h"
+#include "wolfhsm/wh_packet.h"
+#include "wolfhsm/wh_error.h"
+
+int wh_Server_HandleCounter(whServerContext* server, uint16_t action,
+    uint8_t* data, uint16_t* size)
+{
+    whKeyId counterId;
+    int ret;
```
billphipps, line 33:
```suggestion
    whKeyId counterId = 0;
    int ret = 0;
```

src/wh_server_keystore.c
```diff
@@ -304,8 +304,16 @@ int hsmCommitKey(whServerContext* server, whNvmId keyId)
...
+    ret = wh_Nvm_AddObjectPessimistic(server->nvm, cacheSlot->meta,
+        cacheSlot->buffer, cacheSlot->meta->len);
+    /* if we ran out of space clean out duplicate entries and retry */
+    if (ret == WH_ERROR_NOSPACE) {
+        ret = wh_Nvm_DestroyObjects(server->nvm, 0, NULL);
+        if (ret == 0) {
+            ret = wh_Nvm_AddObject(server->nvm, cacheSlot->meta,
+                cacheSlot->meta->len, cacheSlot->buffer);
+        }
+    }
```
billphipps, line 316: Duplicate functionality.  Remove.
```suggestion
```

billphipps (changes requested):
A couple of tweaks necessary but looks good.

src/wh_nvm.c
jpbland1: I'll change it for now but I think all the functions should use buffer, length (like all the woflCrypt, wolfssl, wolfBoot etc. functions)

billphipps (approved):

bigbrett (approved):

PR #40, opened 2024-06-05
cleanups/fixes from tricore integration

bigbrett:
- lingering typecast fixes for wolfCrypt types 
- adds SHE macro protection 
- adds comm init call to unit tests 
- remove hard-coded client ID in SHE server test
- Adds custom `htonl`/`ntohl`/`memeq` to new utils module
- Fix "leaked" NVM objects in crypto and SHE tests

### DIFF
--- a/src/wh_utils.c
+++ b/src/wh_utils.c
@@ -1,0 +1,32 @@
+#include "wolfhsm/wh_utils.h"
+
+static int isLittleEndian() {
+    unsigned int x = 1; /* 0x00000001 */
+    char *c = (char*)&x;
+    return (int)*c;
+}
+
+/* Converts a 32-bit value from host to network byte order */
+uint32_t wh_Utils_htonl(uint32_t hostlong) {
+    if (isLittleEndian()) {
+        return ((hostlong & 0x000000FF) << 24) | ((hostlong & 0x0000FF00) << 8)
+            | ((hostlong & 0x00FF0000) >> 8) | ((hostlong & 0xFF000000) >> 24);
+    }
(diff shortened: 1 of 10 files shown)


jpbland1 (dismissed):
Looks solid

src/wh_utils.c
```diff
@@ -0,0 +1,32 @@
+#include "wolfhsm/wh_utils.h"
+
+static int isLittleEndian() {
+    unsigned int x = 1; /* 0x00000001 */
+    char *c = (char*)&x;
+    return (int)*c;
+}
```
billphipps, line 7: This should be a compile-time constant.  Let's add a backlog item to replace this with a macro based on some of the compiler defines and fallback to dynamic evaluation only if we can't determine it at compile time.

billphipps (dismissed):
All good! Couple of needless nits, but ok to merge

test/wh_test_she.c
```diff
@@ -59,6 +60,23 @@ enum {
         BUFFER_SIZE = 4096,
     };
 
+/* Helper function to destroy a SHE key so the unit tests don't
+ * leak NVM objects across invocations. Necessary, as SHE doesn't expose a
+ * destroy key API since SHE keys are supposed to be fixed hardware keys */
+static int _destroySheKey(whClientContext* client, whNvmId clientSheKeyId)
```
billphipps, line 66: Hrmmm.  This could be added to our SHE library, even if it is outside of the SHE spec.

billphipps (approved):

PR #41, opened 2024-06-05
use the comm packet buffer instead of a stack buffer,

jpbland1:
validate inputs to make sure they won't overrun the comm buffer

### DIFF
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -480,7 +480,7 @@ int wh_Client_KeyCacheRequest_ex(whClientContext* c, uint32_t flags,
         sizeof(packet->keyCacheReq) + inSz > WH_COMM_DATA_LEN) {
         return WH_ERROR_BADARGS;
     }
-    packet = (whPacket*)c->comm->data;
+    packet = (whPacket*)wh_CommClient_GetDataPtr(c->comm);
     packIn = (uint8_t*)(&packet->keyCacheReq + 1);
     packet->keyCacheReq.id = keyId;
     packet->keyCacheReq.flags = flags;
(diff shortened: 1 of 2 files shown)


src/wh_client.c
```diff
@@ -513,9 +516,10 @@ int wh_Client_KeyCacheResponse(whClientContext* c, uint16_t* keyId)
     uint16_t action;
     uint16_t size;
     int ret;
-    whPacket packet[1] = {0};
+    whPacket* packet;
     if (c == NULL || keyId == NULL)
         return WH_ERROR_BADARGS;
+    packet = (whPacket*)c->comm->data;
```
billphipps, line 522: Use accessor function and check return value.
```suggestion
    packet = (whPacket*)wh_CommClient_GetDataPtr(c->comm);
```

src/wh_client_cryptocb.c
```diff
@@ -104,6 +104,8 @@ int wolfHSM_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
             dataSz = sizeof(packet->cipherAesCbcReq) +
                 info->cipher.aescbc.aes->keylen + AES_IV_SIZE +
                 info->cipher.aescbc.sz;
+            if (dataSz > WH_COMM_DATA_LEN)
+                return CRYPTOCB_UNAVAILABLE;
```
billphipps, line 108: Shouldn't this be a BAD_ARGS return?  It's likely the local SW crypto won't be able to handle the call.

billphipps (changes requested):
Looks good. 
1. Use commclient accessor function instead of using comm->data
2. Consider BAD_ARGS return instead of UNAVAILABLE on too large requests.

jpbland1:
@billphipps yeah you're right, we should return `BAD_FUNC_ARG` when we can't use fallback and `CRYPTOCB_UNAVAILABLE` when we can, the pk algorithms can't do fallback because the key is always on the HSM but aes and cmac can have client side keys, just need to check if  the devCtx has been set

billphipps (approved):
Looks great!  Nice check for using an HSM key or not.

PR #42, opened 2024-06-05
dummy names for anonymous structures/unions

bigbrett:
doxybook (the program we use to render doxygen to a nice-looking website) can't handle unnamed enums or unions. David provided this patch that will allow it to render. Normally I'd say we shouldn't change our code just to accommodate an external (buggy) tool, but it would require us to create separate headers for doc purposes which are easy to get out of sync, and in this case the changes are pretty innocuous. I like our code having the doxygen in it, so I say it is worth it in this small case.

### DIFF
--- a/wolfhsm/wh_server.h
+++ b/wolfhsm/wh_server.h
@@ -54,12 +54,12 @@ typedef struct CacheSlot {
     uint8_t       buffer[WOLFHSM_KEYCACHE_BUFSIZE];
 } CacheSlot;
 
-typedef struct {
+typedef struct crypto_context {
     int devId;
 #ifndef WC_NO_RNG
     WC_RNG rng[1];
 #endif
-    union {
+    union crypto_context_cryptoctx {
 #ifndef NO_AES
         Aes aes[1];
 #endif
@@ -76,7 +76,7 @@ typedef struct {
         Cmac cmac[1];
 #endif
     };
-    union {
+    union crypto_context_pubkey {
 #ifdef HAVE_ECC
         ecc_key eccPublic[1];
 #endif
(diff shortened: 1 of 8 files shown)


wolfhsm/wh_server.h
```diff
@@ -76,7 +76,7 @@ typedef struct {
         Cmac cmac[1];
 #endif
     };
-    union {
+    union crypto_context_pubkey {
```
bigbrett, line 79: @jpbland1 can you think of better names for these? Doesn't matter that much since the name of the union will never be referenced, but it is a good opportunity for more self-documenting code

wolfhsm/wh_server.h
```diff
@@ -75,15 +75,15 @@ typedef struct {
 #ifdef WOLFSSL_CMAC
         Cmac cmac[1];
 #endif
-    };
+    } crypto_context_crypto;
     union {
```
dgarske, line 79: I think you might need to move this up to `union crypto_context_pubkey {`. Otherwise you'll have to change the calling code.

dgarske (changes requested):

wolfhsm/wh_server.h
dgarske: For reference... Doxygen will find the "name" is either place, but if you change it after then you'll have to change the calling code to use `. crypto_context_crypto`.

bigbrett: yep, you are correct just saw that. Unfortunately the way you originally had it generates a compiler warning, which I was trying to get rid of. But realize now that the way I did it is invalid. Hmmm....

```
wh_server.h:86:6: warning: declaration does not declare anything
   86 |     };
```

where the `};` reference is the closing bracket of the nested union.

dgarske (approved):

PR #45, opened 2024-06-12
Random lingering tricore cleanups

bigbrett:
- Decreased `WOLFHSM_NUM_RAMKEYS` from 16 to 8 for smaller defaults
- Moved counter tests to `wh_test_clientserver.c`
- No longer protecting client counter and keymgr APIs with `WOLFHSM_NO_CRYPTO` (that macro should only mean we aren't building with wolfcrypt suppport)
- Removes use of `XMEMCPY` from functions that are used when `WOLFHSM_NO_CRYPTO` is defined. This is a temporary workaround, but right now since `XMEMCPY` is defined in `user_settings.h`, which is not included when `WOLFHSM_NO_CRYPTO` is defined, then we get compiler errors. If we want `XMEMCPY` and `Xwhatever()` to be a thing in wolfHSM we are going to need to do a little bit of work around settings/configuration to make that possible when decoupled from wolfCrypt.
- Added missing includes and a missing printf

### DIFF
--- a/test/wh_test_check_struct_padding.c
+++ b/test/wh_test_check_struct_padding.c
@@ -1,0 +1,120 @@
+#ifndef WH_TEST_CHECK_STRUCT_PADDING_H_
+#define WH_TEST_CHECK_STRUCT_PADDING_H_
+
+/* For each included file, define an instance of every struct for which we want
+ * to check padding. Then, when this file is compiled with -Wpadded it will
+ * generate an error if padding is wrong */
+
+
+#include "wolfhsm/wh_message_comm.h"
+whMessageComm_ErrorResponse whMessageComm_ErrorResponse_test;
+whMessageCommLenData        whMessageCommLenData_test;
+whMessageCommInitRequest    whMessageCommInitRequest_test;
+whMessageCommInitResponse   whMessageCommInitResponse_test;
+whMessageCommInfo           whMessageCommInfo_test;
+
+#include "wolfhsm/wh_message_customcb.h"
+whMessageCustomCb_Request  whMessageCustomCb_Request_test;
+whMessageCustomCb_Response whMessageCustomCb_Response_test;
+
+#include "wolfhsm/wh_message_nvm.h"
+whMessageNvm_SimpleResponse        whMessageNvm_SimpleResponse_test;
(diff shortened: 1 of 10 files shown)


wolfhsm/wh_common.h
```diff
@@ -35,7 +35,7 @@
 /** Resource allocations */
 enum WOLFHSM_NUM_ENUM {
     WOLFHSM_NUM_COUNTERS = 8,       /* Number of non-volatile 32-bit counters */
-    WOLFHSM_NUM_RAMKEYS = 16,        /* Number of RAM keys */
+    WOLFHSM_NUM_RAMKEYS = 8,        /* Number of RAM keys */
```
billphipps, line 38: Agree that these should be overridable at compile time by the user and likely placed in a seperate wh_config.h file.  Need to remove these from the enum and make then #defines at some point then.  Ok for now

test/wh_test_clientserver.c
```diff
@@ -458,6 +458,94 @@ static int _testDma(whServerContext* server, whClientContext* client)
     return rc;
 }
 
+int _testClientCounter(whClientContext* client)
+{
+    const whNvmId  counterId              = 1;
+    const uint32_t MAX_COUNTER_VAL        = 0xFFFFFFFF;
```
billphipps, line 464: Grumble.  This should be in the counter header file somewhere.  Probably should be overridable at creation.  I'll add this to the counter backlog item.  No change.

test/wh_test_check_struct_padding.c
```diff
@@ -0,0 +1,120 @@
+#ifndef WH_TEST_CHECK_STRUCT_PADDING_H_
+#define WH_TEST_CHECK_STRUCT_PADDING_H_
+
+/* For each included file, define an instance of every struct for which we want
+ * to check padding. Then, when this file is compiled with -Wpadded it will
+ * generate an error if padding is wrong */
```
billphipps, line 6: Maybe use pragma here?

billphipps (approved):
Looks great!  Couple of comments.  no changes necessary.

jpbland1 (approved):
Reviewed, looks good

PR #48, opened 2024-06-18
refactor the HSM server and keystore to handle a

jpbland1:
seperate cache for larger keys so customers with limited memory can still store a limited number of large keys alongside their smaller keys.

### DIFF
--- a/src/wh_server_crypto.c
+++ b/src/wh_server_crypto.c
@@ -55,9 +55,9 @@ static int hsmCacheKeyRsa(whServerContext* server, RsaKey* key, whKeyId* outId)
     /* wc_RsaKeyToDer doesn't have a length check option so we need to just pass
      * the big key size if compiled */
 #ifdef WOLFHSM_SEPARATE_BIG_CACHE
-    uint16_t keySz = WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE;
+    const uint16_t keySz = WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE;
 #else
-    uint16_t keySz = WOLFHSM_CFG_SERVER_KEYCACHE_BUFSIZE;
+    const uint16_t keySz = WOLFHSM_CFG_SERVER_KEYCACHE_BUFSIZE;
 #endif
     whKeyId keyId = WH_KEYTYPE_CRYPTO;
     /* get a free slot */
@@ -214,8 +214,9 @@ static int hsmCacheKeyCurve25519(whServerContext* server, curve25519_key* key,
     word32 privSz = CURVE25519_KEYSIZE;
     word32 pubSz = CURVE25519_KEYSIZE;
     whKeyId keyId = WH_KEYTYPE_CRYPTO;
+    const uint16_t keySz = CURVE25519_KEYSIZE * 2;
     /* get a free slot */
-    ret = hsmCacheFindSlotAndZero(server, CURVE25519_KEYSIZE * 2, &cacheBuf,
+    ret = hsmCacheFindSlotAndZero(server, keySz, &cacheBuf,
         &cacheMeta);
     if (ret == 0)
         ret = hsmGetUniqueId(server, &keyId);
@@ -227,7 +228,7 @@ static int hsmCacheKeyCurve25519(whServerContext* server, curve25519_key* key,
     if (ret == 0) {
         /* set meta */
         cacheMeta->id = keyId;
-        cacheMeta->len = CURVE25519_KEYSIZE * 2;
+        cacheMeta->len = keySz;
         /* export keyId */
         *outId = keyId;
     }
--- a/src/wh_server_crypto.c
+++ b/src/wh_server_crypto.c
@@ -49,25 +49,30 @@
 #ifndef NO_RSA
 static int hsmCacheKeyRsa(whServerContext* server, RsaKey* key, whKeyId* outId)
 {
+    uint8_t* cacheBuf;
+    whNvmMetadata* cacheMeta;
     int ret = 0;
-    int slotIdx = 0;
+    /* wc_RsaKeyToDer doesn't have a length check option so we need to just pass
+     * the big key size if compiled */
+#ifdef WOLFHSM_SEPARATE_BIG_CACHE
+    uint16_t keySz = WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE;
+#else
+    uint16_t keySz = WOLFHSM_CFG_SERVER_KEYCACHE_BUFSIZE;
+#endif
     whKeyId keyId = WH_KEYTYPE_CRYPTO;
     /* get a free slot */
-    ret = slotIdx = hsmCacheFindSlot(server);
-    if (ret >= 0) {
+    ret = hsmCacheFindSlotAndZero(server, keySz, &cacheBuf, &cacheMeta);
+    if (ret == 0)
         ret = hsmGetUniqueId(server, &keyId);
-    }
     if (ret == 0) {
         /* export key */
         /* TODO: Fix wolfCrypto to allow KeyToDer when KEY_GEN is NOT set */
-        XMEMSET((uint8_t*)&server->cache[slotIdx], 0, sizeof(whServerCacheSlot));
-        ret = wc_RsaKeyToDer(key, server->cache[slotIdx].buffer,
-            WOLFHSM_CFG_SERVER_KEYCACHE_BUFSIZE);
+        ret = wc_RsaKeyToDer(key, cacheBuf, keySz);
     }
     if (ret > 0) {
         /* set meta */
-        server->cache[slotIdx].meta->id = keyId;
-        server->cache[slotIdx].meta->len = ret;
+        cacheMeta->id = keyId;
+        cacheMeta->len = ret;
         /* export keyId */
         *outId = keyId;
         ret = 0;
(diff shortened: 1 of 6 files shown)


jpbland1:
Addressing the issue @bigbrett brought up where we need to support both large RSA keys while still allowing a large number of smaller keys, by making a separate cache for large keys like RSA and the cmac struct customers with limited memory can still use a large number of smaller keys. Considered changing the cache to a heap but we want the cache to be completely deterministic as to the number of keys you can store, rather than having the customer calculate the cachesize they need they should just be able to pick the number of big and small keys they plan to use.

src/wh_server_crypto.c
```diff
@@ -296,30 +302,47 @@ static int hsmCryptoCurve25519(whServerContext* server, whPacket* packet,
...
-    uint32_t qyLen;
-    uint32_t qdLen;
+    uint32_t qxLen = 0;
+    uint32_t qyLen = 0;
+    uint32_t qdLen = 0;
     whKeyId keyId = WOLFHSM_KEYTYPE_CRYPTO;
+    /* get the lengths since they vary by ecc size, expect BUFFER_E since
+     * LENGTH_ONLY_E isn't supported. Note that this is a dangerous operation 
+     * but by having size of 0 or buffer of NULL we avoid NULL checks when
+     * getting the length */
```
bigbrett, line 315: looks like those args are NULL checked in that wolfCrypt function, so shouldn't necessarily be dangerous - can you elaborate on what you mean here?

src/wh_server_crypto.c
```diff
@@ -329,18 +352,18 @@ static int hsmCacheKeyEcc(whServerContext* server, ecc_key* key, whKeyId* outId)
...
     /* decode the key */
-    if (ret >= 0) {
-        keySz = server->cache[slotIdx].meta->len / 3;
-        ret = wc_ecc_import_unsigned(key, server->cache[slotIdx].buffer,
-            server->cache[slotIdx].buffer + keySz,
-            server->cache[slotIdx].buffer + keySz * 2, curveId);
+    if (ret == 0) {
+        keySz = cacheMeta->len / 3;
+        ret = wc_ecc_import_unsigned(key, cacheBuf, cacheBuf + keySz,
+            cacheBuf + keySz * 2, curveId);
```
bigbrett, line 366: while we're here, can we `const`ify the magic numbers with self describing names  just to make it explicitly clear what is going on at a glance

src/wh_server_crypto.c
```diff
@@ -709,6 +747,12 @@ static int hsmCryptoCmac(whServerContext* server, whPacket* packet,
                 keyId = MAKE_WOLFHSM_KEYID(WOLFHSM_KEYTYPE_CRYPTO,
                     server->comm->client_id, packet->cmacReq.keyId);
             }
+#ifdef WOLFHSM_SEPARATE_BIG_CACHE
+            /* evict the aes sized key in the normal cache */
+            if (moveToBigCache == 1) {
+                ret = hsmEvictKey(server, keyId);
+            }
+#endif
```
bigbrett, line 755: Not super clear why we are doing this? Could you explain plz

bigbrett (changes requested):
Nice. A few things and request for clarification but looks solid.

src/wh_server_keystore.c
```diff
@@ -278,36 +473,72 @@ int hsmEvictKey(whServerContext* server, whNvmId keyId)
...
+            slotMeta = server->cache[i].meta;
             break;
         }
     }
+#ifdef WOLFHSM_SEPARATE_BIG_CACHE
+    /* find key in big cache */
+    if (i >= WOLFHSM_NUM_RAMKEYS) {
+        for (i = 0; i < WOLFHSM_NUM_BIG_RAMKEYS; i++) {
+            if (server->bigCache[i].meta->id == keyId) {
+                slotCommited = &server->cache[i].commited;
```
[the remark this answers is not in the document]
jpbland1: good catch

src/wh_server_crypto.c
jpbland1: It's dangerous because I have to pass in a non-NULL nonsense pointer for the length I'm trying to get, the operation will fail as long as at least one of the lengths, qx or qy, is NULL so it should be good but it warrants a comment, might be wise a refactor ecc to allow a `LENGTH_ONLY_E` return

src/wh_server_crypto.c
jpbland1: this one can't be const because it depends on the length of the particular ecc key pulled from the metadata, will make a descriptive variable for it though

src/wh_server_crypto.c
jpbland1: if a user pre-caches an aes key to use with cmac it will be in the regular cache initially, when we cache the cmac struct it will need to go into big cache so the now defunct aes key will be hanging if we don't evict it

billphipps:
Love this idea.  Any chance we can make this option always enabled and simply use a 0 num of big keys if they are not needed?  This would give the same result without a bunch of conditional defines.

Also, I'd like to have the server context use pointers to the cache areas with a config member that indicates the number of slots.  this will keep the server context size more consistent and cause the user to allocate the space themselves, as it may make sense to have this cache area somewhere other than in the middle of the context struct.  just a thought.

bigbrett:
> Any chance we can make this option always enabled and simply use a 0 num of big keys if they are not needed? This would give the same result without a bunch of conditional defines.

Big +1 for this plan. The less code that is macro-ed out, the better! @jpbland1 Could you take a stab?

> Also, I'd like to have the server context use pointers to the cache areas with a config member that indicates the number of slots. this will keep the server context size more consistent and cause the user to allocate the space themselves, as it may make sense to have this cache area somewhere other than in the middle of the context struct. just a thought.

Probably food for a follow-up PR IMO

bigbrett:
@jpbland1 noticed this went stale, could you possibly address Bill's point so we can get it merged:

> Any chance we can make this option always enabled and simply use a 0 num of big keys if they are not needed? This would give the same result without a bunch of conditional defines.

billphipps:
And PR#49 changes the names of config values and provides an info message response as well.  Need to add 2 more entries in the info message and API's to support large keys.  WOLFHSM_CFG_KEYCACHE_LARGE_COUNT/BUFSIZE?

bigbrett (changes requested):
need to address merge conflicts, as well as Bill's points below:

>  Any chance we can make this option always enabled and simply use a 0 num of big keys if they are not needed? This would give the same result without a bunch of conditional defines.
and
> And PR#49 changes the names of config values and provides an info message response as well. Need to add 2 more entries in the info message and API's to support large keys. WOLFHSM_CFG_KEYCACHE_LARGE_COUNT/BUFSIZE?

src/wh_server_keystore.c
```diff
@@ -117,53 +165,99 @@ int hsmCacheKey(whServerContext* server, whNvmMetadata* meta, uint8_t* in)
...
+            }
+        }
+        /* write key if slot found */
+        if (foundIndex != -1) {
+            XMEMCPY((uint8_t*)server->cache[foundIndex].buffer, in, meta->len);
+            XMEMCPY((uint8_t*)server->cache[foundIndex].meta, (uint8_t*)meta,
+                sizeof(whNvmMetadata));
+            /* check if the key is already commited */
+            if (wh_Nvm_GetMetadata(server->nvm, meta->id, meta) ==
+                WH_ERROR_NOTFOUND) {
```
bigbrett, line 203: Probably missing something, but how come we go to NVM here? Can we not just check `server->cache[foundIndex].commited == 1`? Seems like it would prevent a potentially lengthy search through NVM, no?

src/wh_server_keystore.c
```diff
@@ -117,53 +165,99 @@ int hsmCacheKey(whServerContext* server, whNvmMetadata* meta, uint8_t* in)
...
+        }
+        /* write key if slot found */
+        if (foundIndex != -1) {
+            XMEMCPY((uint8_t*)server->bigCache[foundIndex].buffer, in,
+                meta->len);
+            XMEMCPY((uint8_t*)server->bigCache[foundIndex].meta, (uint8_t*)meta,
+                sizeof(whNvmMetadata));
+            /* check if the key is already commited */
+            if (wh_Nvm_GetMetadata(server->nvm, meta->id, meta) ==
+                WH_ERROR_NOTFOUND) {
```
bigbrett, line 238: same question as above

src/wh_server_crypto.c
```diff
@@ -782,6 +798,10 @@ static int hsmCryptoCmac(whServerContext* server, whPacket* packet,
                 keyId = WH_MAKE_KEYID(WH_KEYTYPE_CRYPTO,
                     server->comm->client_id, packet->cmacReq.keyId);
             }
+            /* evict the aes sized key in the normal cache */
+            if (moveToBigCache == 1) {
+                ret = hsmEvictKey(server, keyId);
+            }
```
bigbrett, line 804: so it seems that CMAC support now hinges on `WOLFHSM_CFG_SERVER_KEYCACHE_BIG_COUNT` being > 0, since the CMAC struct is cached across invocations. Since WOLFHSM_CFG_SERVER_KEYCACHE_BIG_COUNT is configurable by the user, and is very much allowed to be 0, I think we probably need to implement some sort of compile time checking for this? Otherwise it will (best case) result in potentially hard to diagnose runtime errors

bigbrett (changes requested):
Might have noticed an issue with the coupling between CMAC and enablement of the big cache, lmk ur thoughts

src/wh_server_keystore.c
jpbland1: keys are cached anytime they are read, if you look in hsmReadKey:
```
    /* cache key if free slot, will only kick out other commited keys */
    if (ret == 0 && out != NULL) {
        hsmCacheKey(server, meta, out);
    }

```
we cache a key after reading it in case it was read from NVM so we don't have to keep reading from nvm every time it's accessed since odds are if they're reading it once they'll read it again soon for more operations, thus we need to check nvm when caching to see if it's already in there

src/wh_server_crypto.c
jpbland1: as long as the normal cache is big enough to hold the cmac struct it will work fine since `hsmEvictKey` and `hsmCacheKey` transparently put the key into the correct cache. if `WOLFHSM_CFG_SERVER_KEYCACHE_BIG_COUNT` is 0 and `WOLFHSM_CFG_SERVER_KEYCACHE_BUFSIZE` is big enough to hold a cmac struct it will work fine, if it's not big enough `WH_ERROR_NOSPACE` will be returned

src/wh_server_keystore.c
billphipps: For the nvm_flash implementation, wh_Nvm_GetMetadata copies the metadata from the directory in memory.  All of the valid metadata is read only once during init.  It shouldn't take very long to do that.

bigbrett (approved):

billphipps (approved):
Got some reservations about this scheme in general, but I'm going to merge this because it does help with handling large keys.  Thanks for putting this together!

PR #49, opened 2024-06-26
add get key ABI functions

jpbland1:


bigbrett (changes requested):
Real nit here, but could we call the functions get/set keyID instead, since that is what it is doing? Always nice to have the functions fully self describing

bigbrett (approved):

PR #52, opened 2024-07-13
Move to user_settings and WOLFHSM_CFG_*.

billphipps:
Further isolation of compile-time configuration values and moving to using user_settings.h to be included in those headers that support reconfiguration (wh_client.h, wh_server.h, wh_nvm.h, wh_she_common.h).

A number of additional typos and minor doc edits as well.


src/wh_server.c
```diff
@@ -184,6 +184,39 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
...
+        resp.cfg_server_customcb_count = WOLFHSM_CFG_SERVER_CUSTOMCB_COUNT;
+        resp.cfg_server_dmaaddr_count = WOLFHSM_CFG_SERVER_DMAADDR_COUNT;
+        resp.cfg_server_keycache_bufsize = WOLFHSM_CFG_SERVER_KEYCACHE_BUFSIZE;
+        resp.cfg_server_keycache_count = WOLFHSM_CFG_SERVER_KEYCACHE_COUNT;
+
+        /* III Growth */
+        resp.debug_state = 1;
+        resp.boot_state = 2;
+        resp.lifecycle_state = 3;
+        resp.nvm_state = 4;
```
bigbrett, line 211: are these just dummy values?

src/wh_server.c
```diff
@@ -289,14 +321,14 @@ int wh_Server_HandleRequestMessage(whServerContext* server)
             rc = wh_Server_HandleCryptoRequest(server, action, data,
                 &size, seq);
         break;
-#endif  /* WOLFHSM_NO_CRYPTO */
+#endif  /* WOLFHSM_CFG_NO_CRYPTO */
```
bigbrett, line 324:
```suggestion
#endif  /* ! WOLFHSM_CFG_NO_CRYPTO */
```

wolfhsm/wh_server.h
```diff
@@ -227,14 +268,14 @@ typedef struct whServerConfig_t {
...
     whServerCryptoContext* crypto;
-    whServerCacheSlot       cache[WOLFHSM_NUM_RAMKEYS];
-#ifdef WOLFHSM_SHE_EXTENSION
+    whServerCacheSlot       cache[WOLFHSM_CFG_SERVER_KEYCACHE_COUNT];
+#ifdef WOLFHSM_CFG_SHE_EXTENSION
     whServerSheContext* she;
 #endif
-#endif /* WOLFHSM_NO_CRYPTO */
-    whServerCustomCb   customHandlerTable[WH_CUSTOM_CB_NUM_CALLBACKS];
+#endif /* WOLFHSM_CFG_NO_CRYPTO */
```
bigbrett, line 277:
```suggestion
#endif /* ! WOLFHSM_CFG_NO_CRYPTO */
```

wolfhsm/wh_server.h
```diff
@@ -42,54 +73,65 @@ typedef struct whServerContext_t whServerContext;
...
-#define WH_CUSTOM_CB_NUM_CALLBACKS 8
+#ifndef WOLFHSM_CFG_SERVER_CUSTOMCB_COUNT
+#define WOLFHSM_CFG_SERVER_CUSTOMCB_COUNT 8
 #endif
 
 /* DMA translation allow entries */
-#ifndef WH_DMA_ADDR_ALLOWLIST_COUNT
-#define WH_DMA_ADDR_ALLOWLIST_COUNT 10
+#ifndef WOLFHSM_CFG_SERVER_DMAADDR_COUNT
+#define WOLFHSM_CFG_SERVER_DMAADDR_COUNT 10
```
bigbrett, line 125: small opinion/nit, feel free to disregard: I kind of like keeping the "allow" word in these defines, because it really indicates the number of addresses in the allowlist.

wolfhsm/wh_server.h
```diff
@@ -42,54 +73,65 @@ typedef struct whServerContext_t whServerContext;
...
+#ifndef WOLFHSM_CFG_NO_CRYPTO
 #include "wolfssl/wolfcrypt/settings.h"
+#include "wolfssl/wolfcrypt/types.h"
 #include "wolfssl/wolfcrypt/random.h"
 #include "wolfssl/wolfcrypt/rsa.h"
 #include "wolfssl/wolfcrypt/ecc.h"
 #include "wolfssl/wolfcrypt/curve25519.h"
 #include "wolfssl/wolfcrypt/cryptocb.h"
-#endif /* WOLFHSM_NO_CRYPTO */
+#endif /* WOLFHSM_CFG_NO_CRYPTO */
```
bigbrett, line 84:
```suggestion
#endif /* ! WOLFHSM_CFG_NO_CRYPTO */
```

wolfhsm/wh_server.h
```diff
@@ -42,54 +73,65 @@ typedef struct whServerContext_t whServerContext;
...
-#ifndef WOLFHSM_NUM_RAMKEYS
-#define WOLFHSM_NUM_RAMKEYS  8
+/* Reported version string */
+#ifndef WOLFHSM_CFG_INFOVERSION
+#define WOLFHSM_CFG_INFOVERSION "01.01.01"
+#endif
+
+/* Reported build identifier string */
+#ifndef WOLFHSM_CFG_INFOBUILD
+#define WOLFHSM_CFG_INFOBUILD "12345678"
```
bigbrett, line 99: no action, just a reminder to myself for the future: Might be nice to pull versioning into a separate `wh_version.h` header for easier post-processing with tooling, automagically including the git hash into the binary when built on CI for release, etc

wolfhsm/wh_server.h
```diff
@@ -19,14 +19,45 @@
...
+ *      Default: 10
+ *
  */
 
 #ifndef WOLFHSM_WH_SERVER_H_
 #define WOLFHSM_WH_SERVER_H_
 
+#ifdef WOLFSSL_USER_SETTINGS
+#include "user_settings.h"
+#endif
```
bigbrett, line 57: see `user_settings.h` comment

wolfhsm/wh_packet.h
```diff
@@ -15,15 +14,22 @@
...
+ *
  */
-#ifndef WOLFHSM_PACKET_H
-#define WOLFHSM_PACKET_H
+#ifndef WOLFHSM_WH_PACKET_H_
+#define WOLFHSM_WH_PACKET_H_
+
+#ifdef WOLFSSL_USER_SETTINGS
+#include "user_settings.h"
+#endif
```
bigbrett, line 28: see `user_settings.h` comment

wolfhsm/wh_message_nvm.h
```diff
@@ -47,9 +48,8 @@ enum WH_MESSAGE_NVM_ACTION_ENUM {
 
 enum WH_MESSAGE_NVM_MAX_ENUM {
     /* must be odd for struct whMessageNvm_DestroyObjectsRequest  alignment */
-    WH_MESSAGE_NVM_MAX_DESTROY_OBJECTS_COUNT = 9,
-    WH_MESSAGE_NVM_MAX_ADD_OBJECT_LEN =
-            WH_COMM_DATA_LEN - sizeof(whNvmMetadata),
+    WH_MESSAGE_NVM_MAX_DESTROY_OBJECTS_COUNT = 19,
```
bigbrett, line 51: change from 9->19?

wolfhsm/wh_comm.h
```diff
@@ -69,7 +81,7 @@ enum WH_COMM_ENUM {
     (_magic                 & WH_COMM_MAGIC_ENDIAN_MASK) ==  \
     (WH_COMM_MAGIC_NATIVE   & WH_COMM_MAGIC_ENDIAN_MASK)
 
-/* Header for a packet, request or response. On-the-wire format */
+/* 8 byte Header for a packet, request or response. On-the-wire format */
```
bigbrett, line 84: I want static asserts :(

wolfhsm/wh_comm.h
```diff
@@ -34,25 +34,37 @@
...
-    WH_COMM_DATA_LEN = 1280,
-    WH_COMM_MTU = (WH_COMM_HEADER_LEN + WH_COMM_DATA_LEN),
-    WH_COMM_MTU_U64_COUNT = (WH_COMM_MTU + 7) / 8,  /* internal U64 buffer */
-};
+#ifndef WOLFHSM_CFG_COMM_DATA_LEN
+#define WOLFHSM_CFG_COMM_DATA_LEN 1280
+#endif
+
+/* III Previous name.  Update */
+#define WH_COMM_DATA_LEN WOLFHSM_CFG_COMM_DATA_LEN
```
bigbrett, line 64: lingering TODO?

test/wh_test_crypto.c
```diff
@@ -918,15 +915,15 @@ static int wh_ClientServer_MemThreadTest(void)
...
-#if defined(WH_CFG_TEST_POSIX)
+#if defined(WOLFHSM_CFG_TEST_POSIX)
     printf("Testing crypto: (pthread) mem...\n");
     WH_TEST_RETURN_ON_FAIL(wh_ClientServer_MemThreadTest());
 #endif
     return 0;
 }
 
-#endif  /* WOLFHSM_NO_CRYPTO */
+#endif  /* WOLFHSM_CFG_NO_CRYPTO */
```
bigbrett, line 929:
```suggestion
#endif  /* ! WOLFHSM_CFG_NO_CRYPTO */
```

test/user_settings.h
```diff
@@ -29,24 +29,38 @@
...
+#define WOLFHSM_CFG_TEST_POSIX
+
+/** wolfHSM settings.  Simple overrides to show they work */
+#define WOLFHSM_CFG_COMM_DATA_LEN 1152
+
+#define WOLFHSM_CFG_NVM_OBJECT_COUNT 30
+#define WOLFHSM_CFG_SERVER_KEYCACHE_COUNT 10
+#define WOLFHSM_CFG_SERVER_KEYCACHE_SIZE 1024
+#define WOLFHSM_CFG_SERVER_DMAADDR_COUNT 8
+#define WOLFHSM_CFG_SERVER_CUSTOMCB_COUNT 6
```
bigbrett, line 43: While this will work for our tests, I don't know if we want to establish the paradigm of putting wolfHSM configuration defines in `user_settings.h`. They should be separate. What if we don't want to use a wolfSSL user settings build? Then `WOLFSSL_USER_SETTINGS` won't be defined, and this header won't be included. 

I think we want a separate wolfHSM user settings file (how I had `wh_config.h`**) so it isn't tied to how wolfSSL is built. 

** or could rename to `wh_user_settings.h` if we want coherency with wolfSSL naming, but since our macros are `WOLFHSM_CFG` then a `wh_config.h` makes sense to me

src/wh_server_keystore.c
```diff
@@ -431,4 +436,4 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
     return 0;
 }
 
-#endif  /* WOLFHSM_NO_CRYPTO */
+#endif  /* WOLFHSM_CFG_NO_CRYPTO */
```
bigbrett, line 439:
```suggestion
#endif  /* ! WOLFHSM_CFG_NO_CRYPTO */
```

src/wh_server_crypto.c
```diff
@@ -911,4 +916,4 @@ int wh_Server_HandleCryptoRequest(whServerContext* server,
     }
     return ret;
 }
-#endif  /* WOLFHSM_NO_CRYPTO */
+#endif  /* WOLFHSM_CFG_NO_CRYPTO */
```
bigbrett, line 919:
```suggestion
#endif  /* ! WOLFHSM_CFG_NO_CRYPTO */
```

bigbrett (approved):
A few random things I noticed but good to merge as is. 

Something that I do want to discuss though (also mentioned inline in a comment below):
 - I don't think we want to establish the paradigm of putting wolfHSM configuration defines in a wolfSSL `user_settings.h`
 - We should retain a separate user configuration header for wolfHSM, as supplying defines on CLI would get unwieldy.

What was wrong with `wh_config.h`?

src/wh_server.c
billphipps: Yup.  Hence the growth above.  Should eventually turn into well-defined constants or bit fields, likely with mappings to SHE and PKCS11 concepts.

wolfhsm/wh_server.h
billphipps: amen!

wolfhsm/wh_message_nvm.h
billphipps: If the MTU has to be no smaller than about 200 bytes, there's really no reason not to expand this a bit.

wolfhsm/wh_comm.h
billphipps: oops!  Missed it.  I'll go back and fix it in the tests and examples too.

test/user_settings.h
billphipps: Yeah, was on the fence about this.  Really, we should have 2 different wolfHSM config files: wolfhsm_cfg_client.h and wolfhsm_cfg_server.h.  Since each side will/should have a separate user_settings, it made sense to steal some space in the same file to avoid any chances of misconfiguration.  

Lemme rework this into 2 files and pull out the CFG_TEST items since these are set by the Makefile.

wolfhsm/wh_comm.h
billphipps: We should add this sizeof() check to the check padding.  prolly lots more too.

test/user_settings.h
billphipps: Corrected to wh_settings.h and wolfhsm_cfg.h.  Much cleaner solution.  Thanks for the guidance!

bigbrett (approved):
LGTM nice work. I'm wondering if we want to be including `wh_settings.h` in every header when it isn't needed externally - think it would be best to keep that inclusion in the `*.c` file and only export it if it is needed for the external API. But that is not a critical item.

billphipps:
---I'm wondering if we want to be including wh_settings.h in every header when it isn't needed externally

I decided to err on the side of caution.  If EVERY .h file in the base library includes wh_settings.h first, then IF some change later needs a new CFG value, it will already be there.  This also means if a customer includes a random wh_XXX.h file, we are assured those configs will be included.

PR #53, opened 2024-07-19
fix missing ecc pub key support and outlen check

jpbland1:


billphipps (approved):
Looks good!  Can you add some tests for the new functions in a future PR?  I'd like to merge this real quick, but the test is hanging.

PR #54, opened 2024-07-22
use non-fixed block size so we're not hashing data

jpbland1:
after the message if it's length is not a multiple of block size


billphipps:
Grumble.  Test is handling again.  Lemme look into this and see if we can't get a fix put in here.

jpbland1:
I'm not sure what's causing the hang but I think it's only on the SHE build before the crypto tests actually start, possibly an issue with the bootloader check logic

bigbrett (approved):

bigbrett:
@billphipps Is this okay to merge, since CI is green?

PR #57, opened 2024-07-23
Fix settings include file order in SHE module

bigbrett:
Another quick fix required by tricore integration: add wh_settings.h include outside macro protection for SHE source code, so it can conditionally compile in/out remainder of code

### DIFF
--- a/src/wh_she_common.c
+++ b/src/wh_she_common.c
@@ -22,6 +22,9 @@
  */
 /* System libraries */
 
+/* Pick up compile-time configuration */
+#include "wolfhsm/wh_settings.h"
+
 #ifdef WOLFHSM_CFG_SHE_EXTENSION
 
 #include <stdint.h>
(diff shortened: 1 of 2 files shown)


src/wh_she_common.c
```diff
@@ -22,6 +22,9 @@
  */
 /* System libraries */
 
+/* Pick up compile-time configuration */
+#include "wolfhsm/wh_settings.h"
+
 #ifdef WOLFHSM_CFG_SHE_EXTENSION
```
billphipps, line 28: Why not simply move the ifdef SHE_EXTENSION below the any of the wh_*.h includes below?  I was hoping to avoid having any reference to wh_settings.h in the .c files.

src/wh_she_crypto.c
```diff
@@ -21,6 +21,8 @@
  *
  */
 
+/* Pick up compile-time configuration */
+#include "wolfhsm/wh_settings.h"
 
 #ifdef WOLFHSM_CFG_SHE_EXTENSION
```
billphipps, line 27: Same comment here.  Simply move the ifdef to below the any of the wh_*.h

billphipps (approved):

PR #58, opened 2024-07-24
add message to try and find hang

jpbland1:


jpbland1:
I refactored the way serverDelay works, now it holds the server before cmac update is called and resumes it after cancel request is called. once the cancel request is sent it resumes the server before calling cancel response so the server can actually respond. previously i was using the regular cancel function which was waiting for a blocked server to send the canceled message header.

bigbrett (approved):
tested on tc3xx

PR #60, opened 2024-07-31
clean up README

bigbrett:
oops we forgot to scrub the partially finished README


billphipps (commented):
Looks good.  We should eliminate the column limit for documentation files and remote all of the extra newlines.  No changes for now

billphipps (approved):
Meant to approve.

PR #61, opened 2024-08-02
Add wolfCrypt tests to unit tests

bigbrett:
Creates a unit test to run wolfCrypt tests on top of wolfHSM, and adds as a CI job. 

This will keep failing CI until we can actually pass wolfCrypt tests, so I've got it **off by default in the unit tests** (use `TESTWOLFCRYPT=1` to enable), and have it set to pass CI even though unit tests fail. 

Once we are able to pass wolfCrypt tests, we can remove the `||true` in the CI job. For now, it is just nice to have as a progress indicator.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -29,9 +29,14 @@ jobs:
         repository: wolfssl/wolfssl
         path: wolfssl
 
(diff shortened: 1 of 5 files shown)


billphipps (approved):
Looks good!

PR #62, opened 2024-08-09
Sha256 support

bigbrett:
Adds non-DMA sha256 support to wolfHSM crypto. Intermediate SHA state is held by the client and passed back and forth on each SHA256 API call. Passes tests and wolfCrypt tests

Note: The message in `wh_packet.h` are fixed size (with no padding needed the way they are implemented), and so I didn't use the "trailing implicit buffer" scheme that the other messages use, as it adds some more complicated array indexing logic since there are two optional buffers that would need to be passed back and forth. This results in a few extra unused bytes being sent on the first update message (unused resume hash state), and potentially on the finalize message (remainder of the SHA block not used for data). 

I can go back and modify this to use the implicit buffers and variable length data if needed, but the code is a little simpler this way so figured I would see what you though first.


src/wh_server_crypto.c
```diff
@@ -841,6 +848,57 @@ static int hsmCryptoCmac(whServerContext* server, whPacket* packet,
...
+    wh_Packet_hash_sha256_res* res    = &packet->hashSha256Res;
+
+    /* THe server SHA256 struct doesn't persist state (it is a union), meaning
+     * the devId may get blown away between calls. We must restore the server
+     * devId each time */
+    sha256->devId = server->crypto->devId;
+
+    /* Init the SHA256 context if this is the first time, otherwise restore the
+     * hash state from the client */
+    if (req->resumeState.resumeHashFlag == 0) {
```
billphipps, line 867: should this be a bitfield like RESUME_HASH_BIT_FLAG?  Not sure we really need anything else in a flag for SHA2

Wouldn't the server already know if this is a resumption based on the hi_len/lo_len > BLOCK_SIZE?

src/wh_client_cryptocb.c
```diff
@@ -849,4 +875,157 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
+    }
+
+    ret = wh_Client_SendRequest(
+        ctx, group, WC_ALGO_TYPE_HASH,
+        WH_PACKET_STUB_SIZE + sizeof(packet->hashSha256Req), (uint8_t*)packet);
+
+#ifdef DEBUG_CRYPTOCB_VERBOSE
+    printf("[client] send SHA256 Req:\n");
+    _hexdump("[client] inBlock: ", req->inBlock, WC_SHA256_BLOCK_SIZE);
+    if (req->resumeState.resumeHashFlag) {
```
billphipps, line 987:
```suggestion
    if (req->resumeState.resumeHashFlag != 0) {
```

wolfhsm/wh_packet.h
```diff
@@ -263,6 +263,37 @@ typedef struct  wh_Packet_cmac_res
...
+        /* intermediate hash value */
+        uint8_t hash[32]; /* TODO (BRN) WC_SHA256_DIGEST_SIZE */
+    } resumeState;
+    /* Flag indicating to the server that this is the last block and it should
+     * finalize the hash. If set, inBlock may be only partially full*/
+    uint32_t isLastBlock;
+    /* Length of the last input block of data. Only valid if isLastBlock=1 */
+    uint32_t lastBlockLen;
+    /* Full sha256 input block to hash */
+    uint8_t inBlock[64]; /* TODO (BRN) WC_SHA256_BLOCK_SIZE */
```
billphipps, line 287: Why not use the WC_SHA256_BLOCK_SIZE?  Same of other todos

src/wh_server_crypto.c
```diff
@@ -965,6 +1023,32 @@ int wh_Server_HandleCryptoRequest(whServerContext* server,
...
+                ret = hsmCryptoSha256(server, (whPacket*)data, size);
+#ifdef DEBUG_CRYPTOCB_VERBOSE
+                if (ret != 0) {
+                    printf("[server] SHA256 ret = %d\n", ret);
+                }
+#endif
+                break;
+
+            default:
+                ret = NOT_COMPILED_IN;
```
billphipps, line 1043: Hrmmm... Had assumed we'd return CRYPTOCB_UNAVAILABLE, but this is more accurate since we are not necessarily in a cryptocb (client could just send the message).  Should we recheck how/if we want to handle this differently?

src/wh_client_cryptocb.c
```diff
@@ -849,4 +875,157 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
+                                           uint32_t isLastBlock)
+{
+    uint16_t                   group  = WH_MESSAGE_GROUP_CRYPTO;
+    uint16_t                   action = WH_MESSAGE_ACTION_NONE;
+    int                        ret    = 0;
+    uint16_t                   dataSz = 0;
+    wh_Packet_hash_sha256_req* req    = &packet->hashSha256Req;
+
+    /* Unused wolfCrypt hash flag value. TODO: use devCtx? */
+    const uint32_t RESUME_HASH_BIT_FLAG = 0x80000000;
```
billphipps, line 952: Double check that there is case where hiLen/loLen > BLOCK_SIZE but RESUME is not set yet. Seems like we can simplify the compile and logic here if not.

billphipps (approved):
Looks good!  I had a question on the need for the resume flag (and sha256->flags then).  If you think its as simple as it can be, then I'll merge.

src/wh_server_crypto.c
bigbrett: I think it makes sense as written, right? We are essentially saying that the wolfHSM server doesn't support the client request for that algorithm, since support for that algorithm isn't compiled in.

src/wh_server_crypto.c
bigbrett: Ah okay, I remember why I did it this way - the empty hash. Sha256 hash of 0 length input data does result in a valid digest and my thoughts were it would need to be held by the client too. But I forgot that a zero length message wouldn't be even sent across the wire since we only send full blocks, and the way sha256 works is that the zero length message, though it produces a valid hash on finalization, doesn't affect the internal state of the hash at all. So it wasn't needed like you said.

src/wh_server_crypto.c
billphipps: Concur.  I'm thinking we should extend wolfHSM error codes to support this status.

wolfhsm/wh_packet.h
billphipps: Understood.  Recommend add additional #define's in here if you want to keep these sizes statically defined.

billphipps (approved):
Excellent simplification.  We can clean up the fixed sizes later.

PR #65, opened 2024-08-13
fix cryptoCb cipher macro

bigbrett:
Fixes missing macro protection

wolfCrypt [guards the definition](https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/cryptocb.h#L280-L283) of wcCryptoInfo->cipher by `#if !defined(NO_AES) || !defined(NO_DES3)`, but wolfHSM doesn't, resulting in a compile error when using wolfCrypt with no support for those algos (e.g. default wolfBoot)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .DS_Store
 Build/
+*.o
+*.a
+*.la
 
--- a/src/wh_client_cryptocb.c
+++ b/src/wh_client_cryptocb.c
@@ -98,6 +98,7 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
     /* Based on the info type, process the request */
     switch (info->algo_type)
     {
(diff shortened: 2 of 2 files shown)


jpbland1 (changes requested):
There are non aes ciphers within cryptocb, cha cha comes to mind, that we'll support when a customer requests it so the top level cipher shouldn't be linked to aes

bigbrett:
> There are non aes ciphers within cryptocb, cha cha comes to mind, that we'll support when a customer requests it so the top level cipher shouldn't be linked to aes

The point is we aren't guarding it the same as the definition of the struct in `cryptocb.h`. This just protects our cryptocb's usage of that struct using the same macros as the definition

jpbland1:
nevermind, I had it wrong, looks good

jpbland1 (approved):
looks good

PR #68, opened 2024-08-20
Fix memtransport, move debug prints to utils

billphipps:
First stage of a sequence of fixes.  This one adds cache flush/mem fencing support to mem transport, stolen from the Bernina private repo, which corrects the surprisingly relaxed memory consistency on Apple silicon when running the sim.

### DIFF
--- a/src/wh_transport_mem.c
+++ b/src/wh_transport_mem.c
@@ -30,6 +30,7 @@
 #include <stdint.h>
 
 #include "wolfhsm/wh_error.h"
+#include "wolfhsm/wh_utils.h"
 #include "wolfhsm/wh_comm.h"
 
 #include "wolfhsm/wh_transport_mem.h"
@@ -51,7 +52,7 @@ int wh_TransportMem_Init(void* c, const void* cf,
         return WH_ERROR_BADARGS;
     }
 
-    memset(context, 0, sizeof(*context));
+    wh_Utils_memset_flush(context, 0, sizeof(*context));
     context->req            = (whTransportMemCsr*)config->req;
     context->req_size       = config->req_size;
     context->req_data       = (void*)(context->req + 1);
@@ -61,6 +62,8 @@ int wh_TransportMem_Init(void* c, const void* cf,
     context->resp_data      = (void*)(context->resp + 1);
 
     context->initialized = 1;
+    XMEMFENCE();
+
     return WH_ERROR_OK;
 }
 
@@ -72,8 +75,8 @@ int wh_TransportMem_InitClear(void* c, const void* cf,
     int rc = wh_TransportMem_Init(c, cf, connectcb, connectcb_arg);
     if (rc == WH_ERROR_OK) {
         /* Zero the buffers */
-        memset((void*)context->req, 0, context->req_size);
-        memset((void*)context->resp, 0, context->resp_size);
+        wh_Utils_memset_flush((void*)context->req, 0, context->req_size);
+        wh_Utils_memset_flush((void*)context->resp, 0, context->resp_size);
     }
     return rc;
 }
@@ -93,6 +96,8 @@ int wh_TransportMem_Cleanup(void* c)
 int wh_TransportMem_SendRequest(void* c, uint16_t len, const void* data)
 {
     whTransportMemContext* context = c;
+    volatile whTransportMemCsr* ctx_req = context->req;
+    volatile whTransportMemCsr* ctx_resp = context->resp;
     whTransportMemCsr resp;
     whTransportMemCsr req;
 
@@ -101,9 +106,13 @@ int wh_TransportMem_SendRequest(void* c, uint16_t len, const void* data)
         return WH_ERROR_BADARGS;
     }
 
-    /* Read current CSR's */
-    resp.u64 = context->resp->u64;
-    req.u64 = context->req->u64;
+    /* Read current CSR's. ctx_req does not need to be invalidated */
+    XMEMFENCE();
+    XCACHEINVLD(ctx_resp);
+    XCACHEINVLD(ctx_req);
+
+    resp.u64 = ctx_resp->u64;
+    req.u64 = ctx_req->u64;
 
     /* Has server completed with previous request */
     if (req.s.notify != resp.s.notify) {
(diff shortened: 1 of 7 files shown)


src/wh_transport_mem.c
```diff
@@ -101,30 +106,43 @@ int wh_TransportMem_SendRequest(void* c, uint16_t len, const void* data)
...
         return WH_ERROR_NOTREADY;
     }
 
     if ((data != NULL) && (len != 0)) {
-        memcpy((void*)context->req_data, data, len);
+        wh_Utils_memcpy_flush((void*)context->req_data, data, len);
     }
+
+    /* Ensure the memcpy is complete */
+    XMEMFENCE();
```
bigbrett, line 127: If we are always fencing after a memcpy, should we just pull this into `wh_Utils_memcpy_flush()`?

src/wh_utils.c
```diff
@@ -84,3 +86,71 @@ int wh_Utils_memeqzero(uint8_t* buffer, uint32_t size)
...
+{
+    int len = (int)n;
+    uint8_t* ptr = (uint8_t*)p;
+    do {
+        XCACHEFLUSH(ptr);
+        ptr += XCACHELINE;
+        len -= XCACHELINE;
+    } while (len > 0);
+    return p;
+}
```
bigbrett, line 113: I wonder if this should be a callback function left up to the port.... For example TC3xx lets you flush/invalidate the entire cache, a wordline, or a block, but the HSM cache is totally different.

src/wh_utils.c
```diff
@@ -84,3 +86,71 @@ int wh_Utils_memeqzero(uint8_t* buffer, uint32_t size)
...
+
+void* wh_Utils_memset_flush(void* p, int c, size_t n)
+{
+    return wh_Utils_CacheFlush(memset(p, c, n), n);
+}
+
+void* wh_Utils_memcpy_invalidate(void* dst, const void* src, size_t n)
+{
+    return memcpy(dst,wh_Utils_CacheInvalidate(src,n),n);
```
bigbrett, line 122: space separate args

src/wh_utils.c
```diff
@@ -84,3 +86,71 @@ int wh_Utils_memeqzero(uint8_t* buffer, uint32_t size)
...
+}
+
+void* wh_Utils_memcpy_flush(void* dst, const void* src , size_t n)
+{
+    return wh_Utils_CacheFlush(memcpy(dst,src,n),n);
```
bigbrett, line 127: space separate args

src/wh_utils.c
```diff
@@ -84,3 +86,71 @@ int wh_Utils_memeqzero(uint8_t* buffer, uint32_t size)
...
+}
+
+
+#if defined(DEBUG_CRYPTOCB) || defined(DEBUG_CRYPTOCB_VERBOSE)
+#include <stdio.h>
+#endif
+
+#ifdef DEBUG_CRYPTOCB_VERBOSE
+void _hexdump(const char* initial, uint8_t* ptr, size_t size)
```
bigbrett, line 136: no longer static, so might as well rename to `wh_Utils_Hexdump`

wolfhsm/wh_utils.h
```diff
@@ -39,4 +40,50 @@ uint32_t wh_Utils_ntohl(uint32_t networklong);
 
 int wh_Utils_memeqzero(uint8_t* buffer, uint32_t size);
 
+/** Cache flushing and memory fencing synchronization primitives */
+/* Create a full sequential memory fence to ensure compiler memory ordering */
+#ifndef XMEMFENCE
+#define XMEMFENCE() __atomic_thread_fence(__ATOMIC_SEQ_CST)
```
bigbrett, line 46: This is GCC specific - either needs to have a tree of compiler-specific #ifdefs or should be instead implemented by ports in `wolfhsm_config.h`?

wolfhsm/wh_utils.h
```diff
@@ -39,4 +40,50 @@ uint32_t wh_Utils_ntohl(uint32_t networklong);
...
+#define XCACHEFLUSH(_p) (void)(_p)
+/* PPC32: __asm__ volatile ("dcbf 0, %0" : : "r" (_p): "memory") */
+#endif
+
+/* Invalidate the cache line at _p. Used prior to reading to ensure
+ * freshness. */
+#ifndef XCACHEINVLD
+#define XCACHEINVLD(_p) (void)(_p)
+/* PPC32: __asm__ volatile ("dcbi 0, %0" : : "r" (_p): "memory") */
+#endif
```
bigbrett, line 66: same thing as above for all these - we should discuss if they should live in `wh_settings.h` and document the best way for ports to define

bigbrett (changes requested):

src/wh_transport_mem.c
billphipps: This is a bug.  The fence needed to be after the memset/cpy but before the flush.  Updated.

src/wh_utils.c
billphipps: good call.  updated.

src/wh_utils.c
billphipps: fixed

src/wh_utils.c
billphipps: fixed

wolfhsm/wh_utils.h
billphipps: Agree.  This one is good for gcc's and clang and probably others too.  Move to settings.h

wolfhsm/wh_utils.h
billphipps: agreed. Moved already.

src/wh_utils.c
billphipps: Done

bigbrett (approved):
LGTM, @billphipps has this been tested on SPC?

billphipps:
I did not test on the SPC.  Maybe I'll fire that up and make sure I didn't break anything.  I'll undo the custom transport and use this instead.

bigbrett (commented):
Tested on TC3xx and works.

bigbrett (approved):

PR #69, opened 2024-08-26
Posix shmem transport

bigbrett:
Added a new example transport to POSIX port using posix shared memory objects. Functions as a simple wrapper around the built-in shared memory transport, but using `mmap()`ed shared objects (e.g. `/dev/shm`), allowing for the client and server to be in different processes / address spaces. Useful for scenarios where you want to test logic on in a simulated environment that would otherwise use shared memory buffers IRL (e.g. wolfBoot simulator communicating with wolfHSM server over shared memory)

Started down this road as a means to test DMA-based crypto on the simulator, but realized it doesn't actually buy us anything that the built-in transport wouldn't since the only thing memmapped between the two processes is the transport buffer (and not the actual stuff needing to be DMA'ed! oops...). Nonetheless, perhaps this is still something useful to add in, if nothing else, just for example purposes. I could see customers wanting to implement a transport of their own along these lines.

Note that the server creates the shared memory object on transport initialization, and the client init will block until the server has created it to prevent race conditions, and to keep the logic simple since this is just a toy example. The client then clears the shared memory once it has access.


billphipps (approved):
Great building block for us to add on to.  Let's get this in and bang on it

PR #71, opened 2024-08-29
DMA cryptoCallbacks with SHA256

bigbrett:
- Creates separate DMA client-side crypto callback, selectable at runtime by using the new `WH_DEV_ID_DMA`
- Makes library DMA support opt-in via `WOLFHSM_CFG_DMA` config option
- Enforces simplification that server can only be compiled to support one DMA pointer size, meaning selection of 32 vs 64 bit implementations of certain functions now is done at compile time. **Note:** By default, this is enforced to be the same pointer size as the server's native architecture (e.g. `sizeof(uintptr_t)`) unless overridden with the `WOLFHSM_CFG_DMA_ALT_PTR_SIZE` option. Given that address translation is entirely handled by port-supplied callbacks, I don't see why this wouldn't work, but leaving it off by default for now
- Adds SHA256 DMA implementation


bigbrett:
@billphipps verified on TC3xx

src/wh_client_nvm.c
```diff
@@ -833,51 +838,47 @@ int wh_Client_NvmAddObjectDma64(whClientContext* c,
...
-    if (sizeof(intptr_t) == sizeof(uint32_t)) {
-        return wh_Client_NvmAddObjectDma32Request(c,
-                (intptr_t)metadata, data_len, (intptr_t)data);
-    }
-    if (sizeof(intptr_t) == sizeof(uint64_t)) {
-        return wh_Client_NvmAddObjectDma64Request(c,
-                (intptr_t)metadata, data_len, (intptr_t)data);
-    }
+#if WH_DMA_IS_32BIT
+    return wh_Client_NvmAddObjectDma32Request(c, (intptr_t)metadata, data_len,
```
billphipps, line 849: Consider simply #defining NvmAddObjectDmaRequest to be NAODR32 or 64 and eliminate this function?

test/wh_test.c
```diff
@@ -46,12 +46,14 @@ int whTest_Unit(void)
 {
     printf("Enter unit tests\n");
     /* Component Tests */
+#if 0
```
billphipps, line 49: ?  Are the busted or sumthin?

test/wh_test_crypto.c
```diff
@@ -67,6 +67,16 @@ enum {
 
 #define PLAINTEXT "mytextisbigplain"
 
+/* Array allowing you to loop over available devIds */
+/* TODO: find a better way to do this - perhaps best to define a global LUT
+ * holding the devIds and refer to them by enumerated index rather than directly
+ */
+#ifdef WOLFHSM_CFG_DMA
+static const int DEV_IDS[WH_NUM_DEVIDS] = {WH_DEV_ID, WH_DEV_ID_DMA};
```
billphipps, line 75: why not something like:
static const int DEV_IDS[] = {x,a,b,};
#define WH_NUM_DEVIDS (sizeof(DEV_IDS) / sizeof(int))

devid == 0 is a valid id

wolfhsm/wh_settings.h
```diff
@@ -209,4 +209,58 @@
...
+#endif
+#endif /* WH_PTR_SIZE */
+
+
+#ifndef WOLFHSM_CFG_DMA_PTR_SIZE
+#define WOLFHSM_CFG_DMA_PTR_SIZE (WH_PTR_SIZE)
+#endif
+
+#define WH_DMA_IS_32BIT (WOLFHSM_CFG_DMA_PTR_SIZE == 4)
+#define WH_DMA_IS_64BIT (WOLFHSM_CFG_DMA_PTR_SIZE == 8)
```
billphipps, line 252: How do we get test coverage of the 32-bit code when the host is 64-bits?  Seems like all of the pointers would be busted.

billphipps (changes requested):
Looks good!  I'd like to reconsider splitting the sha state into a few pointers and ints rather than a pure struct copy.  If you are sure that it wouldn't buy us anything, then I'll approve with a few of the nits resolved.

test/wh_test.c
bigbrett: nope, oversight on my end just to speed up tests, good catch

test/wh_test_crypto.c
bigbrett: good call

wolfhsm/wh_settings.h
bigbrett: I was thinking the same thing. Perhaps just need to add a 32-bit platform to CI, via QEMU or similar. Would be non-trivial to get up and running though

src/wh_client_cryptocb.c
```diff
@@ -1009,6 +1012,155 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
...
+
+#if WH_DMA_IS_32BIT
+    wh_Packet_hash_sha256_Dma32_req* req   = &packet->hashSha256Dma32Req;
+    wh_Packet_hash_sha256_Dma32_res* resp  = &packet->hashSha256Dma32Res;
+#else
+    wh_Packet_hash_sha256_Dma64_req* req   = &packet->hashSha256Dma64Req;
+    wh_Packet_hash_sha256_Dma64_res* resp  = &packet->hashSha256Dma64Res;
+#endif
+
+    /* Caller invoked SHA Update:
```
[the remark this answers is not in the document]
bigbrett: Yep, since client calls wc_Sha256Init() on the context already, and state is held by client, it already has been initialized. CryptoCb doesn't hook the init anyway. Agreed that if state was held by server it would need to be initialized remotely as well, but here the client has already initialized.

src/wh_server_crypto.c
```diff
@@ -1041,16 +1046,179 @@ int wh_Server_HandleCryptoRequest(whServerContext* server,
...
+    int        clientDevId;
+
+    /* Ensure state sizes are the same */
+    if (req->state.sz != sizeof(*sha256)) {
+        res->dmaCryptoRes.badAddr = req->state;
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Copy the SHA256 context from client address space */
+    ret = whServerDma_CopyFromClient(server, sha256, req->state.addr,
```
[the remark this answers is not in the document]
billphipps: Saw that and it will definitely catch the problem, but there's no real indication how to fix it.  Almost want to return a stronger bug than bad args.  Like some sort of internal fault.  Not the user's fault but ours.  Let's keep this moving forward and mark this as needs improvement.  We'll tweak it as we add more DMA's in and get smarter about how to do this.

wolfhsm/wh_settings.h
billphipps: concur.  Maybe if we get more of a fancy DMA mapping working with the shared memory objects, we can force 32-bit offsets into the mapped segments and add/subtract the base.  If the user wants to do DMA, it'll have to be in that shared segment which we can easily detect.

src/wh_client_nvm.c
bigbrett: My thoughts were that this makes it an official API and gets it doxygen documentation. Any reasonable compiler should optimize out the double function call too. But LMK if you would rather it be a macro redirection, I'm happy to change it (and some of the other ones like the NVM DMA read, as well as copy to/from client)

test/wh_test.c
bigbrett: fixed in latest commit

src/wh_client_cryptocb.c
```diff
@@ -1009,6 +1012,155 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
     return ret;
 }
 
+#ifdef WOLFHSM_CFG_DMA
+
+static int _handleSha256Dma(int devId, wc_CryptoInfo* info, void* inCtx,
+                            whPacket* packet)
+{
+    int              ret    = 0;
```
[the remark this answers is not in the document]
bigbrett: fixed in latest commit

src/wh_client_cryptocb.c
```diff
@@ -1009,6 +1012,155 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
...
+                ret = packet->rc;
+            }
+            /* Nothing to do on success, as server will have updated the context
+             * in client memory */
+        }
+    }
+
+    /* Caller invoked SHA finalize:
+     * wc_CryptoCb_Sha256Hash(sha256, NULL, 0, * hash) */
+    if (info->hash.digest != NULL) {
```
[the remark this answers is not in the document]
bigbrett: fixed in latest commit

billphipps (approved):

PR #73, opened 2024-09-11
Update curve25519 and posixshm transport

billphipps:
1. Update curve25519 to match ecc code structure.
2. Update posix shm transport to work on MacOS and add optional DMA mappings
3. Clean up tests
4. Correct other typos

### DIFF
--- a/port/posix/posix_transport_shm.c
+++ b/port/posix/posix_transport_shm.c
@@ -295,6 +295,57 @@ int posixTransportShm_GetDma(posixTransportShmContext* ctx,
     return WH_ERROR_OK;
 }
 
+/* Map the shared object if not already mapped */
+static int posixTransportShm_HandleMap(posixTransportShmContext *ctx)
+{
+    int ret = WH_ERROR_OK;
+    ptshmMapping                map[1]      = { 0 };
+    whTransportMemConfig        tMemCfg[1]  = { 0 };
+
+    if (ctx == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    if (    (ret == WH_ERROR_OK) &&
+            (ctx->state == PTSHM_STATE_NONE) ) {
+        /* Attempt to map */
+        ret = posixTransportShm_UseMap(ctx->name, map);
+        if (ret == WH_ERROR_OK) {
+            /* Configure the underlying transport context */
+            tMemCfg->req_size  = map->header->req_size;
+            tMemCfg->resp_size = map->header->resp_size;
+            tMemCfg->req       = map->req;
+            tMemCfg->resp      = map->resp;
+
+            /* Initialize the shared memory transport */
+            ret = wh_TransportMem_InitClear(ctx->transportMemCtx, tMemCfg, NULL,
+                    NULL);
+
+            if (ret == WH_ERROR_OK) {
+                ctx->ptr = map->ptr;
+                ctx->size = map->size;
+                ctx->dma = map->dma;
+                ctx->dma_size = map->dma_size;
+                ctx->state = PTSHM_STATE_INITIALIZED;
+
+                if (ctx->connectcb != NULL) {
+                    ctx->connectcb(ctx->connectcb_arg, WH_COMM_CONNECTED);
+                }
+            } else {
+                /* Problem initializing the transport */
+                (void)munmap(map->ptr, map->size);
+            }
+        }
+    }
+    if (    (ret == WH_ERROR_OK) &&
+            ((ctx->state == PTSHM_STATE_MAPPED) ||
+             (ctx->state == PTSHM_STATE_DONE) ) ) {
+        /* Mapped is invalid for a client */
+        ret = WH_ERROR_ABORTED;
+    };
+
+    return ret;
+}
 
 /** Callback function definitions */
 int posixTransportShm_ServerInit(void* c, const void* cf,
@@ -321,6 +372,9 @@ int posixTransportShm_ServerInit(void* c, const void* cf,
 
     if (ret == WH_ERROR_OK) {
         memset(ctx, 0, sizeof(*ctx));
+        strncpy(ctx->name,  config->name, sizeof(ctx->name));
+        ctx->connectcb = connectcb;
+        ctx->connectcb_arg = connectcb_arg;
 
         /* Configure the underlying transport context */
         tMemCfg->req_size  = map->header->req_size;
@@ -329,14 +383,16 @@ int posixTransportShm_ServerInit(void* c, const void* cf,
         tMemCfg->resp      = map->resp;
 
         /* Initialize the shared memory transport */
-        ret = wh_TransportMem_Init(ctx->transportMemCtx, tMemCfg, connectcb,
-                                  connectcb_arg);
+        ret = wh_TransportMem_Init(ctx->transportMemCtx, tMemCfg, NULL, NULL);
 
         if (ret == WH_ERROR_OK) {
             ctx->ptr = map->ptr;
             ctx->size = map->size;
             ctx->dma = map->dma;
             ctx->dma_size = map->dma_size;
+
+            ctx->state = PTSHM_STATE_MAPPED;
+
         } else {
             /* Problem initializing the transport */
             (void)munmap(map->ptr, map->size);
@@ -354,8 +410,6 @@ int posixTransportShm_ClientInit(void* c, const void* cf,
     posixTransportShmContext*   ctx         = (posixTransportShmContext*)c;
     posixTransportShmConfig*    config      = (posixTransportShmConfig*)cf;
     int                         ret         = WH_ERROR_OK;
-    ptshmMapping                map[1]      = { 0 };
-    whTransportMemConfig        tMemCfg[1]  = { 0 };
 
     if (    (ctx == NULL) ||
             (config == NULL) ||
@@ -363,6 +417,22 @@ int posixTransportShm_ClientInit(void* c, const void* cf,
         return WH_ERROR_BADARGS;
     }
 
+    memset(ctx, 0, sizeof(*ctx));
+    strncpy(ctx->name,  config->name, sizeof(ctx->name));
+    ctx->connectcb = connectcb;
+    ctx->connectcb_arg = connectcb_arg;
+
+    ret = posixTransportShm_HandleMap(ctx);
+    if (    (ret == WH_ERROR_NOTFOUND) ||
+            (ret == WH_ERROR_NOTREADY) ) {
+        /* Good enough for now.  Set to ok. */
+        ret = WH_ERROR_OK;
+    }
+
+    return ret;
+#if 0
+    ptshmMapping                map[1]      = { 0 };
+    whTransportMemConfig        tMemCfg[1]  = { 0 };
     /* Ignore the config sizes and read from the shared file */
     do {
         ret = posixTransportShm_UseMap( config->name, map);
@@ -372,10 +442,11 @@ int posixTransportShm_ClientInit(void* c, const void* cf,
         } else {
             break;
         }
-    } while (1);
+    } while (retries-- > 0);
 
     if (ret == WH_ERROR_OK) {
         memset(ctx, 0, sizeof(*ctx));
+        strncpy(ctx->name,  config->name, sizeof(ctx->name));
 
         /* Configure the underlying transport context */
         tMemCfg->req_size  = map->header->req_size;
@@ -398,6 +469,7 @@ int posixTransportShm_ClientInit(void* c, const void* cf,
         }
     }
     return ret;
+#endif
 }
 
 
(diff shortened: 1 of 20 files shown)


port/posix/posix_transport_shm.c
```diff
@@ -1,301 +1,581 @@
...
+        struct stat st[1] = { 0 };
+        ret = fstat(fd, st);
+        if (ret == 0) {
+            if (st->st_size != 0) {
+                /* Map the header and get configuration */
+                ptshmHeader* header = (ptshmHeader*)mmap(NULL, sizeof(*header),
+                        PROT_READ, MAP_SHARED, fd, 0);
+                if (header != MAP_FAILED) {
+                    size_t size = 0;
+                    if(header->initialized > PTSHM_INITIALIZED_NONE) {
```
bigbrett, line 113: perhaps an explicit check for `if (header->initialized == PTSHM_INITIALIZED_CREATOR)`, since I think all other states should be an invalid sequence, right? 

Also, there is a missing space after `if`

port/posix/posix_transport_shm.c
```diff
@@ -1,301 +1,581 @@
...
 
     /* Only need to check NULL, mem transport checks other state info */
     if (ctx == NULL) {
         return WH_ERROR_BADARGS;
     }
 
-    return wh_TransportMem_SendRequest(ctx->transportMemCtx, len, data);
+    /* Check connected status */
+    switch(ctx->state) {
+    case PTSHM_STATE_NONE:
```
bigbrett, line 501: A bit confused about this logic (as well as the client initialization. Since the client init is what initially zeros the context and calls `posixTransportShm_{Use/Handle}Map()` on it, and we don't ever want a user to be sending requests without first initializing the client context, should we even be handling the `NONE` state here? I'd have figured that this logic should have been in the `INITIALIZED` state, since that is what the client init handler sets the state to once it successfully initializes. Is this some sort of deferred initialization that is allowed if the original client init fails? If so perhaps we just call client init again here?

port/posix/posix_transport_shm.c
```diff
@@ -1,301 +1,581 @@
...
+    if (ctx == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+    header = (ptshmHeader*)ctx->ptr;
+    if (header == NULL) {
+        return WH_ERROR_NOTREADY;
     }
+    if (header->initialized > PTSHM_INITIALIZED_CREATOR) {
+        connected = WH_COMM_CONNECTED;
+    }
```
bigbrett, line 229: Should this be checking against `PTSHM_INITIALIZED_USER` here, since the client sets `header->initialized = PTSHM_INITIALIZED_USER` only once it has mapped and cleared the shared buffer (which is what I believe what we are indicating as ready to connect)?

port/posix/posix_transport_shm.c
```diff
@@ -1,301 +1,581 @@
...
-    /* Initialize or increment the reference count */
-    atomic_fetch_add(&header->refCount, 1);
 
-    /* allocate a shared memory transport context */
-    ctx->transportMemCtx = malloc(sizeof(whTransportMemContext));
-    if (ctx->transportMemCtx == NULL) {
-        return WH_ERROR_ABORTED;
+/** Custom functions */
+int posixTransportShm_IsConnected(posixTransportShmContext* ctx,
+        whCommConnected *out_connected)
```
bigbrett, line 216: function should probably be static

port/posix/posix_transport_shm.h
```diff
@@ -20,6 +20,31 @@
...
+ * Once the server creates the named shared memory object, it sets the full
+ * size, maps it, and updates the header block with the configured sizes. After
+ * completing updates to the header block, it then builds a TransportMem
+ * server context using the request and response buffers.  Note the server will
+ * attempt to unlink the named shared memory object prior to creation.
+ *
+ * The client is configured with only the name of the shared object and it busy-
+ * retries to open the named shared object.  It then maps the header block,
+ * reads the configuration sizes, and configures a TransportMem client context.
+ * The client also unlinks the shared object on successful mapping.
```
bigbrett, line 39: This behavior is only present in the `#if 0` code in the client init. See below comments for more

port/posix/posix_transport_shm.h
```diff
@@ -29,31 +54,57 @@
...
 } posixTransportShmConfig;
 
 
-/** Common context */
+/** POSIX SHM context */
+typedef enum {
+    PTSHM_STATE_NONE = 0,       /* Not opened, not mapped */
+    PTSHM_STATE_MAPPED,         /* Opened and mapped, no client */
+    PTSHM_STATE_INITIALIZED,    /* Fully initialized */
+    PTSHM_STATE_DONE,           /* Closed and complete */
```
bigbrett, line 76: This state is never set by client or server

bigbrett (changes requested):
Partial review: Haven't yet looked at the crypto or TCP portions yet. But wanted some clarifications of how things are supposed to work re: client initialization

test/wh_test_crypto.c
```diff
@@ -87,16 +88,254 @@ static int _cancelCb(uint16_t seq)
...
+    uint8_t hash[TEST_ECC_KEYSIZE] = {0};
+    uint8_t sig[ECC_MAX_SIG_SIZE] = {0};
+
+#if 0
+    whNvmFlags flags = WH_NVM_FLAGS_NONE;
+    whKeyId key_id_a = WH_KEYID_ERASED;
+    uint8_t* label_a = (uint8_t*)("Ecc Label A");
+    whKeyId key_id_b = 24;
+    uint8_t* label_b = (uint8_t*)("Ecc Label B");
+#endif
```
bigbrett, line 110: dead code

test/wh_test_crypto.c
```diff
@@ -706,7 +949,16 @@ int whTest_CryptoClientConfig(whClientConfig* config)
         goto exit;
     }
 #endif /* HAVE_ECC */
+#endif
```
bigbrett, line 952: dead code. Imagine you just want to remove the old code and instead call the new function

test/wh_test_crypto.c
```diff
@@ -746,9 +998,12 @@ int whTest_CryptoClientConfig(whClientConfig* config)
     }
     printf("CURVE25519 SUCCESS\n");
 #endif /* HAVE_CURVE25519 */
+#endif
```
bigbrett, line 1001: dead code. Imagine you just want to remove the old code and instead call the new function

wolfhsm/wh_client_crypto.h
```diff
@@ -115,27 +115,25 @@ int wh_Client_ImportCurve25519Key(whClientContext* ctx, curve25519_key* key,
...
- * struct accordingly */
-int wh_Client_MakeCurve25519Key(whClientContext* ctx,
-        uint32_t size,
-        whKeyId *inout_key_id, whNvmFlags flags,
-        uint32_t label_len, uint8_t* label,
-        curve25519_key* key);
+int wh_Client_Curve25519ExportKey(whClientContext* ctx, whKeyId keyId,
+        curve25519_key* key, uint16_t label_len, uint8_t* label);
 
 /* TODO: Generate a Curve25519 key on the server and put it in the server keycache */
```
bigbrett, line 121: all these functions are no longer a TODO, and also need doxygen API docs

src/wh_client_crypto.c
```diff
@@ -705,4 +716,360 @@ printf("[client] %s ctx:%p key:%p, sig:%p sig_len:%u, hash:%p hash_len:%u out_re
...
+        if (packet->rc == 0) {
+            /* Key is cached on server or is ephemeral */
+            key_id = (whKeyId)(res->keyId);
+
+            /* Update output variable if requested */
+            if (inout_key_id != NULL) {
+                *inout_key_id = key_id;
+            }
+
+            /* Update the RSA context if provided */
```
bigbrett, line 879:
```suggestion
            /* Update the context if provided */
```

src/wh_crypto.c
```diff
@@ -147,4 +148,51 @@ int wh_Crypto_UpdatePrivateOnlyEccKey(ecc_key* key, uint16_t pub_size,
...
+}
+
+int wh_Crypto_DeserializeCurve25519Key(uint16_t size,
+        const uint8_t* buffer, curve25519_key* key)
+{
+    int ret = 0;
+    word32 privSz = CURVE25519_KEYSIZE;
+    word32 pubSz = CURVE25519_KEYSIZE;
+
+    if (    (size < (CURVE25519_KEYSIZE*2)) ||
```
bigbrett, line 181:
```suggestion
    if (    (size < (CURVE25519_KEYSIZE * 2)) ||
```

src/wh_crypto.c
```diff
@@ -147,4 +148,51 @@ int wh_Crypto_UpdatePrivateOnlyEccKey(ecc_key* key, uint16_t pub_size,
 }
 #endif /* HAVE_ECC */
 
+#ifdef HAVE_CURVE25519
+ int wh_Crypto_SerializeCurve25519Key(curve25519_key* key,
+        uint16_t max_size, uint8_t* buffer, uint16_t *out_size)
```
bigbrett, line 153: do we need to check that the `curve25519_key` is a true keypair and not only public or private part? Also, do we want to support private/public only keys?

bigbrett (changes requested):
overall looks good

bigbrett:
Verified crypto tests pass on tc3xx

port/posix/posix_transport_shm.c
```diff
@@ -1,301 +1,581 @@
...
+/* Memory map and interpret the header block */
+static int posixTransportShm_Map(int fd, size_t size, ptshmMapping* map);
+
+/* Use and map a shared object for transport */
+static int posixTransportShm_UseMap(char* name, ptshmMapping* map);
+
+/* Create and map a shared object for transport */
+static int posixTransportShm_CreateMap(char* name, uint16_t req_size,
+        uint16_t resp_size, size_t dma_size, ptshmMapping* map);
+
```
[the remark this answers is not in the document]
billphipps: Will do.

port/posix/posix_transport_shm.c
billphipps: corrected

port/posix/posix_transport_shm.c
```diff
@@ -1,301 +1,581 @@
...
+                header->req_size = req_size;
+                header->resp_size = resp_size;
+                header->dma_size = dma_size;
+                header->creator_pid = getpid();
+                XMEMFENCE();
+                header->initialized = PTSHM_INITIALIZED_CREATOR;
+
+                /* Unmap the header and remap the full area */
+                (void)munmap((void*)header, sizeof(*header));
+                ret = posixTransportShm_Map(fd, size, map);
```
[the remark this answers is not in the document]
billphipps: The client doesn't know the sizes of the buffers, so it maps the header first to find out, then remaps once the sizes are understood.

port/posix/posix_transport_shm.c
billphipps: This is exposed to the user so they will know when the DMA region is usable by both sides.  Really, the server (creator) needs to understand this since the client (user) will be the last to map it.  The transport handles the connected state internally.

port/posix/posix_transport_shm.c
billphipps: Updated.  In this implementation, the server creates the object, which the OS dutifully zeroes for us.  The client merely maps, updates the initialized member, and starts sending requests.

port/posix/posix_transport_shm.c
```diff
@@ -1,301 +1,581 @@
...
+            ctx->size = map->size;
+            ctx->dma = map->dma;
+            ctx->dma_size = map->dma_size;
+        } else {
+            /* Problem initializing the transport */
+            (void)munmap(map->ptr, map->size);
+        }
+    }
+    return ret;
+#endif
```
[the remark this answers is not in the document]
billphipps: The comments in the header were the first way i tried to improve the connect sequence, and they are wrong.  I'll fix those too.  This code is removed.

port/posix/posix_transport_shm.c
billphipps: The server creates and zeroes the object.  The client only has to map it.  NONE is used when the client was not able to open the object OR the object was not completely setup by the server.  This forces the client to attempt to reopen, map the header, and then remap the entire object. Once the client is able to successfully map the entire segment, it sets the intiitalzed member in the header and leaves the mapping intact.  Recall, the client init only tries once to open and map the object, so the entire sequence is tried again here as well.  Sorry for the misleading comments.

port/posix/posix_transport_shm.h
billphipps: Yup.  This is incorrect.  Updated.

port/posix/posix_transport_shm.h
billphipps: Updated.  Now set during cleanup.

port/posix/posix_transport_tcp.c
```diff
@@ -233,10 +240,157 @@ static int posixTransportTcp_Recv(int fd, uint16_t* buffer_offset,
...
+                    (void)posixTransportTcp_Close(c);
+                    c->state = PTT_STATE_DONE;
+                    ret = WH_ERROR_ABORTED;
+                } else {
+                    /* Likely connected */
+                    if ((pfd.revents & POLLOUT) != 0) {
+                        /* Connected */
+                        c->state = PTT_STATE_CONNECTED;
+                        ret = WH_ERROR_OK;
+                    } else {
```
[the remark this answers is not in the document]
billphipps: Turns out that errors on the socket will set POLLERR, which is checked first.  In some instances, POLLERR and POLLOUT are mutually exclusive (like MacOS), but Linux likes to set multiple flags.  Either way, the socket error doesn't need to be tested independently of these calls.  I did have a version of the code that checked that as well, but that turned out to be superfluous.

port/posix/posix_transport_tcp.h
```diff
@@ -20,6 +20,17 @@
  * port/posix/posix_transport_tcp.h
  *
  * wolfHSM Transport binding using TCP sockets
+ *
+ * Note in this implementation, all function calls are non-blocking and will
+ * return WH_ERROR_NOTREADY when the operation is pending or cannot be started
+ * yet.  Async client connection will initially attempt to connect and tear
```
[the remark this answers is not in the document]
billphipps: fixed

src/wh_crypto.c
billphipps: Hmm.  The wc code handles either/both by using separate parts. We always assume both are present but support either being absent since the buffer is fixed side.  I think an all 0 key is not valid so it would be handled as such, but you do raise a good question.  I'll look into it.

test/Makefile
```diff
@@ -4,7 +4,7 @@ CMD_ECHO =
 # Important directories
 BUILD_DIR = ./Build
 WOLFHSM_DIR = $(CURDIR)/../
-WOLFSSL_DIR ?= $(CURDIR)/../../wolfssl
+WOLFSSL_DIR = $(CURDIR)/../../wolfssl
```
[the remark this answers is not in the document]
billphipps: Yup.  My bad.  I was resolving a compile issue where the include path was getting munged since wh_settings.h is now able to include wc_port.h (for XFENCE).  I'll revert it as this was not the problem.

test/Makefile
```diff
@@ -175,7 +175,8 @@ SRC_C += \
             ./src/wh_test_she.c \
             ./src/wh_test_nvm_flash.c \
             ./src/wh_test_clientserver.c \
-            ./src/wh_test_flash_ramsim.c
+            ./src/wh_test_flash_ramsim.c \
+            ./src/wh_test_check_struct_padding.c
```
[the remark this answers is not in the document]
billphipps: I was actually happy to have this in the main build when I was hacking up packet structures.  Noted.

test/wh_test_comm.c
```diff
@@ -35,11 +34,13 @@
 
 #if defined(WOLFHSM_CFG_TEST_POSIX)
 #include <pthread.h> /* For pthread_create/cancel/join/_t */
-#include <unistd.h>  /* For sleep */
+#include <unistd.h>  /* For usleep */
```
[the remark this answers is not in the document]
billphipps: Oops.  the sleep was added to the server to simplify testing of late servers.  Removed.

test/wh_test_crypto.c
billphipps: I'll add in the additional key testing styles, just like curve25519.  This won't be dead then.

test/wh_test_crypto.c
billphipps: yup. removed.

test/wh_test_crypto.c
billphipps: yup.

wolfhsm/wh_client_crypto.h
billphipps: The TODO is for me to add the doxygen!  :)

test/wh_test_comm.c
```diff
@@ -272,7 +272,6 @@ static void* _whCommServerTask(void* cf)
     whCommServer        server[1];
     int                 counter = 1;
 
-    sleep(0);
```
bigbrett: should be able to pull the `unistd.h` include now in your next RSA+AES PR

test/wh_test_crypto.c
```diff
@@ -26,6 +26,7 @@
 #include <stdint.h>
 #include <stdio.h>  /* For printf */
 #include <string.h> /* For memset, memcpy */
+#include <unistd.h> /* For sleep */
```
bigbrett, line 29: needs to be protected by `WOLFHSM_CFG_TEST_POSIX` (block already present later in this file)

bigbrett (changes requested):
just need to guard `unistd.h` include with `WOLFHSM_CFG_TEST_POSIX` and I can merge

bigbrett (approved):

PR #74, opened 2024-09-25
Update aes, rsa, keystore, and finish wolfcrypt test

billphipps:
1. Fixes RNG, AES, RSA to pass wolfcrypt test
2. Updates test to eliminate goto and provide separate functions per algo
3. Updated devids[] to not be replicated per translation unit
4. Renames, reorgs.  Still more pending.

WIP

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -66,7 +66,7 @@
 /** Forward declarations */
 #ifdef HAVE_ECC
 /* Server creates a key based on incoming flags */
-static int _wh_Client_EccMakeKey(whClientContext* ctx,
+static int _EccMakeKey(whClientContext* ctx,
         int size, int curveId,
         whKeyId *inout_key_id, whNvmFlags flags,
         uint16_t label_len, uint8_t* label,
@@ -74,7 +74,8 @@ static int _wh_Client_EccMakeKey(whClientContext* ctx,
 #endif /* HAVE_ECC */
 
 #ifdef HAVE_CURVE25519
-static int _wh_Client_Curve25519MakeKey(whClientContext* ctx,
+/* Server creates a key based on incoming flags */
+static int _Curve25519MakeKey(whClientContext* ctx,
         uint16_t size,
         whKeyId *inout_key_id, whNvmFlags flags,
         uint16_t label_len, uint8_t* label,
@@ -83,7 +84,7 @@ static int _wh_Client_Curve25519MakeKey(whClientContext* ctx,
 
 #ifndef NO_RSA
 /* Make an RSA key on the server based on the flags */
-static int _wh_Client_RsaMakeKey(whClientContext* ctx,
+static int _RsaMakeKey(whClientContext* ctx,
         uint32_t size, uint32_t e,
         whNvmFlags flags,  uint32_t label_len, uint8_t* label,
         whKeyId *inout_key_id, RsaKey* rsa);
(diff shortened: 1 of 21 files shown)


src/wh_client_cryptocb.c
```diff
@@ -643,103 +709,113 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
+
         /* write request */
-        ret = wh_Client_SendRequest(ctx, group, WC_ALGO_TYPE_CMAC,
-            WH_PACKET_STUB_SIZE + sizeof(packet->cmacReq) +
-            packet->cmacReq.inSz + packet->cmacReq.keySz, (uint8_t*)packet);
+        ret = wh_Client_SendRequest(ctx, group, action, req_len,
+                (uint8_t*)packet);
         if (ret == 0) {
+            /* Update the local type since call succeeded */
+            cmac->type = type;
```
bigbrett, line 789: this looks to be already set earlier in a success case

src/wh_client_cryptocb.c
```diff
@@ -643,103 +709,113 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
             /* if the client marked they may want to cancel, handle the
-             * response in a seperate call */
+             * response req_in a seperate call */
```
bigbrett, line 791:
```suggestion
             * response req_in a separate call */
```

Also, `req_in` is the request input data, what were you trying for with this edit?

src/wh_client_crypto.c
```diff
@@ -1069,7 +1138,340 @@ int wh_Client_Curve25519SharedSecret(whClientContext* ctx,
...
+    uint8_t keyLabel[WH_NVM_LABEL_LEN] = {0};
+
+    if (    (ctx == NULL) ||
+            (keyId == WH_KEYID_ERASED) ||
+            (key == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Now export the key from the server */
+    ret = wh_Client_KeyExport(ctx,keyId,
```
bigbrett, line 1206:
```suggestion
    ret = wh_Client_KeyExport(ctx, keyId,
```

src/wh_client_crypto.c
```diff
@@ -1069,7 +1138,340 @@ int wh_Client_Curve25519SharedSecret(whClientContext* ctx,
...
+    int derSize = 0;
+
+    if (    (ctx == NULL) ||
+            (key == NULL) ||
+            ((label_len != 0) && (label == NULL))) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Convert RSA key to DER format */
+    ret = derSize = wc_RsaKeyToDer(key, keyDer, sizeof(keyDer));
```
bigbrett, line 1177: I don't believe this handles public keys - would we want a separate client function for public key import/export? Or we could just check `key->type` and call `wc_RsaKeyToPublicDer` if it is public? The only reason it is significant is it dictates which wolfCrypt API we use in order to generate the DER, as `wc_RsaKeyToDer` will fail if it is a public key.

src/wh_client_crypto.c
```diff
@@ -1069,7 +1138,340 @@ int wh_Client_Curve25519SharedSecret(whClientContext* ctx,
...
+    }
+
+    /* Now export the key from the server */
+    ret = wh_Client_KeyExport(ctx,keyId,
+            keyLabel, sizeof(keyLabel),
+            keyDer, &derSize);
+    if (ret == 0) {
+        word32 idx = 0;
+        /* Update the RSA key structure */
+        ret = wc_RsaPrivateKeyDecode(
```
bigbrett, line 1212: same as above - this only handles private keys.

src/wh_client_crypto.c
```diff
@@ -1069,7 +1138,340 @@ int wh_Client_Curve25519SharedSecret(whClientContext* ctx,
...
+            if (rsa != NULL) {
+                word32 der_size = (word32)(res->len);
+                uint8_t* rsa_der = (uint8_t*)(res + 1);
+                word32 idx = 0;
+                /* Set the rsa key_id.  Should be ERASED if EPHEMERAL */
+                wh_Client_SetKeyIdRsa(rsa, key_id);
+
+                if (flags & WH_NVM_FLAGS_EPHEMERAL) {
+                    /* Response has the exported key */
+                    ret = wc_RsaPrivateKeyDecode(rsa_der, &idx, rsa, der_size);
```
bigbrett, line 1315: same as above - this only handles private keys

src/wh_crypto.c
```diff
@@ -44,6 +45,49 @@
...
+{
+    int ret = 0;
+    int der_size = 0;
+
+    if (    (key == NULL) ||
+            (buffer == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    der_size = wc_RsaKeyToDer(key, (byte*)buffer, (word32)max_size);
```
bigbrett, line 60: As mentioned above (or below maybe....) this assumes `key` is a private key. Consider switching on `key->type` and calling  `wc_RsaKeyToPublicDer()` if a public key

src/wh_crypto.c
```diff
@@ -44,6 +45,49 @@
...
+    int ret;
+    word32 idx = 0;
+
+    if (    (size == 0) ||
+            (buffer == NULL) ||
+            (key == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+    /* Deserialize the RSA key */
+    ret = wc_RsaPrivateKeyDecode(buffer, &idx, key, size);
```
bigbrett, line 86: this only handles private keys - a trickier problem to determine if the DER is public or private. Not sure we can do that in one pass if no metadata is present indicating whether public or private. Perhaps we try public first, which I think checks the ASN.1 OID, and if that fails try private?

src/wh_server_crypto.c
```diff
@@ -79,6 +103,244 @@ static int wh_Server_HandleSharedSecretCurve25519(whServerContext* ctx,
...
+    if (ret == 0) {
+        ret = wh_Crypto_SerializeRsaKey(key, max_size, cacheBuf, &der_size);
+    }
+
+    if (ret == 0) {
+        /* set meta */
+        cacheMeta->id = keyId;
+        cacheMeta->len = der_size;
+        cacheMeta->flags = flags;
+        cacheMeta->access = WH_NVM_ACCESS_ANY;
```
bigbrett, line 143: eventually we want `access` to be passed as an arg to all the import functions, right?

src/wh_server_crypto.c
```diff
@@ -79,6 +103,244 @@ static int wh_Server_HandleSharedSecretCurve25519(whServerContext* ctx,
...
+    /* load the key from the keystore */
+    if (ret == 0) {
+        ret = wh_Server_CacheExportRsaKey(server, key_id, rsa);
+#ifdef DEBUG_CRYPTOCB_VERBOSE
+        printf("[server] CacheExportRsaKey keyid:%u, ret:%d\n", key_id, ret);
+#endif
+        if (ret == 0) {
+            /* do the rsa operation */
+            ret = wc_RsaFunction(in, in_len, out, &out_len,
+                op_type, rsa, server->crypto->rng);
```
bigbrett, line 291: hmm I know this was existing code, but do all the implementations of RSA support in-place operations? SP, fast math, etc? A quick skim tells me yes, and docs don't mention, but perhaps worth a double check. Here the input and output pointers overlap, since it is overlaid on the same packet. Clearly not causing issues since we are passing wolfCrypt tests, but hopefully not a fluke.

bigbrett (commented):
Awesome stuff Bill, can't wait to see the rest.

 Flagged a few things to consider while finishing this up, most critically the handling of public RSA keys.

Also, if wolfcrypt tests are now passing, please remove the `|| true` from the end[ of this line in `build-and-test.yml`](https://github.com/wolfSSL/wolfHSM/blob/main/.github/workflows/build-and-test.yml#L39) so CI for this PR will actually run wolfCrypt test.

wolfhsm/wh_packet.h
```diff
@@ -57,14 +57,15 @@ typedef struct  wh_Packet_cipher_aescbc_req
...
+    uint8_t WH_PAD[2];
+    /* in[sz] | key[keyLen] | iv[AES_IV_SIZE] */
 } wh_Packet_cipher_aescbc_req;
 
 typedef struct  wh_Packet_cipher_aescbc_res
 {
     uint32_t sz;
-    uint8_t WH_PAD[4];
+    /* Pad to ensure req and res overlap on in/out */
+    uint8_t WH_PAD[sizeof(wh_Packet_cipher_aescbc_req) - sizeof(uint32_t)];
```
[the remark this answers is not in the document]
billphipps: In the server, the req and res packets point to the same spot in ram.  AES (ECB/GCM/CBC) software implementations have special handling for when the input and output buffers are the same, but fail when the overlap.  This change ensures the in/out buffers are in the same spot and handle this case correctly.  Funny, it's not an issue if you offload the AES to hardware on the server.

wolfhsm/wh_crypto.h
```diff
@@ -36,9 +36,27 @@
...
+        Aes* key);
+#endif /* !NO_AES */
+
+#ifndef NO_RSA
+/* Store a RsaKey to a byte sequence (currently DER format) */
+int wh_Crypto_SerializeRsaKey(RsaKey* key, uint16_t max_size,
+        uint8_t* buffer, uint16_t *out_size);
+/* Restore a RsaKey from a byte sequence (currently DER format) */
+int wh_Crypto_DeserializeRsaKey(uint16_t size, const uint8_t* buffer,
+        RsaKey* key);
```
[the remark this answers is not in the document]
billphipps: Done already.  :)

test/wh_test_crypto.c
```diff
@@ -82,9 +82,210 @@ static int _cancelCb(uint16_t seq)
...
+#define RSA_EXPONENT 65537
+
+    int ret = WH_ERROR_OK;
+    RsaKey rsa[1];
+    char plainText[sizeof(PLAINTEXT)] = PLAINTEXT;
+    char cipherText[RSA_KEY_BYTES];
+    char finalText[RSA_KEY_BYTES];
+    whKeyId keyId = WH_KEYID_ERASED;
+
+    /* Using ephemral key */
```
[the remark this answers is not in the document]
billphipps: updated

test/wh_test_crypto.c
```diff
@@ -82,9 +82,210 @@ static int _cancelCb(uint16_t seq)
...
+    }
+    return ret;
+}
+
+#ifndef NO_RSA
+static int whTest_CryptoRsa(whClientContext* ctx, int devId, WC_RNG* rng)
+{
+#define RSA_KEY_BITS 2048
+#define RSA_KEY_BYTES (RSA_KEY_BITS/8)
+#define RSA_EXPONENT 65537
```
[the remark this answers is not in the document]
billphipps: done.  good catch.  didn't even know that was there.

test/wh_test_crypto.c
```diff
@@ -182,501 +385,724 @@ static int whTest_CryptoCurve25519(whClientContext* ctx, WC_RNG* rng)
...
-            if (ret == 0) {
-                if (XMEMCMP(shared_ab, shared_ba, len) != 0) {
-                    WH_ERROR_PRINT("CURVE25519 shared secrets don't match\n");
-                    ret = -1;
+                if (ret == 0) {
+                    len = sizeof(shared_ba);
+                    ret = wc_curve25519_shared_secret(
+                            key_b, key_a, shared_ba, &len);
+                    if (ret != 0) {
+                        WH_ERROR_PRINT("Failed to compute shared secrete %d\n",
```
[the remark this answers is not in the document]
billphipps: updated

test/wh_test_crypto.c
```diff
@@ -182,501 +385,724 @@ static int whTest_CryptoCurve25519(whClientContext* ctx, WC_RNG* rng)
...
-    }
-    outLen = sizeof(keyEnd);
-    if ((ret = wh_Client_KeyExport(client, keyId, labelEnd, sizeof(labelEnd), keyEnd, &outLen)) != 0) {
-        WH_ERROR_PRINT("Failed to wh_Client_KeyExport %d\n", ret);
-        goto exit;
-    }
-    if (ret == 0 && XMEMCMP(key, keyEnd, outLen) == 0 && XMEMCMP(labelStart, labelEnd, sizeof(labelStart)) == 0)
+    ret = whTest_CacheExportKey(ctx, &keyId,
+            labelIn, labelOut, sizeof(labelIn),
+            key, keyOut, sizeof(key));
```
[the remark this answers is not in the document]
billphipps: Thank your for finding that.  I moved this around and didn't repeat the random key.  Updated.

wolfhsm/wh_packet.h
```diff
@@ -57,14 +57,15 @@ typedef struct  wh_Packet_cipher_aescbc_req
     uint32_t keyLen;
     uint32_t sz;
     uint16_t keyId;
-    uint8_t WH_PAD[6];
-    /* key[keyLen] | iv[AES_IV_SIZE] | in[sz] */
+    uint8_t WH_PAD[2];
+    /* in[sz] | key[keyLen] | iv[AES_IV_SIZE] */
```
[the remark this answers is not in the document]
billphipps: This ensures the req->in lines up with res->out.

src/wh_client_crypto.c
```diff
@@ -1069,7 +1138,340 @@ int wh_Client_Curve25519SharedSecret(whClientContext* ctx,
...
+
+    req = &packet->pkRsakgReq;
+    res = &packet->pkRsakgRes;
+
+    /* Use the supplied key id if provided */
+    if (inout_key_id != NULL) {
+        key_id = *inout_key_id;
+    }
+
+    /* Set up the reqeust packet */
```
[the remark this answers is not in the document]
billphipps: updated

src/wh_client_cryptocb.c
billphipps: for the "init" case (null or 0 input args), it does set the type but then breaks out of the case statement.  I should eliminate the break so that it is more clear.

src/wh_client_cryptocb.c
billphipps: Used refactor function in eclipse editor.  It found some comments when i was renaming "in" to "req_in". ill fix.

src/wh_client_crypto.c
billphipps: I saw this today and it should be changed to use the Serialize helper in wh_crypto.h that correctly calls the right export function based on the kind of rsa key.  I'll update and verify the functionality.  Probably need a few new test case to make sure all 3 cases work.

src/wh_client_crypto.c
billphipps: Yup.  This should be updated to use deserialize.

src/wh_server_crypto.c
billphipps: yup.  and we need to return more of the keyid to the client.  right now, we ONLY return the ID portion, stripping off the higher bits.

src/wh_server_crypto.c
billphipps: Definitely supported.  Remember that all of the RSA functions use the large integer math libraries and simply export the answer from that at the end of the RSA function.

src/wh_server_crypto.c
```diff
@@ -251,6 +251,9 @@ static int wh_Server_HandleRsaFunction(whServerContext* server, whPacket* packet
     RsaKey rsa[1] = {0};
 
     int op_type     = (int)(packet->pkRsaReq.opType);
+    uint32_t options = packet->pkRsaReq.options;
+    int evict       = options & WH_PACKET_PK_ECCSIGN_OPTIONS_EVICT;
```
bigbrett, line 255: nano-nit: I like `!!` to be explicit that this is turning it into a boolean flag

src/wh_client_crypto.c
```diff
@@ -156,6 +157,284 @@ int wh_Client_RngGenerate(whClientContext* ctx, uint8_t* out, uint32_t size)
     return ret;
 }
 
+#ifndef NO_AES
+
+#ifdef HAVE_AES_CBC
+int wh_Client_AesCbc(whClientContext* ctx,
+        Aes* aes, int enc,
```
bigbrett, line 164: missing NULL check on `Aes* aes`. Pretty sure you only want this function called by the cryptoCb, in which case you know that the calling wolfCrypt AES API does the NULL check, but since it is exposed in a header, we should check it anyway in case we were to want to use it elsewhere

src/wh_client_crypto.c
```diff
@@ -156,6 +157,284 @@ int wh_Client_RngGenerate(whClientContext* ctx, uint8_t* out, uint32_t size)
...
+#ifdef DEBUG_CRYPTOCB_VERBOSE
+    printf("[client] %s: enc:%d keylen:%d ivsz:%d insz:%d reqsz:%u ressz:%u "
+                      "blocks:%u lastoffset:%u\n",
+                __func__,enc, key_len, iv_len, len,  req_len, res_len,
+                blocks, last_offset);
+    printf("[client] %s: req:%p in:%p key:%p iv:%p res:%p out:%p\n",
+            __func__, req, req_in, req_key, req_iv, res, res_out);
+#endif
+
+    /* III 0 size check is done in wolfCrypt */
```
bigbrett, line 211: Just a thought addressing the TODO - I'd say keep the check in, I'd rather do redundant checking of args with a microscopic perf penalty, but keep functions safe for use outside the cryptoCallback.

src/wh_client_crypto.c
```diff
@@ -156,6 +157,284 @@ int wh_Client_RngGenerate(whClientContext* ctx, uint8_t* out, uint32_t size)
...
+        /* CBC requires only full blocks */
+        return BAD_LENGTH_E;
+
+    }
+
+    if (    (req_len > WOLFHSM_CFG_COMM_DATA_LEN) ||
+            (res_len > WOLFHSM_CFG_COMM_DATA_LEN)) {
+        /* if we're using an HSM key return BAD_FUNC_ARG */
+        if (WH_KEYID_ISERASED(key_id)) {
+            return CRYPTOCB_UNAVAILABLE;
```
bigbrett, line 223: Are we sure we want  to silently fall back to local crypto here? I guess I'm trying to figure out what would get you here....is this triggered when you want to do HSM crypto, have an input larger than the comm MTU, but but didn't call the Client SetKeyID API? 

Perhaps this should be `BAD_FUNC_ARG` too then? I feel like if they supply the HSM devID, then they should expect the operation to be RPC'ed to the HSM, not secretly done locally with no way of checking. And if it is too big to send across the transport, it should just be an error right?

That said, I think the key would almost always be NULL in this case since they most likely didn't call setKey if using an HSM key, so believe the regular AES logic would check it. Though if for some reason they called AesSetKey I think this would end up doing the operation locally without any error reported to the user, which I don't think we should be doing?

src/wh_client_crypto.c
```diff
@@ -156,6 +157,284 @@ int wh_Client_RngGenerate(whClientContext* ctx, uint8_t* out, uint32_t size)
...
+
+        }
+    }
+    return ret;
+}
+#endif /* HAVE_AES_CBC */
+
+#ifdef HAVE_AESGCM
+int wh_Client_AesGcm(whClientContext* ctx,
+        Aes* aes, int enc,
```
bigbrett, line 302: same as above re: `Aes* aes` NULL check

src/wh_client_crypto.c
```diff
@@ -156,6 +157,284 @@ int wh_Client_RngGenerate(whClientContext* ctx, uint8_t* out, uint32_t size)
...
+                enc, key_len, iv_len, len, authin_len, tag_len,
+                req_len, res_len);
+    printf("[client] AESGCM: req:%p in:%p key:%p iv:%p authin:%p dec_tag:%p res:%p out:%p res_enc_tag:%p\n",
+            req, req_in, req_key, req_iv, req_authin, req_dec_tag, res, res_out, res_enc_tag);
+#endif
+    if (    (req_len > WOLFHSM_CFG_COMM_DATA_LEN) ||
+            (res_len > WOLFHSM_CFG_COMM_DATA_LEN)) {
+        /* if we're using an HSM key return BAD_FUNC_ARG */
+        if (WH_KEYID_ISERASED(key_id)) {
+            return CRYPTOCB_UNAVAILABLE;
```
bigbrett, line 352: same comment as above re: considering making this an error

src/wh_client_crypto.c
```diff
@@ -791,6 +1070,58 @@ printf("[client] %s ctx:%p key:%p, sig:%p sig_len:%u, hash:%p hash_len:%u out_re
     return ret;
 }
 
+#if 0
```
bigbrett, line 1073: are we going to leave this in for the release or is it pending?

bigbrett (changes requested):

src/wh_server_crypto.c
billphipps: updated.

src/wh_client_crypto.c
billphipps: updated and added to gem as well.  changed argument order to match wc.

src/wh_client_crypto.c
billphipps: Updated to handle locally.

src/wh_client_crypto.c
billphipps: If there is a key_id set, then you can't fallback and currently this function can't handle inputs/outputs that are larger than the MTU.  If there is no keyid set, then the key is local and we can fall back to local AES, if available.  This check is if the keyid is erased (not set) then assume local fallback is ok, otherwise the hsm client function was incorrectly called (badargs), which is a calling/coding error.

src/wh_client_crypto.c
billphipps: updated.

src/wh_client_crypto.c
billphipps: Leave it for now since I think it is working, but there's now way to directly exercise it until I update wc.

src/wh_client_cryptocb.c
```diff
@@ -430,24 +430,60 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
+            ret = wh_Client_RsaFunction(ctx,
+                    rsa, rsa_type, in, in_len,
+                    out, &len);
+
+            if (    (ret == WH_ERROR_OK) &&
+                    (out_len != NULL) ) {
+                *out_len = len;
+            }
+
+#if 0
```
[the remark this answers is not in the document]
billphipps: updated

src/wh_client_crypto.c
bigbrett: yep exactly, that was what I was wondering - it is the "you got it half right" case - they set the HSM devId but not the KeyID. My original thoughts were just that if they specify the HSM devId, they are assuming operations should be RPC'ed to the HSM, full stop, and if anything were to prohibit that from happening, we should return an error instead of falling back to local crypto. IDK that is what seems right in my head. Current functionality is fine now, but we should def put a pin in this and flesh it out later.

bigbrett (approved):
vunderbar. Approving with a few comments that could be fixed in 1.1

bigbrett (approved):

PR #81, opened 2024-10-23
Fix ECC verify bug

bigbrett:
Fix bug where error codes returned from wolfCrypt were stored in an unsigned value, preventing the error from being detected and leading to read of garbage comm buffer data on the client side

### DIFF
--- a/src/wh_server_crypto.c
+++ b/src/wh_server_crypto.c
@@ -697,7 +697,7 @@ static int _HandleEccVerify(whServerContext* ctx, whPacket* packet,
     byte* res_pub       = (uint8_t*)(res + 1);
     word32 max_size     = (word32)(WOLFHSM_CFG_COMM_DATA_LEN -
                                 (res_pub - (uint8_t*)packet));
-    uint16_t pub_size   = 0;
+    uint32_t pub_size   = 0;
     int result;
 
     /* init public key */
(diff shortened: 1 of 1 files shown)


src/wh_server_crypto.c
```diff
@@ -697,7 +697,7 @@ static int _HandleEccVerify(whServerContext* ctx, whPacket* packet,
     byte* res_pub       = (uint8_t*)(res + 1);
     word32 max_size     = (word32)(WOLFHSM_CFG_COMM_DATA_LEN -
                                 (res_pub - (uint8_t*)packet));
-    uint16_t pub_size   = 0;
+    uint32_t pub_size   = 0;
```
billphipps, line 700:
```suggestion
   word32 pub_size   = 0;
```
Best to match the wc type here

billphipps (approved):
LGTM! Minor tweak to match types

src/wh_server_crypto.c
bigbrett: Match which wolfCrypt size? It isn't being passed to any wolfCrypt API and it being assigned to `res->pubSize`, which is a `uint32_t`? The size is returned from wolfCrypt as an `int` too, which gets stored in `ret`, then assigned to `pubSize`.

PR #84, opened 2024-11-14
(initial) ML-DSA support

bigbrett:
Adds initial support for ML-DSA to wolfHSM. Note that this requires https://github.com/wolfSSL/wolfssl/pull/8177 for wolfCrypt "blind deserialization" of ML-DSA DER keys. **CI tests will fail until this merges.** You can manually test against this branch if desired.

This PR does not include DMA support, which will be required (due to large key/signature sizes) for this to be reasonable to use on real hardware. I'm working on that part now, but wanted initial support to be merged in first.

This PR also fixes a few bugs in the transport mem and comm context layers, where improperly sized buffers could result in buffer overflows.

PR #86, opened 2024-11-21
ML-DSA DMA support

bigbrett:
- Adds support for DMA-based ML-DSA operations
- Adds support for ser/de of public and private only keys
- Adds additional testing
- Addresses  review feedback from #85 and #84
- Random cleanups and fixes


src/wh_client_cryptocb.c
```diff
@@ -752,21 +757,36 @@ static int _handleSha256Dma(wc_CryptoInfo* info, void* inCtx, whPacket* packet)
...
     int     size = info->pk.pqc_sig_kg.size;
     void*   key  = info->pk.pqc_sig_kg.key;
     int     type = info->pk.pqc_sig_kg.type;
 
+#ifndef WOLFHSM_CFG_DMA
+    if (useDma) {
+        /* TODO: proper error code? */
+        return WC_HW_E;
+    }
+#endif
```
billphipps, line 775:
```suggestion
#else
    (void)useDma;
#endif
```

src/wh_client_cryptocb.c
```diff
@@ -789,14 +809,29 @@ static int _handlePqcSign(whClientContext* ctx, wc_CryptoInfo* info)
...
-    WC_RNG*     rng     = info->pk.pqc_sign.rng;
     void*       key     = info->pk.pqc_sign.key;
     int         type    = info->pk.pqc_sign.type;
 
+#ifndef WOLFHSM_CFG_DMA
+    if (useDma) {
+        /* TODO: proper error code? */
+        return WC_HW_E;
+    }
+#endif
```
billphipps, line 820:
```suggestion
#else
    (void)useDma;
#endif
```

src/wh_client_cryptocb.c
```diff
@@ -851,14 +904,30 @@ static int _handlePqcSigCheckPrivKey(whClientContext* ctx, wc_CryptoInfo* info)
     word32      pubKeySz = info->pk.pqc_sig_check.pubKeySz;
     int         type     = info->pk.pqc_sig_check.type;
 
+#ifndef WOLFHSM_CFG_DMA
+    if (useDma) {
+        /* TODO: proper error code? */
+        return WC_HW_E;
+    }
+#endif
```
billphipps, line 912:
```suggestion
#else
    (void)useDma;
#endif
```

src/wh_client_cryptocb.c
```diff
@@ -851,14 +904,30 @@ static int _handlePqcSigCheckPrivKey(whClientContext* ctx, wc_CryptoInfo* info)
...
 #ifdef HAVE_DILITHIUM
         case WC_PQC_SIG_TYPE_DILITHIUM:
-            ret = wh_Client_MlDsaCheckPrivKey(ctx, key, pubKey, pubKeySz);
+#ifdef WOLFHSM_CFG_DMA
+            if (useDma) {
+                ret =
+                    wh_Client_MlDsaCheckPrivKeyDma(ctx, key, pubKey, pubKeySz);
+            }
+            else
+#endif /* WOLFHSM_CFG_DMA */
```
billphipps, line 923:
```suggestion
#else
    (void)useDma;
#endif /* WOLFHSM_CFG_DMA */
```

src/wh_crypto.c
```diff
@@ -257,7 +257,36 @@ int wh_Crypto_MlDsaSerializeKeyDer(MlDsaKey* key, uint16_t max_size,
...
 
-    ret = wc_Dilithium_KeyToDer(key, buffer, max_size);
+    /* Choose appropriate serialization based on key flags */
+    if (key->prvKeySet && key->pubKeySet) {
+#if defined(WOLFSSL_DILITHIUM_PRIVATE_KEY) && \
+    defined(WOLFSSL_DILITHIUM_PUBLIC_KEY)
+        /* Full keypair - use KeyToDer */
+        ret = wc_Dilithium_KeyToDer(key, buffer, max_size);
+#else
+        ret = WH_ERROR_NOHANDLER;
```
billphipps, line 267: Bad args instead?  This is a locally detected error.  No handler is expected as a server response when the client doesn't know any better.

wolfhsm/wh_packet.h
```diff
@@ -651,6 +651,115 @@ typedef struct wh_Packet_hash_sha256_Dma64_res {
     wh_Packet_Dma64_addr_status dmaAddrStatus;
 } wh_Packet_hash_sha256_Dma64_res;
 
+/* ML-DSA DMA key generation messages */
+typedef struct wh_Packet_pq_mldsa_keygen_Dma32_req {
```
billphipps, line 655: Make sure we add these new messages to the packed struct test file too.

src/wh_server_crypto.c
```diff
@@ -523,7 +523,14 @@ int wh_Server_MlDsaKeyCacheImport(whServerContext* ctx, MlDsaKey* key,
...
 
-    const uint16_t MAX_MLDSA_DER_SIZE = 5000;
+    const uint16_t MAX_MLDSA_DER_SIZE =
+#if !defined(WOLFSSL_NO_ML_DSA_87)
+        ML_DSA_LEVEL5_PRV_KEY_DER_SIZE;
+#elif !defined(WOLFSSL_NO_ML_DSA_65)
+        ML_DSA_LEVEL3_PRV_KEY_DER_SIZE;
+#else
+        ML_DSA_LEVEL2_PRV_KEY_DER_SIZE;
+#endif
```
billphipps, line 534: Shouldn't this be in a dilithium or asn header file?  Ok for now.

src/wh_server_crypto.c
```diff
@@ -1963,6 +2007,348 @@ static int _HandleSha256Dma(whServerContext* server, whPacket* packet,
...
+    wh_Packet_pq_mldsa_keygen_Dma32_req* req = &packet->pqMldsaKeygenDma32Req;
+    wh_Packet_pq_mldsa_Dma32_res*        res = &packet->pqMldsaDma32Res;
+#else
+    wh_Packet_pq_mldsa_keygen_Dma64_req* req = &packet->pqMldsaKeygenDma64Req;
+    wh_Packet_pq_mldsa_Dma64_res*        res = &packet->pqMldsaDma64Res;
+#endif
+
+    /* Check the ML-DSA security level is valid and supported */
+    if (0 == _IsMlDsaLevelSupported(req->level)) {
+        ret = WH_ERROR_BADARGS;
```
billphipps, line 2037: It's ok to return(BAD_ARGS) here instead of wait til the end.  No change necessary.  Just trying to save an indent

src/wh_server_crypto.c
```diff
@@ -1963,6 +2007,348 @@ static int _HandleSha256Dma(whServerContext* server, whPacket* packet,
...
+    evict  = !!(req->options & WH_PACKET_PQ_MLDSA_SIGN_OPTIONS_EVICT);
+
+    /* Initialize key */
+    ret = wc_MlDsaKey_Init(key, NULL, ctx->crypto->devId);
+    if (ret != 0) {
+        return ret;
+    }
+
+    /* Export key from cache */
+    /* TODO: sanity check security level against key pulled from cache? */
```
billphipps, line 2160: The Sign operation later should get the BAD_ARGS, no?  Ok to check it early, but not required.

src/wh_server_crypto.c
```diff
@@ -1963,6 +2007,348 @@ static int _HandleSha256Dma(whServerContext* server, whPacket* packet,
...
+}
+
+static int _HandleMlDsaCheckPrivKeyDma(whServerContext* server, whPacket* packet,
+                                        uint16_t* size)
+{
+    return WH_ERROR_NOHANDLER;
+}
+#endif /* HAVE_DILITHIUM */
+
+#if defined(HAVE_DILITHIUM) || defined(HAVE_FALCON)
```
billphipps, line 2312: wc should #define HAVE_PQC as the OR of any of the PQC algos.  No change.  Just a grumble.

billphipps (dismissed):
LGTM!  Some minor nits.  Let me know when you want me to merge

src/wh_client_cryptocb.c
bigbrett: I don't think we need to void this here, since this #else would be the case where `WOLFHSM_CFG_DMA` is defined, and that flag will be checked in the `if` statement below. Maybe you missed the macro switch was `ifndef WOLFHSM_CFG_DMA` and not `ifdef ...`? 

The idea is that if you don't have `WOLFHSM_CFG_DMA` defined and you pass the `useDmaFlag` then the function returns error. If `WOLFHSM_CFG_DMA` is defined, then that argument will always be checked, so don't think we need the `(void)` cast on an unused arg.

src/wh_crypto.c
bigbrett: Yeah you are right. I feel like we need a "not compiled in" error, to somehow signify that "user input is valid (e.g. not malformed) but not supported by the current build configuration". I'll change to BADARGS for now.

billphipps (approved):

src/wh_client_cryptocb.c
billphipps: Hah!!  Yup.  I was doing the opposite logic.  No changes necessary!

PR #100, opened 2025-04-02
fix compile error for CC-RH

miyazakh:


miyazakh:
CC-RH doesn't allow the following code:
```
const uint16_t MAX_DER_SIZE = 128;
...
uint8_t der_buf[MAX_DER_SIZE];
```

I am not sure if this is intentional. But, this causes an error `-132` from `wc_Curve25519KeyToDer`
```
uint16_t       keySz        = keySz;
```

src/wh_server_crypto.c
```diff
@@ -455,8 +455,12 @@ int wh_Server_CacheImportCurve25519Key(whServerContext* server,
     int            ret;
     /* Max size of a DER encoded curve25519 keypair with SubjectPublicKeyInfo
      * included. Determined by experiment */
+#if !defined(__CCRH__)
     const uint16_t MAX_DER_SIZE = 128;
-    uint16_t       keySz        = keySz;
+#else
+    #define MAX_DER_SIZE 128
+#endif
+    uint16_t       keySz        = MAX_DER_SIZE;
```
billphipps, line 463: Grumble.  The maximum buffer size needed to hold a curve25519 private/public key pair SHOULD be in wolfcrypt/curve25519.h as something like CURVE25519_BUFSIZE, following the similar ECC_BUFSIZE.  It should also be declared as an enum, rather than const int nor #define.

Recommend to change this to:
enum { CURVE25519_BUFSIZE=128};
This is compatible with all compilers.  I'll see if I can put this into wolfcrypt.

billphipps (changes requested):
Looks good.  Recommend to change to enum.

src/wh_server_crypto.c
miyazakh: Agree. We should have consistency.

billphipps:
Recommend to close this PR.  The guard code in wh_test_crypto.c was added as part of wolfHSM PR#102 and the added enum to wolfCrypt (CURVE25519_BUFSIZE) will be added as wolfssl PR#8651.

PR #106, opened 2025-05-01
quickfix: compiler warnings for 32-bit targets

bigbrett:
Fixes warnings on 32-bit targets

### DIFF
--- a/src/wh_client_cryptocb.c
+++ b/src/wh_client_cryptocb.c
@@ -364,14 +364,14 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
     case WC_ALGO_TYPE_CMAC:
     {
         /* Extract info parameters */
-        const uint8_t* in = info->cmac.in;
-        uint32_t in_len = (in == NULL) ? 0 : info->cmac.inSz;
-        const uint8_t* key = info->cmac.key;
-        uint32_t key_len = (key == NULL) ? 0 : info->cmac.keySz;
-        uint8_t* outMac = info->cmac.out;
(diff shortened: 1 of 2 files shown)


src/wh_client_cryptocb.c
```diff
@@ -364,14 +364,14 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
-        uint8_t* outMac = info->cmac.out;
-        word32 *out_mac_len = info->cmac.outSz;
-        Cmac* cmac = info->cmac.cmac;
-        int type = info->cmac.type;
+        const uint8_t* in          = info->cmac.in;
+        uint32_t       in_len      = (in == NULL) ? 0 : info->cmac.inSz;
+        const uint8_t* key         = info->cmac.key;
+        uint32_t       key_len     = (key == NULL) ? 0 : info->cmac.keySz;
+        uint8_t*       outMac      = info->cmac.out;
+        uint32_t*      out_mac_len = (uint32_t*)info->cmac.outSz;
```
billphipps, line 372: Hmm.  If there is a discrepancy between the size of word32 and uint32_t, then this could be a problem.  Worst case, provide a local uint32_t to pass into our function call and then info->cmac.outSz = localuint32_t;

What was the warning?

billphipps (approved):

src/wh_client_cryptocb.c
bigbrett: @billphipps how could there be a discrepancy between sizeof(word32) and sizeof(uint32_t)? Aren't they both explicitly 32 bit hah?

PR #109, opened 2025-05-10
set RNG on curve25519 keys to support blinding

bigbrett:
[Upstream wolfSSL ](https://github.com/wolfSSL/wolfssl/pull/8736)broke wolfHSM after it turned on curve25519 blinding by default, which requires setting RNG on key context for shared secret generation. Found out the hard way after updating wolfSSL to latest master. This PR fixes.


billphipps (approved):
All good!

PR #110, opened 2025-05-10
new x509 API: verify and cache pubKey 

bigbrett:
Adds new API for verification that instructs the server to  cache the public key of the leaf cert if the provided cert chain verifies. Requires new wolfCrypt API introduced in https://github.com/wolfSSL/wolfssl/pull/8758, so that should merge first and keeping as draft until then.

Needed for upcoming wolfBoot + wolfHSM cert chain verification of firmware images, as wolfBoot will use the leaf cert public key to verify firmware authenticity after the chain is verified.

I thought about just adding this as a flag to the existing verify API, but ended up splitting it into a new API function for backwards compat. Can revert to doing it that way instead if you think it is better, just LMK.


src/wh_client_cert.c
```diff
@@ -329,9 +329,10 @@ int wh_Client_CertReadTrusted(whClientContext* c, whNvmId id, uint8_t* cert,
     return rc;
 }
 
-/* Verify a certificate */
-int wh_Client_CertVerifyRequest(whClientContext* c, const uint8_t* cert,
-                                uint32_t cert_len, whNvmId trustedRootNvmId)
+/* Helper function to send a verify request */
+int _certVerifyRequest(whClientContext* c, const uint8_t* cert,
```
billphipps, line 333: Should these be static?  If so, add declarations at the top plz.

src/wh_client_cert.c
```diff
@@ -381,15 +384,18 @@ int wh_Client_CertVerifyResponse(whClientContext* c, int32_t* out_rc)
...
         }
     }
 
     return rc;
 }
 
-int wh_Client_CertVerify(whClientContext* c, const uint8_t* cert,
-                         uint32_t cert_len, whNvmId trustedRootNvmId,
-                         int32_t* out_rc)
+int _certVerify(whClientContext* c, const uint8_t* cert, uint32_t cert_len,
```
billphipps, line 396: static?

src/wh_server_cert.c
```diff
@@ -99,6 +102,51 @@ static int _verifyChainAgainstCmStore(WOLFSSL_CERT_MANAGER* cm,
...
+                    }
+                }
+
+                if (rc != WH_ERROR_OK) {
+                    wc_FreeDecodedCert(&dc);
+                    return rc;
+                }
+
+                /* Propagate cached keyId back to client */
+                *out_keyId = WH_KEYID_ID(keyId);
```
billphipps, line 148:
```suggestion
                if (rc != WH_ERROR_OK) {
                    /* Propagate cached keyId back to client */
                    *out_keyId = WH_KEYID_ID(keyId);
                }
```

test/wh_test_cert.c
```diff
@@ -132,75 +142,86 @@ int whTest_CertClient(whClientContext* client)
...
+
+    /* Export the cached public key and verify it matches the expected leaf
+     * public key */
+    WH_TEST_RETURN_ON_FAIL(wh_Client_KeyExport(
+        client, out_keyId, NULL, 0, exportedPubKey, &exportedPubKeyLen));
+
+    WH_TEST_ASSERT_RETURN(exportedPubKeyLen == LEAF_A_PUBKEY_len);
+    WH_TEST_ASSERT_RETURN(
+        0 == memcmp(exportedPubKey, LEAF_A_PUBKEY, LEAF_A_PUBKEY_len));
+
```
billphipps, line 216: Need to evict the cached public key?

test/wh_test_cert.c
```diff
@@ -288,75 +312,87 @@ int whTest_CertClientDma_ClientServerTestInternal(whClientContext* client)
...
+    /* Export the cached public key and verify it matches the expected leaf
+     * public key */
+    WH_TEST_RETURN_ON_FAIL(wh_Client_KeyExportDma(
+        client, out_keyId, exportedPubKey, sizeof(exportedPubKey), NULL, 0,
+        &exportedPubKeyLen));
+
+    WH_TEST_ASSERT_RETURN(exportedPubKeyLen == LEAF_A_PUBKEY_len);
+    WH_TEST_ASSERT_RETURN(
+        0 == memcmp(exportedPubKey, LEAF_A_PUBKEY, LEAF_A_PUBKEY_len));
+
```
billphipps, line 387: Evict cached pub key?

wolfhsm/wh_message_cert.h
```diff
@@ -155,12 +165,25 @@ typedef struct {
     uint64_t cert_addr;
     uint32_t cert_len;
     whNvmId  trustedRootNvmId;
+    uint16_t flags;
     uint8_t  WH_PAD[2];
```
billphipps, line 169: Hrmm. Not sur e why this pad is still here?  Shouldn't this cause a padding warning?

wolfhsm/wh_server_cert.h
```diff
@@ -80,7 +80,8 @@ int wh_Server_CertReadTrusted(whServerContext* server, whNvmId id,
  * @return WH_ERROR_OK on success, error code on failure
  */
 int wh_Server_CertVerify(whServerContext* server, const uint8_t* cert,
-                         uint32_t cert_len, whNvmId trustedRootNvmId);
+                         uint32_t cert_len, whNvmId trustedRootNvmI,
```
billphipps, line 83: I think copilot was right! This should be renamed to trusted RootNvmId, no?

billphipps (changes requested):
Looks really good!  Great that this feature is so quickly exported into wolfHSM.  Can you add the new messages to the padding test and add static where necessary?

src/wh_server_cert.c
bigbrett: @billphipps confused about this one... I don't think we would want to propagate that keyId back to the client because we never end up writing something to it if rc != WH_ERROR_OK. So we would be lying to them about the cache slot. That should only be updated on success?

wolfhsm/wh_message_cert.h
bigbrett: it would have if I ever added any of the structs lol. Fixed

wolfhsm/wh_server_cert.h
bigbrett: yay AI. Fixed

test/wh_test_cert.c
bigbrett: good catch. FIxed

test/wh_test_cert.c
bigbrett: good catch, fixed

src/wh_client_cert.c

src/wh_client_cert.c

PR #123, opened 2025-06-19
server keystore/cert fixes and a fix for test breakage due to upstream wolfCrypt

bigbrett:
- Refactors `wh_Server_CertVerify` to operate on raw (server) keyIds so it can be used outside a client message handler. Relocates the client keyId mapping on input/output to the message dispatch function better aligning with the rest of the code.
- Adds keyId parameter to client verifyAndCachePubKey API in case client wants to cache to a specific keyId, thus matching the server implementation
- Adds fix to guard against condition where a client could potentially cache two keys with the same ID in both little and big key caches
- Adds test coverage for the above fix

Also unrelated fix: wolfCrypt upstream changed default RNG seed generation causing our wolfCrypt tests to break. I removed NO_FILESYSTEM from the POSIX tests so it can use `/dev/random` for seeding. I added in the use of the cert buffers macro as otherwise filesystem support requires you to run wolfCrypt tests from the wolfCrypt root directory which we don't want.


benchmark/config/user_settings.h
```diff
@@ -65,15 +65,24 @@ extern "C" {
...
+#define USE_CERT_BUFFERS_1024
+#define USE_CERT_BUFFERS_2048
+#define USE_CERT_BUFFERS_3072
+#define USE_CERT_BUFFERS_4096
+
 /** Remove unneeded features*/
 #define NO_MAIN_DRIVER
 #define NO_ERROR_STRINGS
 #define NO_ERROR_QUEUE
-#define NO_FILESYSTEM
```
billphipps, line 72: Are we sure this is what we want?  We don't really expect or need a filesystem for anything within wolfHSM

benchmark/config/user_settings.h
```diff
@@ -65,15 +65,24 @@ extern "C" {
...
 #define NO_INLINE
 #define NO_OLD_TLS
 #define WOLFSSL_NO_TLS12
 #define NO_DO178
+#define NO_WRITE_TEMP_FILES
```
billphipps, line 84: If this is only used for testing, let's move it up with the USE_CERT_BUFFERS area so it's clear it has nothing to do with the runtime.  Probably other stuff like NO_MAIN_DRIVER is in the same boat.

src/wh_client_cert.c
```diff
@@ -411,21 +412,30 @@ static int _certVerifyResponse(whClientContext* c, whKeyId* out_keyId,
...
     if ((c == NULL) || (cert == NULL) || (cert_len == 0)) {
         return WH_ERROR_BADARGS;
     }
 
+    if (inout_keyId != NULL) {
+        keyId = *inout_keyId;
+    }
+    else {
+        keyId = WH_KEYID_ERASED;
+    }
```
billphipps, line 429:
```suggestion
```

Already done above

src/wh_server_cert.c
```diff
@@ -264,6 +264,12 @@ int wh_Server_CertVerify(whServerContext* server, const uint8_t* cert,
         return WH_ERROR_BADARGS;
     }
 
+    /* If the leaf public key is to be cached, then the user must provide a
+     * keyId */
+    if ((flags & WH_CERT_FLAGS_CACHE_LEAF_PUBKEY) && (inout_keyId == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
```
billphipps, line 271: I get this is a sanity check by the server.  Is this also caught on the client side before the request is sent?

src/wh_server_keystore.c
```diff
@@ -183,6 +183,24 @@ int wh_Server_KeystoreCacheKey(whServerContext* server, whNvmMetadata* meta,
         return WH_ERROR_BADARGS;
     }
 
+    /* Check for cross-cache duplicates to prevent same keyId in both caches */
```
billphipps, line 186: Great idea to check the caches for duplicates, but returning an error when attempting to cache a different-sized key is a different mechanic than what we have in nvm.  Here's an example:
nvm->AddObject(meta.id1, stuff) -> add this object into nvm
nvm->AddObject(meta.id1, new stuff) -> replace/update id1 with the new stuff

Shouldn't the cache work the same way?  The size of the caches isn't necessarily known by the user, so we would expect the user to handle the duplicate error by Evicting the keyid and then trying to Cache the key again, no?

Can you consider the server handling a request to reuse an existing keyid to first remove it from either cache, then insert the new one into the proper cache?   This would eliminate the need for WH_ERROR_DUPLICATE

billphipps (changes requested):
Looks good!  Can you consider my comments about duplicate?  I'm good either way.

benchmark/config/user_settings.h
bigbrett: as discussed on slack - leaving this in due to POSIX test specific behavior

benchmark/config/user_settings.h
bigbrett: fixed

src/wh_client_cert.c

PR #131, opened 2025-07-09
bugfixes for SHE server

DrKatieL:
added error checking and restructured top level call to improve reliability and readability

### DIFF
--- a/src/wh_server_she.c
+++ b/src/wh_server_she.c
@@ -24,8 +24,6 @@
 /* Pick up compile-time configuration */
 #include "wolfhsm/wh_settings.h"
 
-#if !defined(WOLFHSM_CFG_NO_CRYPTO) && defined(WOLFHSM_CFG_ENABLE_SERVER)
-
 /* System libraries */
 #include <stdint.h>
 #include <stddef.h> /* For NULL */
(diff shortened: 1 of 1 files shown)


src/wh_server_she.c
```diff
@@ -1545,90 +1403,29 @@ int wh_Server_HandleSheRequest(whServerContext* server, uint16_t magic,
...
-        return WH_ERROR_BADARGS;
+    if (server == NULL || server->she == NULL|| req_packet == NULL || resp_packet==NULL) {
+        return WH_ERROR_BADARGS;//*out_resp_size is directly written to so it's OK if it's null
     }
 
-    ret = _ReportInvalidSheState(server, magic, action, req_size, req_packet,
-                                 out_resp_size, resp_packet);
-    if (ret != 0) {
-        return ret;
+    // _ReportInvalidSheState function was mostly checking if endianness of client message is right and fixing it if it isn't but just for certain parts.
```
bigbrett, line 1410: Please don't refer to code in a comment that you removed, as it is no longer present and serves no purpose. Also, as a side note, this interpretation of what the `_ReportInvalidSheState()` function did is incorrect. The point of that function was to detect if the request from the client was invalid based on the current SHE state and to report the sequence error back to the client.

PR #142, opened 2025-08-06
add size checking for AesCbc request

rizlik:
Test is somehow hackish.

Probably other type of request need to be fixed as well.

### DIFF
--- a/src/wh_server_crypto.c
+++ b/src/wh_server_crypto.c
@@ -1260,6 +1260,12 @@ static int _HandleAesCbc(whServerContext* ctx, uint16_t magic, const void* crypt
     uint32_t enc     = req.enc;
     uint32_t key_len = req.keyLen;
     uint32_t len     = req.sz;
+    uint64_t needed_size = sizeof(whMessageCrypto_AesCbcResponse) + len +
+                          key_len + AES_BLOCK_SIZE;
+    if (needed_size > inSize) {
+        return WH_ERROR_BADARGS;
+    }
+
     whKeyId  key_id =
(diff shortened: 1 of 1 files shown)


bigbrett (changes requested):
First off, could we pls put this new test in wh_test_crypto.c alongside the other crypto tests?

Second, I agree with you that the level of hackery for this test is high and I'm debating if it should even be included. Could you clarify what usage of the client API leads into the case where the input size doesn't match like you are guarding against? Would they need to use a similar method that your test does? Could you update the PR description with a little more context demonstrating the usage pattern this check guards against?

rizlik:
> First off, could we pls put this new test in wh_test_crypto.c alongside the other crypto tests?
> 
> Second, I agree with you that the level of hackery for this test is high and I'm debating if it should even be included. 

Indeed after small internal discussion with @billphipps  we decided that this would better handled by testing the lower level function. I opened an issue #143 to keep track of this. 

>Could you clarify what usage of the client API leads into the case where the input size doesn't match like you are guarding against? Would they need to use a similar method that your test does? Could you update the PR description with a little more context demonstrating the usage pattern this check guards against?

This guards against malicious clients that use the server-side buffer overflow to attack the server, so it's not triggered by any normal API usage.

bigbrett:
@rizlik okay, if we are going to add a uniform testing solution for all of this, it will likely be in its own standalone test harness. In that case can you just remove the new "hacky" test from this PR and I can merge? Will be simpler and not require knowingly merging code we are about to "undo".

PR #146, opened 2025-08-11
server: add dma memcpy callback and add guard for CMAC DMA and KEY DMA

rizlik:


wolfhsm/wh_settings.h
```diff
@@ -67,6 +67,16 @@
  *  WOLFHSM_CFG_SERVER_IMG_MGR_MAX_SIG_SIZE - Maximum signature size for image
  * verification Default: 512 bytes (RSA4096)
  *
+ *  WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY - if defined, allows to setup a custom
+ * callback to handle client to server and/or server to client memory copy
+ * operation in DMA requests.
+ *     Default: Not defined
+ *
+ *  WOLFHSM_CFG_NO_CMAC_DMA - if defined, disables the use of CMAC for DMA
```
bigbrett, line 75:
```suggestion
 *  WOLFHSM_CFG_NO_CMAC_DMA - if defined, disables CMAC DMA support
```

wolfhsm/wh_settings.h
```diff
@@ -67,6 +67,16 @@
...
+ *     Default: Not defined
+ *
+ *  WOLFHSM_CFG_NO_KEY_DMA  - if defined, disables the use of DMA for keys
```
bigbrett, line 78: Seems like this macro just disables the DMA keystore test and doesn't actually compile out support for DMA keystore operations? If that is the case maybe we shoudl make it a test config option (`WOLFHSM_CFG_TEST_NO_KEYSTORE_DMA`).

Could you clarify the need for this macro? Is it because the keystore layer directly copies to client memory and doesn't go through the copy two/from client functions (since NVM layer directly exports) and that will break on your platform? 

I think if that is the case then we should be able to fully compile out DMA support from the keystore layer, not just from the tests.

More broadly, I wonder if we should instead be able to support this use case fully instead of just turning off tests, or adding macros to turn off individual problematic funcitons. Some sort of configuration where there is still "DMA" but only through a single controlled interface and never directly accessed by pointer.

wolfhsm/wh_server.h
```diff
@@ -384,6 +405,23 @@ int wh_Server_HandleCustomCbRequest(whServerContext* server, uint16_t magic,
...
                             whServerDmaClientMemCb    cb);
 
+
+#ifdef WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY
+/**
+ * @brief Registers a custom memory copy callback for DMA operations.
+ * This function allows the server to register a callback that will be invoked
+ * during DMA memory copy operations. The callback can be used to
+ * perform custom memory copy operations, such as remapping addresses
+ * or handling special cases.
```
bigbrett, line 415:
```suggestion
 * @brief Registers a custom memory copy callback for DMA operations.
 * This function allows the server to register a callback that will be invoked
 * during DMA memory copy operations. The callback overrides the use of
 * memcpy when copying to and from client memory. This is useful if standard
 * memcpy cannot be used to copy data back and forth between the client, even
 * after client addresses are transformed through the standard DMA callbacks
 * (e.g. if client memory can only be accessed though a hardware FIFO or register
 * interface)
```

wolfhsm/wh_server.h
```diff
@@ -129,6 +129,13 @@ typedef enum {
     WH_DMA_OPER_CLIENT_WRITE_POST = 3,
 } whServerDmaOper;
 
+#ifdef WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY
+typedef enum {
+    WH_DMA_OPER_CLIENT_READ = 0,
+    WH_DMA_OPER_CLIENT_WRITE = 1,
```
bigbrett, line 135: Clarify in the name that these are only useable for the copy callback (don't want users passing them to regular callbacks). Sure wish we had real typed enums to work with...
```suggestion
    WH_DMA_COPY_OPER_CLIENT_READ = 0,
    WH_DMA_COPY_OPER_CLIENT_WRITE = 1,
```

wolfhsm/wh_settings.h
```diff
@@ -67,6 +67,16 @@
  *  WOLFHSM_CFG_SERVER_IMG_MGR_MAX_SIG_SIZE - Maximum signature size for image
  * verification Default: 512 bytes (RSA4096)
  *
+ *  WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY - if defined, allows to setup a custom
+ * callback to handle client to server and/or server to client memory copy
+ * operation in DMA requests.
+ *     Default: Not defined
```
bigbrett, line 73: Somewhat similar comment to the below one on the DMA keystore.  Not sure if we want to be in the business of making every individual DMA crypto algorithm potentially turned off via its own config macro. That will result in a lot of clutter. Perhaps a more generic solution is in order. And how come this isn't an issue for SHA256?

wolfhsm/wh_settings.h
rizlik: You are completely right to be confused on the need of this macros, I don't thing they are needed and I'm reverting the commit.
NO_KEY_DMA  was used during development because I was working without the dma copy callaback function and tests were failing, now tests are working.
NO_CMAC_DMA tests were also failing, and I noticed that the CMAC DMA operation doesn't fallback to sw implementation (in the server) or Non-DMA version (in the client) in case the CMAC DMA is not implemented in the server. So I though it was a good idea to have a way to disable it, but I did it very quickly and I poorly double checked what I did before doing this PR. Sorry for that. I'm reverting this commit, I can keep the development by just disabling CMAC in wolfSSL until implementing CMAC DMA.
TL;DR: The macros NO_KEY_DMA and NO_CMAC_DMA were temporary hacks for development and testing. They’re unnecessary now, tests pass without them, and the commit adding them is being reverted. CMAC DMA will just be disabled in wolfSSL until it’s properly implemented.

rizlik:
> A few nits and some edits for clarity. I also bring up a global architecture question for discussion.

Thanks for the review, fixed nits and reverted unneeded macros, let me know if you want me to do some git history cleaning before merging.

PR #147, opened 2025-08-12
non-exportable NVM

bigbrett:
- Add support for new "non-exportable" flag to NVM objects, keys, and certs
- Fixes unrelated tests to accommodate new feature


src/wh_client_cert.c
```diff
@@ -115,7 +115,8 @@ int wh_Client_CertInit(whClientContext* c, int32_t* out_rc)
 
 /* Add a trusted certificate */
 int wh_Client_CertAddTrustedRequest(whClientContext* c, whNvmId id,
-                                    const uint8_t* cert, uint32_t cert_len)
+                                    const uint8_t* cert, uint32_t cert_len,
+                                    whNvmFlags flags)
```
billphipps, line 119: Couple of thoughts:
1. Perhaps we should expose/add whAccess and label to this API now, even if they are likely to be defaults/null.
2. Should the "metadata" stuff be before the "data" stuff, like in the AddObject?  Consider moving the access, flags, and label in front of the cert* and cert_len.

src/wh_server_keystore.c
```diff
@@ -726,6 +726,15 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
                 WH_MAKE_KEYID(WH_KEYTYPE_CRYPTO, server->comm->client_id,
```
billphipps, line 726: Hrmmm.   I had expected the same sequence as below in DMA:
1. _FindInCache(id, meta, &stored_stuff)
2. Check Meta.flags & nonexport
3. memcpy (outbuffer, stored_stuff).
But, the KeyStoreReadKey doesn't use FindInCache, embeds the memcpy, and has special junk for SHE.  

No Change necessary at this point, but consider using the same sequence as in DMA and possibly refactor ready to do basically the same thing.

src/wh_client_cert.c

PR #157, opened 2025-09-02
Keywrap feature

AlexLanzano:


bigbrett (changes requested):
Amazing work @AlexLanzano. A few final tweaks and it is good to go

PR #158, opened 2025-09-02
add DMA by offset feature, SHM POSIX example, and support for args with benchmarking

JacobBarthelmeh:
This PR adds support for SHM (Shared Memory) POSIX DMA transport to wolfHSM, enabling pass-by-reference communication between client and server through shared memory segments with optimized memory allocation.

Implements DMA transport layer with static memory management for improved performance
Adds client/server configuration infrastructure for multiple transport types (TCP, SHM, DMA)
Enhances POSIX examples to support configurable transport types with command-line options


examples/posix/README.md
```diff
@@ -0,0 +1,15 @@
+# POSIX Examples
+
+There are a couple of different ways POSIX can be used for transport.
+
+- TCP
+- SHM (Shared memory)
+- REF (Shared memory pass by reference)
```
billphipps, line 7: Is this the DMA version?

examples/posix/wh_example_posix.h
billphipps, line 1: Recommend to refactor this file to separate configuration constants from structures. Let me talk with you to describe what I think is the best solution.

examples/posix/wh_server_posix/wh_server_posix.c
```diff
@@ -25,6 +25,8 @@
 #include "wolfhsm/wh_nvm_flash.h"
 #include "wolfhsm/wh_flash_ramsim.h"
 
+#include "wolfssl/version.h"
```
billphipps, line 28: make sure we have no includes from wolfssl/wolfcrypt when WOLFHSM_CFG_NO_CRYPTO is set.  Make sure you need this header.

examples/Makefile
```diff
@@ -1,15 +1,15 @@
 
-.PHONY: all posix_tcp_wh_client_tcp posix_tcp_wh_server_tcp clean
+.PHONY: all posix_wh_client_posix posix_wh_server_posix clean
```
bigbrett, line 2: why the double posix in the target names?

examples/README.md
```diff
@@ -11,6 +11,9 @@ Currently, the only public example for wolfHSM uses the POSIX simulator. If you
 ### Posix TCP server and client
 This example spawns a wolfHSM client and server, both in their own thread, and runs the client-side unit tests against the server.
 
+### Posix SHM server and client
+This example spawns a wolfHSM client and server, both in their own thread, and runs the client-side unit tests against the server.
```
bigbrett, line 15:
```suggestion
This example spawns a wolfHSM client and server, both in their own thread, and runs the client-side unit tests against the server using the POSIX SHM transport
```

examples/posix/README.md
```diff
@@ -0,0 +1,15 @@
+# POSIX Examples
+
+There are a couple of different ways POSIX can be used for transport.
+
+- tcp
+- shm (Shared memory)
+- dma (Shared memory pass by reference)
```
bigbrett, line 7: It isn't just shared memory pass by reference, but also using the wolfCrypt static memory allocator

examples/posix/wh_posix_client/user_settings.h
```diff
@@ -47,6 +47,12 @@
 #define NO_MAIN_DRIVER
 #define BENCH_EMBEDDED
 
+#ifdef WOLFHSM_CFG_DMA
+#undef WOLFSSL_STATIC_MEMORY
+#define WOLFSSL_STATIC_MEMORY
+#define WOLFSSL_STATIC_MEMORY_TEST_SZ 100000
+#endif
```
bigbrett, line 54: hmmm - I don't love gating the wolfSSL configuration on wolfHSM configuration...not sure how else to make this dynamic though.

examples/posix/wh_posix_client/wh_posix_client.c
```diff
@@ -0,0 +1,212 @@
...
+            return -1;
+        }
+#else
+        return -1;
+#endif
+    }
+
+    printf("Client connecting to server...\n");
+    if (test) {
+        return wh_DemoClient_wcTest(client);
```
bigbrett, line 89: we should probably just do `wh_DemoClient_All(..)`?

examples/posix/wh_posix_client/wh_posix_client_cfg.h
```diff
@@ -0,0 +1,8 @@
+#ifndef WH_POSIX_CLIENT_CFG_H
+#define WH_POSIX_CLIENT_CFG_H
+
+int Client_ExampleDMAConfig(void* c_conf);
+int Client_ExampleSHMConfig(void* c_conf);
+int Client_ExampleTCPConfig(void* c_conf);
+int Client_ExampleSetupDmaMemory(void* ctx, void* c_conf);
```
bigbrett, line 7: This doesn't match our naming conventions. `wh_PosixClient_InitExampleXXXConfig()` perhaps?

examples/posix/wh_posix_server/wh_posix_server_cfg.c
```diff
@@ -1,50 +1,132 @@
...
+posixTransportTcpServerContext tscTcp;
+
+#ifdef WOLFSSL_STATIC_MEMORY
+whTransportServerCb            dmaCb = POSIX_TRANSPORT_SHM_SERVER_CB;
+posixTransportShmServerContext tscDma;
+whServerDmaConfig              dmaConfig;
+
+/* Server configuration setup example for transport
+ * Does not setup flash, nvm, crypto, she, etc. */
+int Server_ExampleDMAConfig(void* conf)
```
bigbrett, line 35: same comment as on client for naming

benchmark/wh_bench.c
```diff
@@ -39,6 +39,11 @@
 #include "wolfhsm/wh_client.h"
 #include "wolfhsm/wh_utils.h"
 
+/* Include transport-specific headers */
+#include "port/posix/posix_transport_shm.h"
+#include "port/posix/posix_transport_tcp.h"
+#include "port/posix/posix_transport_dma.h"
```
bigbrett, line 45: we can't unconditionally include POSIX files as the benchmarks will be directly compiled by non-POSIX targets (bare metal). Should be gated by `WOLFHSM_CFG_TEST_POSIX`

benchmark/wh_bench.c
```diff
@@ -477,6 +527,28 @@ int wh_Bench_ClientCfg(whClientConfig* clientCfg)
         return ret;
     }
 
+#ifdef WOLFSSL_STATIC_MEMORY
```
bigbrett, line 530: This section should be gated on `WOLFHSM_CFG_TEST_POSIX` too, not just static memory right? It seems that we only support this static memory functionality when used on top of the POSIX SHM transport using the DMA callbacks in `wh_posix_transport_dma`?

benchmark/wh_bench.c
```diff
@@ -459,7 +509,7 @@ static int _runClientBenchmarks(whClientContext* client)
 
 /* Initializes a client context based on the provided config, runs the
  * benchmarks, then cleans up the context */
-int wh_Bench_ClientCfg(whClientConfig* clientCfg)
+int wh_Bench_ClientCfg(whClientConfig* clientCfg, int transport)
```
bigbrett, line 512: We should check the transport against the enabled build options - for example if you pass the POSIX transport enum on a non-POSIX system (e.g. bare metal) then this should throw an error.

port/posix/posix_transport_dma.h
```diff
@@ -0,0 +1,79 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with wolfHSM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * port/posix_transport_dma.h
```
bigbrett, line 21: 1. We need to make this module name (`posix_transport_dma`) a lot more descriptive. It really isn't its own transport, but rather is an implementation of the DMA callbacks for use with the standard posix SHM transport that integrates with/supports the wolfCrypt static memory allocator.
2. Similar to 1., the function names in this file should indicate they are specifically meant to be used as DMA callbacks 
3. Since this isn't really its own transport, consider just adding these functions to `posix_transport_shm` as DMA callback implementations?

bigbrett:
@JacobBarthelmeh one additional tweak now that #172 merged: could you please rebase on main and ensure the POSIX SHM client and server init functions are protected by `WOLFHSM_CFG_ENABLE_CLIENT` and `WOLFHSM_CFG_ENABLE_SERVER`? Those got missed in that PR and it is somewhat relevant to the code you are touching. Thanks!

wolfhsm/wh_client.h
```diff
@@ -2194,6 +2272,70 @@ int wh_Client_CertVerifyAcertDmaRequest(whClientContext* c, const void* cert,
...
+                            whClientDmaClientMemCb    cb);
+
+
+/**
+ * @brief Processes a client address for DMA operations, using the native
+ * pointer size of the system
+ *
+ * This function transforms a client address for DMA operations. It performs
+ * user-supplied address transformations, cache manipulations, and checks the
+ * transformed address against the server's allowlist if registered.
```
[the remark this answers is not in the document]
JacobBarthelmeh: For the wh_Client_DmaProcessClientAddress function my thoughts is it is the same polarity as what the server side does (client address in - > server understandable address out). Just doing the processing before crafting the message that gets sent to the server. Updated the comment though, it should have said the client's allowlist if registered was being used to check if the address was allowed.

src/wh_client_crypto.c
```diff
@@ -2832,8 +2857,22 @@ int wh_Client_Sha256Dma(whClientContext* ctx, wc_Sha256* sha, const uint8_t* in,
...
+        wh_Client_DmaProcessClientAddress(
+            ctx, (uintptr_t)sha256, (void**)&stateAddr, req->state.sz,
+            WH_DMA_OPER_SERVER_WRITE_POST, (whClientDmaFlags){0});
+        wh_Client_DmaProcessClientAddress(
+            ctx, (uintptr_t)in, (void**)&inAddr, req->input.sz,
+            WH_DMA_OPER_SERVER_READ_POST, (whClientDmaFlags){0});
+        wh_Client_DmaProcessClientAddress(
+            ctx, (uintptr_t)out, (void**)&outAddr, req->output.sz,
+            WH_DMA_OPER_SERVER_WRITE_POST, (whClientDmaFlags){0});
+    }
```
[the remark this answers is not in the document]
JacobBarthelmeh: I made this (void) and left a comment about the reasoning behind not checking the return values. They honestly would not need called if not for the sake of giving the DMA callback function an opportunity to do post transformation operations / free'ing.

test/wh_test_clientserver.c
```diff
@@ -742,12 +742,17 @@ int whTest_ClientServerSequential(void)
     }};
 #endif
 
-    whServerConfig  s_conf[1] = {{
-         .comm_config = cs_conf,
-         .nvm         = nvm,
+    whServerDmaConfig dma_config = {
+        .cb               = NULL, /* Disable DMA callback for test */
+        .dmaAddrAllowList = NULL};
```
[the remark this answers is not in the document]
JacobBarthelmeh: Was added when debugging during development. Removed it now, it's unnecessary.

bigbrett (dismissed):
Good on my end, over to @billphipps for final review/merge

wolfhsm/wh_client.h
```diff
@@ -2194,6 +2224,70 @@ int wh_Client_CertVerifyAcertDmaRequest(whClientContext* c, const void* cert,
...
+/**
+ * @brief Getter function for the current heap hint set
+ *
+ * This function is used to get the current heap hint set for use with the
+ * transport context.
+ *
+ * @param[in] c Pointer to the client context.
+ * @return a pointer to the heap hint
+ */
+void* wh_Client_GetHeap(whClientContext* c);
```
billphipps, line 2237: Hrmm.  Not convinced that the base client context needs this information.  Seems like the transport COULD provide a heap area to support DMA transactions.  On other hardware, the entire client RAM space is globally mapped already so there are no restrictions for the HSM to use it.  If we are simply using this getter/setter to API to help associate the external usage of the shared memory area (and no INTERNAL client logic depends on it), then this should be exposed by the posix shm transport privately, not via an exposed public transport API.  The user will know it's using a DMA-capable SHM transport since they are setting up the context structures already and can then directly grab the heap hint from the posix shm transport object after initialization.

Lemme stew on this.

billphipps (dismissed):
Looks really good!  Consider my comments on heap pointer within the base client context and how you could solve this use case a different way.  if there is no other way, then i'm good as is.

PR #162, opened 2025-09-03
Intialized req and resp as well as potenial int overflow issues

jackctj117:
Based on coverity report I added some initialization and potential int overflow issues.

### DIFF
--- a/src/wh_server_keystore.c
+++ b/src/wh_server_keystore.c
@@ -551,8 +551,8 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
 
     switch (action) {
         case WH_KEY_CACHE: {
-            whMessageKeystore_CacheRequest  req;
-            whMessageKeystore_CacheResponse resp;
+            whMessageKeystore_CacheRequest  req = {0};
+            whMessageKeystore_CacheResponse resp = {0};
 
             /* translate request */
             (void)wh_MessageKeystore_TranslateCacheRequest(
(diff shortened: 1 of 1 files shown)


src/wh_server_counter.c
```diff
@@ -93,8 +94,8 @@ int wh_Server_HandleCounter(whServerContext* server, uint16_t magic,
             /* read the counter, stored in the metadata label */
             ret     = wh_Nvm_GetMetadata(server->nvm,
                                          WH_MAKE_KEYID(WH_KEYTYPE_COUNTER,
-                                                       server->comm->client_id,
-                                                       req.counterId),
+                                                       (uint16_t)server->comm->client_id,
+                                                       (uint16_t)req.counterId),
                                          meta);
             resp.rc = ret;
```
bigbrett, line 100: I see the following issue on coverity
> 
> CID 572951: (#3 of 4): Overflowed constant (INTEGER_OVERFLOW)
> 8. overflow_const: Expression resp.rc, where ret is known to be equal to -2000, overflows the type of resp.rc, which is type uint32_t

Not sure why we are casting the args to WH_MAKE_KEYID...imagine the proper fix for this is changing `resp.rc` to be an `int32_t`. The error flagged is due to us storing an `int32_t` in a `uint32_t`, which causes an implicit conversion making the value a very large positive number. This happens to be innocuous since we treat it like a signed integer on the client side when we use it, but we might as well just properly type the return value.

bigbrett (changes requested):

src/wh_server_counter.c
bigbrett: @billphipps FYI I'm seeing a bunch of `uint32_t` return values in the keystore and counter message modules. These were all picked up by coverity as integer overflows. Going to have @jackctj117  fix them in the message structs - lmk if you don't think that is a good idea

billphipps: properly typing signedness is a good idea.  I thought we had updated a bunch of the crypto messages already in this fashion, so it's good that we'll get all of the messages updated.  Carry on!

src/wh_server_keystore.c
```diff
@@ -448,7 +448,8 @@ int wh_Server_KeystoreReadKey(whServerContext* server, whKeyId keyId,
     }
     /* cache key if free slot, will only kick out other commited keys */
     if (ret == 0 && out != NULL) {
-        wh_Server_KeystoreCacheKey(server, meta, out);
+        /* Best-effort caching: failure does not affect read result */
```
bigbrett, line 451: pointless AI comment, please remove

bigbrett (changes requested):

PR #176, opened 2025-09-18
Reject unformatted PRs

bigbrett:
Adds a GH action that enforces code formatting on a PR (only on the diff against the PR target branch, not on all files).

The action runs `git-clang-format` against the target branch (usually `main`) to see if there are any formatting changes that would be made (indicating the PR author didn't run the command locally). If there are changes, the action fails, prompting the PR author to run `git-clang-format` locally and commit the changes.


billphipps (approved):
Awesome!

PR #174, opened 2025-09-17
Fixes scan-build issues 

miyazakh:
Fixes issues on scan-build LLVM-19

```
scan-build --status-bugs make -C test/ 
scan-build --status-bugs make -C benchmark/ 
scan-build --status-bugs make -C examples/ 
scan-build --status-bugs make -C tools/whnvmtool/
```

There is one issue, which is `warning: Value stored to 'ret' is never read` in `wolfcrypt\src\random.c` against `tools/whnvmtool/`, remaining. That will be separately fixed.

Github action addition for changes in the future will be submitted as separate PR.

PR #190, opened 2025-09-25
Add GH action for scan build

miyazakh:
Add scan target to Makefile


miyazakh:
The GH action for `scan-build` should be OK once [PR#174](https://github.com/wolfSSL/wolfHSM/pull/174) is merged.

PR #199, opened 2025-10-02
dma: sha512: avoid passing DMA address to the callback

rizlik:
wolfcrypt current SHA512 implementation copy into the destination
address (in the client address space) outside of the crypto callback.
This is a problem for DMA engine that do not work well with the DMA
mapping callbacks.

Use a local buffer to use with wolfcrypt and copy the data back to the
client using `whServerDma_CopyToClient` after the operation.


bigbrett:
@rizlik yep if there is no other way then we gotta do it but we should make absolutely sure  that is the case first

PR #224, opened 2025-10-22
Global keystore

bigbrett:
Adds support for global keys, enabling cryptographic keys to be shared across multiple wolfHSM clients. When a key is marked as global, it becomes accessible to all clients rather than being isolated to the client that cached it.  The global keycache is currently located in the NVM context, as this is the only global state accessible from every server struct instance, and also serves to explicitly indicate the NVM (if there were to be multiple instances) that should be used as the backing store.

Also refactors wrapped keys to use a similar signaling mechanism to the server, such that they can be stored as a different keytype, enabling them to have the same client-facing IDs as regular keys, meaning the feature is able to coexist with dynamic ID generation for non-wrapped keys.
                                                                                                                             
## Feature Additions
- Refactors key cache into its own structure that can be instantiated either locally (in a server context) or globally (in the NVM context)
- Adds new global keycache to NVM struct and new keyId translation macros that handle managing client's global indicator flag
- Refactors crypto and keystore layers to properly handle global keys
- Refactors wrapped key support to use an indicator flag in the keyId, enabling separate keyId space
- Adds new multi-client test harness, currently supporting only sequential operations, and adds global key tests to it
- Adds default keyId for tests that don't care about it

## Details 

  - Global keys stored in globalCache in NVM context (shared across all servers)                                                                                                                                                                                                                                              
  - Local keys remain in per-client localCache in server context                                                                                                                                                                                                                                                              
  - Global keys are indicated to the server by the client setting the `WH_KEYID_GLOBAL` bit in the keyId (bit 8) when caching/using global keys (e.g. a client specifying keyId of `0x1005` indicates global key 5, whereas `0x0005` indicates client-local key 5)
  - Server strips the flag and associates key with USER=`WH_KEYUSER_GLOBAL==0`  encoding via WH_TRANSLATE_CLIENT_KEYID macro
- Cache routing via _GetCacheContext() helper allows the server to automatically select the appropriate cache based on the updated user field 
- Refactors wrapped keys to use the same scheme, but instead of storing the key to the global user, it stores it as different keytype.

```                                                                                                                                                                                   
  Client API:                                                                                                                                                                                                                                                                                                                 
  // Mark key as global                                                                                                                                                                                                                                                                                                       
  whKeyId globalKey = WH_MAKE_KEYID_GLOBAL(5);                                                                                                                                                                                                                                                                                
  wh_Client_KeyCache(client, globalKey, data, len, ...);                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                              
  // Regular local key (not shared)                                                                                                                                                                                                                                                                                           
  whKeyId localKey = 10;                                                                                                                                                                                                                                                                                                      
  wh_Client_KeyCache(client, localKey, data, len, ...);
```

### DIFF
--- a/src/wh_keyid.c
+++ b/src/wh_keyid.c
@@ -24,22 +24,22 @@
 
 #include "wolfhsm/wh_keyid.h"
 
-whKeyId wh_KeyId_TranslateClient(uint16_t type, uint16_t clientId,
-                                 whKeyId reqId)
+whKeyId wh_KeyId_TranslateFromClient(uint16_t type, uint16_t clientId,
+                                     whKeyId reqId)
 {
     uint16_t user = clientId;
     whKeyId  id   = reqId & WH_KEYID_MASK;
 
 #ifdef WOLFHSM_CFG_GLOBAL_KEYS
-    /* Check for global flag (bit 8: 0x0100) */
-    if ((reqId & 0x0100) != 0) {
+    /* Convert global flag to USER=0 */
+    if ((reqId & WH_KEYID_CLIENT_GLOBAL_FLAG) != 0) {
         user = WH_KEYUSER_GLOBAL;
     }
 #endif
 
 #ifdef WOLFHSM_CFG_KEYWRAP
-    /* Check for wrapped flag (bit 9: 0x0200) */
-    if ((reqId & 0x0200) != 0) {
+    /* Convert wrapped flag to TYPE=WH_KETYPE_WRAPPED */
+    if ((reqId & WH_KEYID_CLIENT_WRAPPED_FLAG) != 0) {
         type = WH_KEYTYPE_WRAPPED;
     }
 #endif
@@ -47,21 +47,21 @@ whKeyId wh_KeyId_TranslateClient(uint16_t type, uint16_t clientId,
     return WH_MAKE_KEYID(type, user, id);
 }
 
-whKeyId wh_KeyId_ToClient(whKeyId serverId)
+whKeyId wh_KeyId_TranslateToClient(whKeyId serverId)
 {
     whKeyId clientId = WH_KEYID_ID(serverId);
 
 #ifdef WOLFHSM_CFG_GLOBAL_KEYS
-    /* Convert USER=0 to global flag (bit 8: 0x0100) */
+    /* Convert USER=0 to global flag */
     if (WH_KEYID_USER(serverId) == WH_KEYUSER_GLOBAL) {
-        clientId |= 0x0100; /* WH_CLIENT_KEYID_GLOBAL_FLAG */
+        clientId |= WH_KEYID_CLIENT_GLOBAL_FLAG;
     }
 #endif
 
 #ifdef WOLFHSM_CFG_KEYWRAP
-    /* Convert TYPE=WRAPPED to wrapped flag (bit 9: 0x0200) */
+    /* Convert TYPE=WRAPPED to wrapped flag */
     if (WH_KEYID_TYPE(serverId) == WH_KEYTYPE_WRAPPED) {
-        clientId |= 0x0200; /* WH_CLIENT_KEYID_WRAPPED_FLAG */
+        clientId |= WH_KEYID_CLIENT_WRAPPED_FLAG;
     }
 #endif
 
--- a/src/wh_server_keystore.c
+++ b/src/wh_server_keystore.c
@@ -171,7 +171,7 @@ static int _GetKeyCacheSlot(whKeyCacheContext* ctx, uint16_t keySz,
 
         /* Zero slot and return pointers */
         if (foundIndex >= 0) {
-            memset(&ctx->cache[foundIndex], 0, sizeof(whServerCacheSlot));
+            memset(&ctx->cache[foundIndex], 0, sizeof(whCacheSlot));
             *outBuf  = ctx->cache[foundIndex].buffer;
             *outMeta = ctx->cache[foundIndex].meta;
         }
@@ -197,7 +197,7 @@ static int _GetKeyCacheSlot(whKeyCacheContext* ctx, uint16_t keySz,
 
         /* Zero slot and return pointers */
         if (foundIndex >= 0) {
-            memset(&ctx->bigCache[foundIndex], 0, sizeof(whServerBigCacheSlot));
+            memset(&ctx->bigCache[foundIndex], 0, sizeof(whBigCacheSlot));
             *outBuf  = ctx->bigCache[foundIndex].buffer;
             *outMeta = ctx->bigCache[foundIndex].meta;
         }
@@ -485,26 +485,6 @@ static int _ExistsInCache(whServerContext* server, whKeyId keyId)
 }
 #endif /* WOLFHSM_CFG_KEYWRAP */
 
-#ifdef WOLFHSM_CFG_KEYWRAP
-int wh_Server_KeystoreIsWrappedKey(whServerContext* server, whKeyId keyId,
-                                   int* outIsWrapped)
-{
-    int isWrapped;
-
-    if (server == NULL || WH_KEYID_ISERASED(keyId)) {
-        return WH_ERROR_BADARGS;
-    }
-
-    (void)server;
-    isWrapped = (WH_KEYID_TYPE(keyId) == WH_KEYTYPE_WRAPPED);
-    if (outIsWrapped != NULL) {
-        *outIsWrapped = isWrapped;
-    }
-
-    return WH_ERROR_OK;
-}
-#endif /* WOLFHSM_CFG_KEYWRAP */
-
 /* try to put the specified key into cache if it isn't already, return pointers
  * to meta and the cached data*/
 int wh_Server_KeystoreFreshenKey(whServerContext* server, whKeyId keyId,
@@ -742,8 +722,8 @@ static int _AesGcmWrapKey(whServerContext* server, whKeyId serverKeyId,
     /* Get the server side key */
     ret = wh_Server_KeystoreReadKey(
         server,
-        wh_KeyId_TranslateClient(WH_KEYTYPE_CRYPTO, server->comm->client_id,
-                                 serverKeyId),
+        wh_KeyId_TranslateFromClient(WH_KEYTYPE_CRYPTO, server->comm->client_id,
+                                     serverKeyId),
         NULL, serverKey, &serverKeySz);
     if (ret != WH_ERROR_OK) {
         return ret;
@@ -815,8 +795,8 @@ static int _AesGcmUnwrapKey(whServerContext* server, uint16_t serverKeyId,
     /* Get the server side key */
     ret = wh_Server_KeystoreReadKey(
         server,
-        wh_KeyId_TranslateClient(WH_KEYTYPE_CRYPTO, server->comm->client_id,
-                                 serverKeyId),
+        wh_KeyId_TranslateFromClient(WH_KEYTYPE_CRYPTO, server->comm->client_id,
+                                     serverKeyId),
         NULL, serverKey, &serverKeySz);
     if (ret != WH_ERROR_OK) {
         return ret;
@@ -1115,7 +1095,7 @@ _HandleUnwrapAndCacheKeyRequest(whServerContext*                         server,
     }
 
     /* Store the assigned key ID in the response, preserving client flags */
-    resp->keyId = wh_KeyId_ToClient(metadata.id);
+    resp->keyId = wh_KeyId_TranslateToClient(metadata.id);
 
     /* Cache the key */
     return wh_Server_KeystoreCacheKey(server, &metadata, key);
@@ -1153,7 +1133,7 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
             in = (uint8_t*)req_packet + sizeof(req);
 
             /* set the metadata fields */
-            meta->id = wh_KeyId_TranslateClient(
+            meta->id = wh_KeyId_TranslateFromClient(
                 WH_KEYTYPE_CRYPTO, server->comm->client_id, req.id);
             meta->access = WH_NVM_ACCESS_ANY;
             meta->flags  = req.flags;
@@ -1181,7 +1161,7 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
             }
             if (ret == WH_ERROR_OK) {
                 /* Translate server keyId back to client format with flags */
-                resp.id = wh_KeyId_ToClient(meta->id);
+                resp.id = wh_KeyId_TranslateToClient(meta->id);
 
                 (void)wh_MessageKeystore_TranslateCacheResponse(
                     magic, &resp,
@@ -1202,7 +1182,7 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
                 magic, (whMessageKeystore_CacheDmaRequest*)req_packet, &req);
 
             /* set the metadata fields */
-            meta->id = wh_KeyId_TranslateClient(
+            meta->id = wh_KeyId_TranslateFromClient(
                 WH_KEYTYPE_CRYPTO, server->comm->client_id, req.id);
             meta->access = WH_NVM_ACCESS_ANY;
             meta->flags  = req.flags;
@@ -1236,7 +1216,7 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
             }
 
             /* Translate server keyId back to client format with flags */
-            resp.id = wh_KeyId_ToClient(meta->id);
+            resp.id = wh_KeyId_TranslateToClient(meta->id);
 
             (void)wh_MessageKeystore_TranslateCacheDmaResponse(
                 magic, &resp, (whMessageKeystore_CacheDmaResponse*)resp_packet);
@@ -1254,8 +1234,8 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
 
             ret = wh_Server_KeystoreExportKeyDma(
                 server,
-                wh_KeyId_TranslateClient(WH_KEYTYPE_CRYPTO,
-                                         server->comm->client_id, req.id),
+                wh_KeyId_TranslateFromClient(WH_KEYTYPE_CRYPTO,
+                                             server->comm->client_id, req.id),
                 req.key.addr, req.key.sz, meta);
             resp.rc = ret;
             /* propagate bad address to client if DMA operation failed */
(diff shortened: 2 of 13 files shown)


wolfhsm/wh_server_cache.h
```diff
@@ -0,0 +1,63 @@
...
+
+/** Server cache slot structures */
+typedef struct whServerCacheSlot {
+    uint8_t       commited;
+    whNvmMetadata meta[1];
+    uint8_t       buffer[WOLFHSM_CFG_SERVER_KEYCACHE_BUFSIZE];
+} whServerCacheSlot;
+
+typedef struct whServerBigCacheSlot {
+    uint8_t       commited;
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_server_keystore.c
```diff
@@ -494,48 +578,55 @@ int wh_Server_KeystoreReadKey(whServerContext* server, whKeyId keyId,
...
     }
 
-    ret = _FindInCache(server, keyId, &index, &big, &slotBuf, &slotMeta);
+    /* Get the appropriate cache context for this key */
+    ctx = _GetCacheContext(server, keyId);
+
+    /* Find the key in cache.
+     * Note: We call _FindInKeyCache() directly (not _FindInCache wrapper)
+     * because we already obtained the correct cache context above. Using
+     * the wrapper would redundantly call _GetCacheContext() again. */
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_nvm.c
```diff
@@ -46,6 +46,11 @@ int wh_Nvm_Init(whNvmContext* context, const whNvmConfig *config)
     context->cb = config->cb;
     context->context = config->context;
 
+#if !defined(WOLFHSM_CFG_NO_CRYPTO) && defined(WOLFHSM_CFG_GLOBAL_KEYS)
+    /* Initialize the global key cache */
+    memset(&context->globalCache, 0, sizeof(context->globalCache));
+#endif
+
```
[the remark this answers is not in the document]
bigbrett: good call. Fixed.

src/wh_server_keystore.c
```diff
@@ -895,11 +984,54 @@ _HandleUnwrapAndCacheKeyRequest(whServerContext*                         server,
...
+    }
+    else {
+        /* Client didn't specify an ID (wrapped with id=ERASED) - generate a
+         * unique local key ID. Note: We always generate a LOCAL key here
+         * because the metadata came from the wrapped blob, not from the client
+         * request, so there's no way for the client to signal "make this
+         * global" via the wrapped metadata. */
+        metadata.id =
+            WH_MAKE_KEYID(WH_KEYTYPE_CRYPTO, server->comm->client_id, 0);
+        ret = wh_Server_KeystoreGetUniqueId(server, &metadata.id);
```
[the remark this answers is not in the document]
bigbrett: why would it matter whether or not it is stored in NVM accessible to the server? 

Also, the line you highlighted is explictly not the global case, so the server does actually manage the keyIds in this case.

But for the global case, key collision is a very real concern, and I'm not sure what the best way to address it for wrapped keys is other than placing the burden of properly coordinating global keyIds across multiple clients entirely on the user. Since the global keyIds would be set by the client either offline or at "wrap" time, it is up to them to ensure they have a registry of global keys in the system. I imagine 99% of real-world utilization of wrapped keys deals with offline provisioning anyway so it should be easy enough to prevent against. I'm not sure what a server should do other than error if it unwraps two global keys, both with the same keyId.

examples/posix/wh_posix_server/wh_posix_server.c
```diff
@@ -118,6 +118,15 @@ static int loadAndStoreKeys(whServerContext* server, whKeyId* outKeyId,
     return ret;
 }
 
+static int _InitDemoServer(whServerContext* server, whServerConfig* config)
```
bigbrett: this is no longer needed after we abandoned the registry idea. Revert.

wolfhsm/wh_common.h
```diff
@@ -51,6 +51,20 @@ typedef uint16_t whKeyId;
...
+/* Provide a portable inline macro for headers consumed under C90 */
+#ifndef WH_INLINE
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+#define WH_INLINE static inline
+#elif defined(__GNUC__) || defined(__clang__)
+#define WH_INLINE static __inline__
+#else
+#define WH_INLINE static
+#endif
+#endif
```
bigbrett: remove this

wolfhsm/wh_common.h
```diff
@@ -62,12 +76,54 @@ typedef uint16_t whKeyId;
 #define WH_KEYID_ID(_kid)   (((_kid) & WH_KEYID_MASK) >> WH_KEYID_SHIFT)
 
 #define WH_KEYID_ISERASED(_kid) (WH_KEYID_ID(_kid) == WH_KEYID_ERASED)
+#define WH_KEYID_ISWRAPPED(_kid) (((_kid) & WH_KEYID_WRAPPED) != 0)
+
+#ifdef WOLFHSM_CFG_GLOBAL_KEYS
+/* Global key flag - client-to-server signal only, NOT stored in keyId.
+ * This flag is stripped before storage and translated to USER=0 encoding */
+#define WH_KEYID_GLOBAL ((whKeyId)0x0100) /* Bit 8: temporary flag for client requests */
```
bigbrett: macros like this don't need to be conditionally defined, only helper macros who's definition changes depending on the configuration

wolfhsm/wh_common.h
```diff
@@ -62,12 +76,54 @@ typedef uint16_t whKeyId;
...
 
 /* Key Types */
 #define WH_KEYTYPE_NVM      0x0     /* Ordinary NvmId.  Not a key */
 #define WH_KEYTYPE_CRYPTO   0x1     /* Key for Crypto operations */
 #define WH_KEYTYPE_SHE      0x2     /* SKE keys are AES or CMAC binary arrays */
 #define WH_KEYTYPE_COUNTER  0x3     /* Monotonic counter */
+#define WH_KEYTYPE_WRAPPED  0x4     /* Wrapped key metadata */
+
+/** Translate client keyId (with flags) to server keyId encoding */
+WH_INLINE whKeyId WH_TRANSLATE_CLIENT_KEYID(uint16_t type, uint16_t clientId,
```
bigbrett: just make normal function

wolfhsm/wh_common.h
```diff
@@ -62,12 +76,54 @@ typedef uint16_t whKeyId;
...
+#ifdef WOLFHSM_CFG_GLOBAL_KEYS
+    if ((reqId & WH_KEYID_GLOBAL) != 0) {
+        user = WH_KEYUSER_GLOBAL;
+    }
+#endif
+
+    if ((reqId & WH_KEYID_WRAPPED) != 0) {
+        type = WH_KEYTYPE_WRAPPED;
+    }
+
```
bigbrett: gate on `WOLFHSM_CFG_KEYWRAP`

wolfhsm/wh_nvm.h
```diff
@@ -44,6 +44,10 @@
 
 #include "wolfhsm/wh_common.h"  /* For whNvm types */
 
+#if !defined(WOLFHSM_CFG_NO_CRYPTO) && defined(WOLFHSM_CFG_GLOBAL_KEYS)
+#include "wolfhsm/wh_server_cache.h" /* For whKeyCacheContext */
```
bigbrett: dont gate include

wolfhsm/wh_server_keystore.h
```diff
@@ -186,4 +189,18 @@ int wh_Server_KeystoreCacheKeyDma(whServerContext* server, whNvmMetadata* meta,
 int wh_Server_KeystoreExportKeyDma(whServerContext* server, whKeyId keyId,
                                    uint64_t keyAddr, uint64_t keySz,
                                    whNvmMetadata* outMeta);
+
+#ifdef WOLFHSM_CFG_KEYWRAP
```
bigbrett: don't gate definition in header

AlexLanzano (changes requested):
Whoops meant to approve

AlexLanzano (approved):

wolfhsm/wh_server_cache.h
```diff
@@ -0,0 +1,63 @@
...
+
+/* Pick up compile-time configuration */
+#include "wolfhsm/wh_settings.h"
+#include "wolfhsm/wh_common.h"
+
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+
+/** Server cache slot structures */
+typedef struct whServerCacheSlot {
+    uint8_t       committed;
```
billphipps:
```suggestion
    uint32_t       committed:1;
```
Consider allocating a bitfield here or a larger variable to avoid internal structures padding.

wolfhsm/wh_server_cache.h
```diff
@@ -0,0 +1,63 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with wolfHSM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * wolfhsm/wh_server_cache.h
```
billphipps: The nvm object is not specific to the server, nor is the idea of key cache.  Any chance we can remove "server" from the name?

wolfhsm/wh_keyid.h
```diff
@@ -0,0 +1,116 @@
...
+ * - 0x0200 (bit 9): WH_CLIENT_KEYID_WRAPPED_FLAG → TYPE = WH_KEYTYPE_WRAPPED
+ *
+ * @param type Key type to use as the TYPE field. Input value is ignored and
+ *  WH_KEYTYPE_WRAPPED is used if the input clientId has the
+ *  WH_CLIENT_KEYID_WRAPPED flag set.
+ * @param clientId Client identifier to use as USER field
+ * @param reqId Requested keyId from client (may include flags)
+ * @return Server-internal keyId with TYPE, USER, and ID fields properly set.
+ */
+whKeyId wh_KeyId_TranslateClient(uint16_t type, uint16_t clientId,
```
billphipps: Perhaps these functions should be in wh_server.h, similar to how the client keyid functions/macros are in wh_client.h?

wolfhsm/wh_common.h
```diff
@@ -24,54 +24,13 @@
 #ifndef WOLFHSM_WH_COMMON_H_
 #define WOLFHSM_WH_COMMON_H_
 
+#include <stdint.h>
+
 /* Pick up compile-time configuration */
 #include "wolfhsm/wh_settings.h"
```
billphipps:
```suggestion
/* Pick up compile-time configuration */
#include "wolfhsm/wh_settings.h"

#include <stdint.h>
```
Other library files have wh_settings.h before anything else, just in case.

wolfhsm/wh_client.h
```diff
@@ -2493,4 +2493,102 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
...
+/* Bit 8: Client-to-server signal for global key (shared across all clients) */
+#define WH_CLIENT_KEYID_GLOBAL_FLAG ((whKeyId)0x0100)
+
+/* Bit 9: Client-to-server signal for wrapped key */
+#define WH_CLIENT_KEYID_WRAPPED_FLAG ((whKeyId)0x0200)
+
+/* Combined mask of all client-facing flags */
+#define WH_CLIENT_KEYID_FLAGS_MASK \
+    (WH_CLIENT_KEYID_GLOBAL_FLAG | WH_CLIENT_KEYID_WRAPPED_FLAG)
+
```
billphipps:
```suggestion

#define WH_CLIENT_KEYID_ID_MASK ((whKeyId)0x00FF)
```
Ensure client ID's are only the bottom bits

wolfhsm/wh_client.h
```diff
@@ -2493,4 +2493,102 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
...
+ *
+ * @param _id The key ID (0-255)
+ * @return keyId with global and wrapped flags set
+ *
+ * Example:
+ *   whKeyId globalWrappedKey = WH_CLIENT_KEYID_MAKE_WRAPPED_GLOBAL(2);
+ *   wh_Client_AesSetKeyId(aes, globalWrappedKey);
+ */
+#define WH_CLIENT_KEYID_MAKE_WRAPPED_GLOBAL(_id) \
+    ((_id) | WH_CLIENT_KEYID_GLOBAL_FLAG | WH_CLIENT_KEYID_WRAPPED_FLAG)
```
billphipps:
```suggestion
    (((_id) & WH_CLIENT_KEYID_ID_MASK) | WH_CLIENT_KEYID_GLOBAL_FLAG | WH_CLIENT_KEYID_WRAPPED_FLAG)
```

wolfhsm/wh_client.h
```diff
@@ -2493,4 +2493,102 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
...
+ * to USER=0 encoding.
+ *
+ * @param _id The key ID (0-255)
+ * @return keyId with global flag set
+ *
+ * Example:
+ *   whKeyId globalKey = WH_CLIENT_KEYID_MAKE_GLOBAL(5);
+ *   wh_Client_KeyCache(client, globalKey, ...);  // Stored as global key
+ */
+#define WH_CLIENT_KEYID_MAKE_GLOBAL(_id) ((_id) | WH_CLIENT_KEYID_GLOBAL_FLAG)
```
billphipps:
```suggestion
#define WH_CLIENT_KEYID_MAKE_GLOBAL(_id) (((_id) & WH_CLIENT_KEYID_ID_MASK) | WH_CLIENT_KEYID_GLOBAL_FLAG)
```

src/wh_server.c
```diff
@@ -188,8 +188,17 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
         wh_MessageComm_TranslateInitRequest(magic,
                 (whMessageCommInitRequest*)req_packet, &req);
 
+#ifdef WOLFHSM_CFG_GLOBAL_KEYS
```
billphipps: Does this need a NO_CRYPTO check as well?

test/wh_test_multiclient.c
```diff
@@ -0,0 +1,1636 @@
...
+
+#include "wh_test_common.h"
+
+/* Test configuration */
+#define FLASH_RAM_SIZE (1024 * 1024)   /* 1MB */
+#define FLASH_SECTOR_SIZE (128 * 1024) /* 128KB */
+#define FLASH_PAGE_SIZE (8)            /* 8B */
+#define BUFFER_SIZE 4096
+
+#ifdef WOLFHSM_CFG_GLOBAL_KEYS
```
billphipps: NO_CRYPTO check here as well?

billphipps (approved):
Excellent addition and awesome testing!  I had a few questions and nits, but nothing that would stop this from being merged.

wolfhsm/wh_client.h
```diff
@@ -2493,4 +2493,102 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
...
+ * KEYTYPE=WH_KEYTYPE_WRAPPED.
+ *
+ * @param _id The key ID (0-255)
+ * @return keyId with wrapped flag set
+ *
+ * Example:
+ *   whKeyId wrappedKey = WH_CLIENT_KEYID_MAKE_WRAPPED(2);
+ *   wh_Client_KeyExportRequest(client, wrappedKey, ...);
+ */
+#define WH_CLIENT_KEYID_MAKE_WRAPPED(_id) ((_id) | WH_CLIENT_KEYID_WRAPPED_FLAG)
```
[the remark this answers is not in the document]
bigbrett: I suppose I just wanted this to add in the global/wrapped bit and not modify anything else, in case we build on top of this scheme to require the client to signal other things. I think masking off other bits here would prohibit future schemes. Same comment on global and wrapped global below.

src/wh_server.c
bigbrett: I went back and forth on this. There are a large number of macros that are dependent on eachother that make it really messy to have `if defined(X) && defined(Y)` everywhere. I think maybe I should just enforce this in one place in `wh_settings.h` instead? e.g. enforce that `NO_CRYPTO` is not defined if `GLOBAL_KEYS` is defined using `#error`? That way the actual code switches in the library are clearer and don't need to take into account a potential tangled web of feature flag dependencies?

test/wh_test_multiclient.c
```diff
@@ -0,0 +1,1636 @@
...
+                              whServerContext* server1,
+                              whClientContext* client2,
+                              whServerContext* server2)
+{
+    int     ret;
+    whKeyId serverKeyId = WH_CLIENT_KEYID_MAKE_GLOBAL(DUMMY_KEYID_1);
+    uint8_t wrapKey[AES_256_KEY_SIZE]  = "GlobalWrapKey123456789012345!";
+    uint8_t plainKey[AES_256_KEY_SIZE] = "PlainKeyToWrap1234567890123!";
+    /* Wrapped key size = IV(12) + TAG(16) + KEYSIZE(32) + metadata */
+#define WRAPPED_KEY_SIZE (12 + 16 + AES_256_KEY_SIZE + sizeof(whNvmMetadata))
```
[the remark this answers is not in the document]
bigbrett: probably...

wolfhsm/wh_keyid.h
bigbrett: these need to actually be global and not server specific, as there are occasionally cases where you need to know the translated format a-priori (think wrapping a key offline) where you  might not want to pull in the server just to do it?

wolfhsm/wh_server_cache.h
bigbrett: that is fair, I can rename

wolfhsm/wh_server_cache.h
bigbrett: I didn't change how it was defined before. Didn't want to go changing code outside the scope of this PR. Padding in this case doesn't matter, since it isnt a wire struct

src/wh_keyid.c
```diff
@@ -0,0 +1,69 @@
...
+
+whKeyId wh_KeyId_TranslateClient(uint16_t type, uint16_t clientId,
+                                 whKeyId reqId)
+{
+    uint16_t user = clientId;
+    whKeyId  id   = reqId & WH_KEYID_MASK;
+
+#ifdef WOLFHSM_CFG_GLOBAL_KEYS
+    /* Check for global flag (bit 8: 0x0100) */
+    if ((reqId & 0x0100) != 0) {
```
AlexLanzano: Make a common #define for `0x0100` so both client and server code can reference it

src/wh_keyid.c
```diff
@@ -0,0 +1,69 @@
...
+        user = WH_KEYUSER_GLOBAL;
+    }
+#endif
+
+#ifdef WOLFHSM_CFG_KEYWRAP
+    /* Check for wrapped flag (bit 9: 0x0200) */
+    if ((reqId & 0x0200) != 0) {
```
AlexLanzano: Make a common #define for `0x0200` so both client and server code can reference it

AlexLanzano (approved):

billphipps (approved):

PR #226, opened 2025-10-23
Implement generic data wrap/unwrap helper functions

AlexLanzano:


test/wh_test_keywrap.c
```diff
@@ -201,6 +201,37 @@ static int _AesGcm_KeyWrap(whClientContext* client, WC_RNG* rng)
     return ret;
 }
 
+static int _AesGcm_DataWrap(whClientContext* client)
```
bigbrett:
```suggestion
static int _TestAesGcm_DataWrap(whClientContext* client)
```

src/wh_client_keywrap.c
```diff
@@ -326,5 +326,160 @@ int wh_Client_KeyUnwrapAndCache(whClientContext*   ctx,
...
+    /* Prepend IV + authTag to the wrapped data buffer */
+    memcpy(wrappedDataOut, iv, sizeof(iv));
+    memcpy(wrappedDataOut + sizeof(iv), authTag, sizeof(authTag));
+
+    wc_AesFree(aes);
+
+    return WH_ERROR_OK;
+}
+
+int wh_Client_AesGcmDataUnwrap(whClientContext* ctx, uint16_t serverKeyId,
```
bigbrett: FIne for now, but it looks like these functions will result in a lot of back and forth between the server and client. I'm wondering if it makes more sense to have them be dedicated messages and just have all the wrapping happen on the server side. Plus we would want DMA versions of these, as what if you want to wrap/unwrap large data in place?

bigbrett (changes requested):

src/wh_client_keywrap.c
AlexLanzano: Yeah I originally wanted to do this but couldn't find a good group to place this new message under. I was thinking of putting it under crypto but the actions use wolfCrypt enums.

I also felt like this didnt fit under keystore like the current key wrap stuff.

Was wondering your thoughts on this.

src/wh_client_keywrap.c
```diff
@@ -326,5 +326,162 @@ int wh_Client_KeyUnwrapAndCache(whClientContext*   ctx,
...
 
+static int wh_Client_AesGcmDataWrap(whClientContext* ctx, uint16_t serverKeyId,
+                                    void* dataIn, uint32_t dataInSz,
+                                    void*    wrappedDataOut,
+                                    uint32_t wrappedDataOutSz)
+{
+    int      ret = 0;
+    Aes      aes[1];
+    uint8_t  authTag[WOLFHSM_KEYWRAP_AES_GCM_TAG_SIZE];
+    uint8_t  iv[WOLFHSM_KEYWRAP_AES_GCM_IV_SIZE];
```
bigbrett: if these are global non-overrideable macros they should be `WH_XXX`. If they are overridable they should be `WOLFHSM_CFG_XXX`. If they are local then just scope the name to the module in a sane fashion.

I think these should probably be public though?

bigbrett (changes requested):

src/wh_client_keywrap.c
```diff
@@ -18,16 +18,17 @@ int wh_Client_KeyWrapRequest(whClientContext*   ctx,
                              whNvmMetadata* metadata)
 {
     uint16_t                       group  = WH_MESSAGE_GROUP_KEY;
-    uint16_t                       action = WH_KEY_WRAP;
-    whMessageKeystore_WrapRequest* req    = NULL;
+    uint16_t                          action = WH_KEY_KEYWRAP;
```
bigbrett: run `git-clang-format` against `main` after updating your local `main` to the latest

src/wh_server_keystore.c
```diff
@@ -1100,6 +1216,132 @@ _HandleUnwrapAndCacheKeyRequest(whServerContext*                         server,
...
+
+#ifndef NO_AES
+#ifdef HAVE_AESGCM
+        case WC_CIPHER_AES_GCM: {
+            uint16_t wrappedDataSz = WOLFHSM_KEYWRAP_AES_GCM_IV_SIZE +
+                                     WOLFHSM_KEYWRAP_AES_GCM_TAG_SIZE +
+                                     req->dataSz;
+
+            /* Check if the response data can fit the wrapped data */
+            if (respDataSz < wrappedDataSz) {
```
bigbrett: we don't want to return this from this function (essentially crashing the server if anything bad happens) but instead need to propagate the failure back to the client. Look at the changes I made to this PR to the regular keywrap functions below (which I did to get it to pass CI)

bigbrett (changes requested):

src/wh_server_keystore.c
```diff
@@ -1100,6 +1216,132 @@ _HandleUnwrapAndCacheKeyRequest(whServerContext*                         server,
...
+    uint8_t* wrappedData;
+    uint8_t  data[WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE];
+
+    if (server == NULL || req == NULL || reqData == NULL || resp == NULL ||
+        respData == NULL || req->dataSz > WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Check if the reqData is big enough to hold the data */
+    if (reqDataSz < req->dataSz) {
```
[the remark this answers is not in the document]
AlexLanzano: Understood! It looks like from the new changes you pushed in there's no action here since the return value is being set to the resp.rc when this  function is called?

bigbrett (approved):

PR #229, opened 2025-10-28
HMAC support

rizlik:
Implement HMAC state caching with a dedicated key type and mirrored lifecycle handling.

  - Add WH_KEYTYPE_HMAC_STATE so cached HMAC contexts are isolated from normal crypto keys, permanently flagged as NONEXPORTABLE/EPHEMERAL, and prevented from being reused or committed to NVM.
  - Streamed HMAC requests now serialize the wolfCrypt Hmac struct into the keystore on each update, reload it for subsequent chunks, and evict it whenever a final/abort/error path runs, with the client clearing its stateId in
    lockstep to avoid dangling handles.
    
    
NOTE: the PR stores the HMAC state inside the sever. Ideally we want to use the new wrapping API and store the state securely on the client instead. This will be an improvement on this PR. My main open questions on that approach are:

1. Where to store the state on the client? 
2. How to prevent replay of the state (if the attack is in scope)

### DIFF
--- a/tools/whnvmtool/test/Makefile
+++ b/tools/whnvmtool/test/Makefile
@@ -25,7 +25,8 @@ WOLFCRYPT_SRC = \
 	$(WOLFSSL_DIR)/wolfcrypt/src/sha256.c \
 	$(WOLFSSL_DIR)/wolfcrypt/src/aes.c \
 	$(WOLFSSL_DIR)/wolfcrypt/src/ecc.c \
-	$(WOLFSSL_DIR)/wolfcrypt/src/cmac.c
+	$(WOLFSSL_DIR)/wolfcrypt/src/cmac.c \
+	$(WOLFSSL_DIR)/wolfcrypt/src/hmac.c
 
 SRC = \
 	$(WOLFHSM_SRC) \


rizlik:
Hi @bigbrett ,

Yeah, I agree both on the bad timing and on the huge limitation on having the state cached on server but I don't think the caveat can work, as I don't see a safe way to cache the state on the client. 

In HMAC just caching the partial hash state of H(K^opad) and H(K'^ipad) is catastrophic as it provide easy offline forgery. 
Having these two partial states allows to forge the HMAC for any message offline. If you have a more advanced partial state, like H(K^opad)  K(K^ipad || m) you can always forge message M' with fixed prefix M. 

It's also stated clearly in the FIPS document:
```
These stored intermediate
values shall be treated and protected in the same manner as secret keys.
```

In CMAC leaking the state is not such as catastrophic but it's a very bad practice, as no intermediate key-based material that's based on a key whose exporting is restricted  should never leave the security boundary (exception for the output of the cryptographic operation, ofc).

The only solution I see right now is to wrap the sensible state on the server, after thoughtful consideration of replay attacks across users. 

We also need a space to save wrapped data along HMAC and CMAC object on the client side.

The reason this was rushed up is that it might become critical on a delivery. 

I'm OK in dropping this but then we probably need to define a clear roadmap to solve the blockers for the wrapping state solution.

PR #230, opened 2025-10-29
Scan-build improvements

bigbrett:
- limits scan-build to only running on unit test suite (no reduction in coverage since all targets wildcard all source files anyway)
- uses `--exclude` flag to scan-build to ignore wolfCrypt and wolfSSL source files from analysis
- adds `-j` flag for parallel make

These changes should dramatically speed up the `scan-build` CI job and prevent issues in wolfCrypt from breaking wolfHSM CI


miyazakh:
Thanks. It seems that `make -j` is the most effective. The `--exclude` option does not seem to work, as shown by the failure when running `make -j SCAN=1 -C tools/whnvmtool/`.

PR #233, opened 2025-11-05
Key usage policies

bigbrett:
This PR adds usage policy flags to cached keys that are set by the client and enforced by the server. Clients can now restrict usage of keys to specific operations (encrypt, decrypt, sign, verify, wrap, derive).

Usage of the wolfCrypt API with local keys (e.g. not referring to keys via ID)  will automatically set the appropriate flags under the hood.

### DIFF
--- a/src/wh_server_keystore.c
+++ b/src/wh_server_keystore.c
@@ -144,8 +144,10 @@ static int _GetKeyCacheSlot(whKeyCacheContext* ctx, uint16_t keySz,
 {
     int foundIndex = -1;
     int i;
+    uint8_t*       slotBuf  = NULL;
+    whNvmMetadata* slotMeta = NULL;
 
-    if (ctx == NULL || outBuf == NULL || outMeta == NULL) {
+    if (ctx == NULL) {
         return WH_ERROR_BADARGS;
     }
 
@@ -169,11 +171,11 @@ static int _GetKeyCacheSlot(whKeyCacheContext* ctx, uint16_t keySz,
             }
         }
 
-        /* Zero slot and return pointers */
+        /* Zero slot and capture pointers */
         if (foundIndex >= 0) {
             memset(&ctx->cache[foundIndex], 0, sizeof(whCacheSlot));
-            *outBuf  = ctx->cache[foundIndex].buffer;
-            *outMeta = ctx->cache[foundIndex].meta;
+            slotBuf  = ctx->cache[foundIndex].buffer;
+            slotMeta = ctx->cache[foundIndex].meta;
         }
     }
     else {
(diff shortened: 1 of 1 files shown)


wolfhsm/wh_error.h
```diff
@@ -43,6 +43,8 @@ enum WH_ERROR_ENUM {
     WH_ERROR_NOHANDLER = -2007,   /* No customcb handler registered */
     WH_ERROR_NOTIMPL   = -2008,   /* Functionality not implemented given the
                                       compile-time configuration */
+    WH_ERROR_USAGE =
```
billphipps: Contrast this with WH_ERROR_ACCESS?  Consider WH_ERROR_KEYUSAGE.

wolfhsm/wh_server_keystore.h
```diff
@@ -190,4 +190,45 @@ int wh_Server_KeystoreExportKeyDma(whServerContext* server, whKeyId keyId,
...
+ *
+ * @param[in] meta          Pointer to key metadata
+ * @param[in] requiredUsage Required usage policy flags (e.g.,
+ *                          WH_NVM_FLAGS_USAGE_ENCRYPT |
+ * WH_NVM_FLAGS_USAGE_DECRYPT)
+ * @return WH_ERROR_OK if the key has all required usage flags set
+ * @return WH_ERROR_USAGE if the key does not have the required flags
+ * @return WH_ERROR_BADARGS if meta is NULL
+ */
+int wh_Server_KeystoreEnforceKeyUsage(const whNvmMetadata* meta,
```
billphipps: Consider moving this function out of keystore (or at least renaming to wh_Metadata_CheckKeyUsage) since this has nothing to do with the server nor keystore itself since it uses only metadata.  The keystore variant below (named xxxFindxxx) could then simply be named wh_Server_KeystoreCheckKeyUsage and would use the metadata checkkeyusage once it found the metadata.

wolfhsm/wh_common.h
```diff
@@ -66,14 +66,49 @@ typedef uint16_t whNvmAccess;
...
+ *
+ * Key usage flags control which cryptographic operations are permitted.
+ * Multiple usage flags can be combined. If no usage flags are set, the key
+ * cannot be used for any operation. Use WH_NVM_FLAGS_USAGE_ANY to allow all
+ * operations.
+ */
+/* Key can be used for encryption */
+#define WH_NVM_FLAGS_USAGE_ENCRYPT ((whNvmFlags)1 << 5)
+/* Key can be used for decryption */
+#define WH_NVM_FLAGS_USAGE_DECRYPT ((whNvmFlags)1 << 6)
```
billphipps: Consider removing this since it overlaps with ENCRYPT.  For symmetric algos, knowledge of the key is sufficient for any operation.  Most streaming modes (GCM, CTR, etc) only use the ENCRYPT direction and the encrypt and decrypt operations are the same.

billphipps (approved):
Looks great!  A few naming questions and some design questions, but these can be punted as we further refine the design.  We may want to isolate all of the checks/usage code into a new file and support callback overrides for the logic.

test/wh_test_keywrap.c
```diff
@@ -91,7 +91,7 @@ static int _AesGcm_TestKeyWrap(whClientContext* client, WC_RNG* rng)
                                                        WH_TEST_AESGCM_KEYID),
             .label = "AES Key Label",
             .len   = WH_TEST_AES_KEYSIZE,
-            .flags = WH_NVM_FLAGS_NONE,
+            .flags = WH_NVM_FLAGS_USAGE_ANY,
```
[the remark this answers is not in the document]
bigbrett: the key isn't actually used for anything so I just used that as a placeholder

wolfhsm/wh_server_keystore.h
bigbrett: Good point on the metadata-only path. For the other function, I wanted to keep "Find" in the name to indicate that it does do an internal lookup (potentially even from NVM). There are lots of code paths that first do a lookup and have access to the metadata already, so calling this out in the name prevents accidental redundant use.

PR #234, opened 2025-11-08
wh_client: fix: cap labelSz and add defensive NULL checks

kaabia:
This PR addresses two related issues in `wh_client.c`:

1. It fixes a potential **buffer overflow** in key cache request functions (`wh_Client_KeyCacheRequest_ex` and `wh_Client_KeyCacheDmaRequest`) by ensuring that the key **label** size is always **capped** at `WH_NVM_LABEL_LEN` before copying it into the request structure and setting the size field (`req->labelSz`).
  
2. It corrects a variable name logic error in `wh_Client_CommInfoResponse` where the output parameters for "big" key cache configuration were incorrectly using the same variable check as the standard key cache configuration.

### DIFF
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -339,10 +339,10 @@ int wh_Client_CommInfoResponse(whClientContext* c,
             if (out_cfg_keycache_bufsize != NULL) {
                 *out_cfg_keycache_bufsize = msg.cfg_server_keycache_bufsize;
             }
-            if (out_cfg_keycache_count != NULL) {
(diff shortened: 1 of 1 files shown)


bigbrett (approved):
The variable name fix was necessary, the original code clearly contained a copy-paste error.

The criticality of the label length fix is overstated, as there is no possibility of buffer overflow, since the server truncates the label length sent by the client or returns error depending on the context. That said the fix is innocuous and it doesn't hurt to check it on the client side.

PR #242, opened 2025-11-14
Use smaller 16 byte message for ECC demo

AlexLanzano:


examples/demo/client/wh_demo_client_crypto.c
```diff
@@ -492,7 +492,7 @@ int wh_DemoClient_CryptoEcc(whClientContext* clientContext)
     WC_RNG     rng[1];
     byte       sharedOne[32];
     byte       sharedTwo[32];
-    const char plainMessage[] = "The quick brown fox jumps over the lazy dog.";
+    const char plainMessage[WC_MAX_DIGEST_SIZE] = "message example";
```
bigbrett: can you make the message match the array length, just for clarity?

AlexLanzano: Now that I think about it. Why don't I leave plainMessage as is and add a variable `hash[WC_MAX_DIGEST_SIZE]` and compute the sha256 hash of plainMessage. Might be a more valid demo of ECC?

examples/demo/client/wh_demo_client_crypto.c
```diff
@@ -492,12 +492,17 @@ int wh_DemoClient_CryptoEcc(whClientContext* clientContext)
...
-    const char plainMessage[] = "The quick brown fox jumps over the lazy dog.";
-    byte       message[sizeof(plainMessage)];
+    const byte plainMessage[] = "The quick brown fox jumps over the lazy dog.";
+    byte       hash[WC_MAX_DIGEST_SIZE];
     byte       signature[128];
 
-    /* Set the message to the test string */
-    strcpy((char*)message, plainMessage);
+    /* Hash the plainMessage using SHA256 for signing and verification */
+    ret = wc_Hash(WC_HASH_TYPE_SHA256, plainMessage, sizeof(plainMessage),
```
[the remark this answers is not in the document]
AlexLanzano: It looks like `WC_HASH_TYPE_SHA256` should be exist even if `NO_SHA256` is defined. wc_Hash will also return an error if we try to do a sha256 with NO_SHA256 defined. So we should be good here.

PR #253, opened 2025-11-21
Logging

bigbrett:
- Introduces generic logging framework consisting of front-end API with back-end callbacks
 - Adds two basic built-in back-end engines: Single threaded ring buffer and POSIX file log
 - Adds global wolfHSM system time macro for things like logging or benchmarks

Unrelated:
- Truncate metadata label when caching keys

### DIFF
--- a/wolfhsm/wh_log_ringbuf.h
+++ b/wolfhsm/wh_log_ringbuf.h
@@ -29,7 +29,7 @@
 #include "wolfhsm/wh_settings.h"
 #include "wolfhsm/wh_log.h"
 
-#include <stdint.h>
+#include <stddef.h>
 
 /* Ring buffer configuration structure */
 typedef struct whLogRingbufConfig_t {


wolfhsm/wh_log_ringbuf.h
```diff
@@ -0,0 +1,72 @@
...
+    void*  buffer;      /* User-supplied buffer */
+    size_t buffer_size; /* Size of buffer in bytes */
+} whLogRingbufConfig;
+
+/* Ring buffer context structure */
+typedef struct whLogRingbufContext_t {
+    whLogEntry* entries;     /* Pointer to user-supplied buffer */
+    size_t      capacity;    /* Number of entries buffer can hold */
+    size_t      head;        /* Next write position (0 to capacity-1) */
+    size_t      count;       /* Current entries (0 to capacity) */
```
billphipps: Consider allowing count to increment freely to understand how many messages have been overwritten.

wolfhsm/wh_log.h
```diff
@@ -0,0 +1,200 @@
...
+#include "wolfhsm/wh_utils.h"
+
+#include <stdint.h>
+#include <stdarg.h>
+#include <string.h>
+#include <stdio.h>
+
+/** Log levels */
+typedef enum {
+    WH_LOG_LEVEL_INFO     = 0, /* Informational message */
```
billphipps: WH_LOG_LEVEL_DEBUG?

wolfhsm/wh_server.h
```diff
@@ -164,6 +165,9 @@ typedef struct whServerConfig_t {
 #ifdef WOLFHSM_CFG_DMA
     whServerDmaConfig* dmaConfig;
 #endif /* WOLFHSM_CFG_DMA */
+#ifdef WOLFHSM_CFG_LOGGING
+    whLogConfig* logConfig;
+#endif /* WOLFHSM_CFG_LOGGING */
```
billphipps: Consider always having the logging in the source, but handling default NULL pointers when being invoked when not needed

src/wh_log.c
```diff
@@ -0,0 +1,221 @@
...
+ */
+
+/* Pick up compile-time configuration */
+#include "wolfhsm/wh_settings.h"
+
+#include <stdint.h>
+#include <stddef.h>
+#include <stdarg.h>
+#include <string.h>
+#include <stdio.h>
```
billphipps: Maybe formatted messages is a CFG option so we can isolate stdio.h requirements?

wolfhsm/wh_log.h
```diff
@@ -0,0 +1,200 @@
...
+        wh_Log_InternalFormat((ctx), (lvl), __FILE__, __func__, __LINE__, \
+                              (fmt), ##__VA_ARGS__);                      \
+    } while (0)
+#else
+/* CCRH workaround for empty __VA_ARGS__ */
+#define WH_LOG_F(ctx, lvl, ...) WH_LOG_F2((ctx), (lvl), __VA_ARGS__, "")
+#define WH_LOG_F2(ctx, lvl, fmt, ...)                                     \
+    do {                                                                  \
+        wh_Log_InternalFormat((ctx), (lvl), __FILE__, __func__, __LINE__, \
+                              (fmt), ##__VA_ARGS__);                      \
```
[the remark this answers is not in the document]
bigbrett: Yes this is a GCC thing but has been adopted by pretty much every compiler, including clang and hightec.

test/config/wolfhsm_cfg.h
```diff
@@ -59,6 +65,7 @@
 #define WOLFHSM_CFG_CANCEL_API
 #endif
 
+/* Test log-based NVM flash backend */
 #define WOLFHSM_CFG_SERVER_NVM_FLASH_LOG
```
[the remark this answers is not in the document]
bigbrett: agree

wolfhsm/wh_log.h
```diff
@@ -0,0 +1,200 @@
...
+} whLogConfig;
+
+/** Frontend API */
+int wh_Log_Init(whLogContext* ctx, const whLogConfig* config);
+int wh_Log_Cleanup(whLogContext* ctx);
+int wh_Log_AddEntry(whLogContext* ctx, const whLogEntry* entry);
+int wh_Log_Export(whLogContext* ctx, void* export_arg);
+int wh_Log_Iterate(whLogContext* ctx, whLogIterateCb iterate_cb,
+                   void* iterate_arg);
+int wh_Log_Clear(whLogContext* ctx);
```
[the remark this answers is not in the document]
bigbrett: fixed

wolfhsm/wh_log.h
```diff
@@ -0,0 +1,200 @@
...
+
+/*
+ * Internal logging helpers. Both silently truncate to WOLFHSM_CFG_LOG_MSG_MAX -
+ * 1 and always null-terminate the stored message.
+ */
+
+void wh_Log_InternalSubmit(whLogContext* ctx, whLogLevel level,
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_log.c
bigbrett: keep in mind the logging layer is largely meant to be used internally. So we either want to capitalize on formatted logs or not. We would have things like `WH_LOG("%s event occured") throughout our code base, so not sure how we would conditionally support something like that. Seems weird to have a bunch of conditonally compilled log points in our code?

billphipps: Gotcha and agree that this should be human readable and useable.  Can we at least add a comment that we support "limited printf" formatting?  We shouldn't need to leverage any formatting beyond what we support in our string.c.   No floating point, no width specifiers, no variable length.  Just simple formatting.

src/wh_log_ringbuf.c
billphipps: prolly ought to add: && (context->capacity != 0) to avoid a div/0 below

bigbrett: @billphipps OK just pushed a fix - so instead of checking for `capacity==0` everywhere, I instead ensure that capacity is greater than zero in the init funciton (where capacity is computed) and then all the other operations with the modulo logic are conditional on the context being initialized. 

Capacity==0 really is an invalid state. 

Is this okay?

billphipps: Yup!

wolfhsm/wh_log_ringbuf.h
```diff
@@ -0,0 +1,71 @@
...
+ * and not thread-safe. Overwrites oldest entries when full.
+ */
+
+#ifndef WOLFHSM_WH_LOG_RINGBUF_H_
+#define WOLFHSM_WH_LOG_RINGBUF_H_
+
+#include "wolfhsm/wh_settings.h"
+#include "wolfhsm/wh_log.h"
+
+#include <stdint.h>
```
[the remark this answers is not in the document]
bigbrett: fixed

PR #270, opened 2026-01-16
authentication manager feature addition

JacobBarthelmeh:
The authentication manager feature adds support for a user login and checking a users permissions for performing a group+action. The API was designed with PKCS11 in mind.

Some things of note:

- I added a callback function framework for checking authorization of key use based on key ID and user permissions but did not tie in that check yet. I would like to tie that in later when/if needed. This currently checks for authorization of user for a group/action that they can do. Which ties a user ID to crypto actions done.
- The user list in port/posix/posix_auth.c is a simple list not yet in NVM. This initial simplicity is deliberate.
- There is a TODO listed for logging of authentication events. Login failures, success, crypto actions should have logging additions in the future.

### DIFF
--- a/src/wh_auth.c
+++ b/src/wh_auth.c
@@ -1,0 +1,454 @@
+/*
+ * Copyright (C) 2025 wolfSSL Inc.
+ *
+ * This file is part of wolfHSM.
+ *
+ * wolfHSM is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfHSM is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with wolfHSM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * src/wh_auth.c
+ *
+ * Core Auth Manager implementation. Provides wrapper functions that delegate
+ * to the configured auth backend callbacks.
+ *
+ * - Verifies PINs/credentials
+ * - Calls to implemented callbacks for managing users and permissions
+ * - Authorization decisions are routed through the implemented callbacks
+ *
+ * The Auth Manager is agnostic to the transport used and manages authentication
+ * of a session. It can take a PIN or certificate for verification. An
+ * authenticated session is separate from a comm connection and sits on top of
+ * a comm connection. Allowing for multiple authenticated sessions opened and
+ * closed multiple times through out the span of a single comm connection
+ * established. Currently there is a restriction of one user logged in at a time
+ * per comm connection.
+ */
+
+/* Pick up compile-time configuration */
+#include "wolfhsm/wh_settings.h"
+
+#ifdef WOLFHSM_CFG_ENABLE_AUTHENTICATION
+
+#include <stdint.h>
+#include <stddef.h>
+#include <string.h>
+
+#include "wolfhsm/wh_common.h"
+#include "wolfhsm/wh_error.h"
+
+#include "wolfhsm/wh_auth.h"
+#include "wolfhsm/wh_message_auth.h"
+
+
+int wh_Auth_Init(whAuthContext* context, const whAuthConfig* config)
+{
+    int rc = WH_ERROR_OK;
+
+    if ((context == NULL) || (config == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    context->cb      = config->cb;
+    context->context = config->context;
+    memset(&context->user, 0, sizeof(whAuthUser));
+
+#ifdef WOLFHSM_CFG_THREADSAFE
+    /* Initialize the lock for thread-safe auth operations */
+    rc = wh_Lock_Init(&context->lock, config->lockConfig);
+    if (rc != WH_ERROR_OK) {
+        context->cb      = NULL;
+        context->context = NULL;
+        return rc;
+    }
+#endif /* WOLFHSM_CFG_THREADSAFE */
+
+    if (context->cb != NULL && context->cb->Init != NULL) {
+        rc = context->cb->Init(context->context, config->config);
+        if (rc != WH_ERROR_OK) {
+#ifdef WOLFHSM_CFG_THREADSAFE
+            (void)wh_Lock_Cleanup(&context->lock);
+#endif
+            context->cb      = NULL;
+            context->context = NULL;
+        }
+    }
+
+    return rc;
+}
+
+
+int wh_Auth_Cleanup(whAuthContext* context)
+{
+    int rc = WH_ERROR_OK;
+
+    if ((context == NULL) || (context->cb == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    if (context->cb->Cleanup == NULL) {
+        return WH_ERROR_ABORTED;
+    }
+
+    rc = context->cb->Cleanup(context->context);
+
+#ifdef WOLFHSM_CFG_THREADSAFE
+    /* Cleanup the lock for thread-safe auth operations */
+    (void)wh_Lock_Cleanup(&context->lock);
+#endif /* WOLFHSM_CFG_THREADSAFE */
+
+    return rc;
+}
+
+
+/* Returns a wolfHSM error code: WH_ERROR_OK (0) if the call completed
+ * successfully (regardless of authentication result), or a negative error code
+ * if a fatal error occurred. The result of the login attempt is stored in
+ * loggedIn: 1 for successful authentication, 0 for failed authentication. */
+int wh_Auth_Login(whAuthContext* context, uint8_t client_id,
+                  whAuthMethod method, const char* username,
+                  const void* auth_data, uint16_t auth_data_len, int* loggedIn)
+{
+    int               rc;
+    whUserId          out_user_id;
+    whAuthPermissions out_permissions;
+
+    if (loggedIn == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+    *loggedIn = 0;
+
+    if ((context == NULL) || (context->cb == NULL) ||
+        (context->cb->Login == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    rc = WH_AUTH_LOCK(context);
+    if (rc == WH_ERROR_OK) {
+        /* allowing only one user logged in to an open connection at a time */
+        if (context->user.user_id != WH_USER_ID_INVALID) {
+            *loggedIn = 0;
+            rc        = WH_ERROR_OK; /* login attempt happened but failed */
+        }
+        else {
+            rc = context->cb->Login(context->context, client_id, method,
+                                    username, auth_data, auth_data_len,
+                                    &out_user_id, &out_permissions, loggedIn);
+            if (rc == WH_ERROR_OK && *loggedIn) {
+                context->user.user_id     = out_user_id;
+                context->user.permissions = out_permissions;
+                context->user.is_active   = true;
+            }
+        }
+
+        (void)WH_AUTH_UNLOCK(context);
+    } /* LOCK() */
+    return rc;
+}
+
+
+int wh_Auth_Logout(whAuthContext* context, whUserId user_id)
+{
+    int rc;
+
+    if ((context == NULL) || (context->cb == NULL) ||
+        (context->cb->Logout == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    rc = WH_AUTH_LOCK(context);
+    if (rc == WH_ERROR_OK) {
+        rc = context->cb->Logout(context->context, context->user.user_id,
+                                 user_id);
+        if (rc == WH_ERROR_OK) {
+            /* Clear local session only when logging out the current user */
+            if (user_id == context->user.user_id) {
+                memset(&context->user, 0, sizeof(whAuthUser));
+            }
+        }
+
+        (void)WH_AUTH_UNLOCK(context);
+    } /* LOCK() */
+    return rc;
+}
+
+
+/* Check on request authorization and action permissions for current user
+ * logged in */
+int wh_Auth_CheckRequestAuthorization(whAuthContext* context, uint16_t group,
+                                      uint16_t action)
+{
+    uint16_t    user_id;
+    int         rc;
+    whAuthUser* user;
+
+    if ((context == NULL) || (context->cb == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    rc = WH_AUTH_LOCK(context);
+    if (rc == WH_ERROR_OK) {
+        user    = &context->user;
+        user_id = user->user_id;
+        /* @TODO add logging call here and with resulting return value  */
+
+        if (user_id == WH_USER_ID_INVALID) {
+            /* allow user login request attempt and comm */
+            if (group == WH_MESSAGE_GROUP_COMM ||
+                (group == WH_MESSAGE_GROUP_AUTH &&
+                 action == WH_MESSAGE_AUTH_ACTION_LOGIN)) {
+                rc = WH_ERROR_OK;
+            }
+            else {
+                rc = WH_ERROR_ACCESS;
+            }
+        }
+        else {
+            int groupIndex = (group >> 8) & 0xFF;
+
+            /* A user logged in should by default have access to logging out */
+            if (group == WH_MESSAGE_GROUP_AUTH &&
+                action == WH_MESSAGE_AUTH_ACTION_LOGOUT) {
+                rc = WH_ERROR_OK;
+            }
+            else {
+                if (groupIndex >= WH_NUMBER_OF_GROUPS) {
+                    rc = WH_ERROR_ACCESS;
+                }
+                else if (user->permissions.groupPermissions[groupIndex]) {
+                    /* Check if action is within supported range */
+                    if (action < WH_AUTH_ACTIONS_PER_GROUP) {
+                        /* Get word index and bitmask for this action */
+                        uint32_t wordIndex;
+                        uint32_t bitmask;
+
+                        WH_AUTH_ACTION_TO_WORD_AND_BITMASK(action, wordIndex,
+                                                           bitmask);
+
+                        if (wordIndex < WH_AUTH_ACTION_WORDS &&
+                            (user->permissions
+                                 .actionPermissions[groupIndex][wordIndex] &
+                             bitmask)) {
+                            rc = WH_ERROR_OK;
+                        }
+                        else {
+                            rc = WH_ERROR_ACCESS;
+                        }
+                    }
+                    else {
+                        rc = WH_ERROR_ACCESS;
+                    }
+                }
+                else {
+                    rc = WH_ERROR_ACCESS;
+                }
+            }
+        }
+
+        /* allow authorization override if callback is set */
+        if (context->cb->CheckRequestAuthorization != NULL) {
+            rc = context->cb->CheckRequestAuthorization(context->context, rc,
+                                                        user_id, group, action);
+        }
+        (void)WH_AUTH_UNLOCK(context);
+    } /* LOCK() */
+    return rc;
+}
+
+
+/* Check on key ID use after request has been parsed */
+int wh_Auth_CheckKeyAuthorization(whAuthContext* context, uint32_t key_id,
+                                  uint16_t action)
+{
+    uint16_t    user_id;
+    int         rc = WH_ERROR_ACCESS;
+    int         i;
+    whAuthUser* user;
+
+    if ((context == NULL) || (context->cb == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    rc = WH_AUTH_LOCK(context);
+    if (rc == WH_ERROR_OK) {
+        /* Reset rc to default access denied after successful lock */
+        rc = WH_ERROR_ACCESS;
+
+        user_id = context->user.user_id;
+        user    = &context->user;
+        if (user->user_id == WH_USER_ID_INVALID) {
+            (void)WH_AUTH_UNLOCK(context);
+            return WH_ERROR_ACCESS;
+        }
+
+        /* Check if the requested key_id is in the user's keyIds array */
+        for (i = 0; i < user->permissions.keyIdCount && i < WH_AUTH_MAX_KEY_IDS;
+             i++) {
+            if (user->permissions.keyIds[i] == key_id) {
+                rc = WH_ERROR_OK;
+                break;
+            }
+        }
+
+        if (context->cb->CheckKeyAuthorization != NULL) {
+            rc = context->cb->CheckKeyAuthorization(context->context, rc,
+                                                    user_id, key_id, action);
+        }
+        (void)WH_AUTH_UNLOCK(context);
+    } /* LOCK() */
+    return rc;
+}
+
+/********** API That Manages User Database ******************************/
+
+int wh_Auth_UserAdd(whAuthContext* context, const char* username,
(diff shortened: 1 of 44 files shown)


bigbrett:
@JacobBarthelmeh merge conflicts

bigbrett (changes requested):
Great work Jacob! Just did a first pass of this, and it mostly aligns with what I was envisioning, but with one caveat. I think that the actual authorization check and logic should be part of the generic "front end" and not delegated to the back-end. See my "architectural concern" comment below.

Also, a few requests right off the bat, before I get too deep into review:

1. Make the auth feature "opt-in" so it is off by default (e.g change to `WOLFHSM_CFG_CLIENT_AUTH` or something like that - I like calling out client auth specifically as a feature description).  Protect all auth specific code (including the struct elements in the server context) with the feature macro. Reason being I think the bulk of our customers will not be using this feature, even though it is of critical importance for FIPS L2+ customers. 
2. Since the feature should be opt-in, ensure that when the feature is turned on, the server is able to be initialized with a NULL auth config and handle this scenario safely. This should indicate that there is no auth needed. This facilitates backwards compatibility with tests (e.g. dont need to log in as admin in every test).
3. Try and use `WH_ERROR_OK` for success return. We were inconsistent about this in old code, but want to standardize around it going forward. Also helps distinguish wolfHSM returns from other lib returns like wolfCrypt/wolfSSL

Will keep going deeper into this but figured I'd speak up early so we can discuss. Let me know your thoughts.

src/wh_auth.c
```diff
@@ -0,0 +1,261 @@
...
+
+    /* Clear the user context */
+    memset(&context->user, 0, sizeof(whAuthUser));
+    return WH_ERROR_OK;
+}
+
+
+/* Check on request authorization and action permissions for current user
+ * logged in */
+int wh_Auth_CheckRequestAuthorization(whAuthContext* context, uint16_t group,
```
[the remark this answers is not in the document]
JacobBarthelmeh: Yeah, it's a good point. A lot of the authentication could be moved into wolfHSM itself. Will keep the callback though because there may be cases where a user wishes to override the default behavior.

port/posix/posix_auth.c
bigbrett: 1) I'm wondering if this shouldn't be scoped to the POSIX port and called "posix auth" since there is nothing POSIX-specific about it. You could theoretically use this on any of our ports. Perhaps we can include this as a built-in auth mechanism that could be used by any port, similar to our shared memory transport impl (`wh_transport_mem.{c,h}`). Any objection to this? Could call it `wh_auth_basic` or something like that? 

2) Did you have a scheme to persist the auth database in mind? Here it seems like it is intended for the server to have a hardcoded list and just iterate over each element and "add user" on startup before processing requests. Is that the intent? Could we add an `InitFromNvm` function that takes an object ID as an argument, where the auth list could be pulled out of NVM and configured out-of-band (e.g. provisioned into NVM at factory)? I know there will be the typical issues of serializing a raw struct, compat across build config updates, but perhaps this is ok for now?

bigbrett (changes requested):
Found a bug and have a few suggestions.

wolfhsm/wh_message.h
```diff
@@ -98,6 +99,16 @@ enum {
...
 
+/* auth actions */
+enum {
+    WH_AUTH_ACTION_LOGIN,
+    WH_AUTH_ACTION_LOGOUT,
+    WH_AUTH_ACTION_USER_ADD,
+    WH_AUTH_ACTION_USER_DELETE,
+    WH_AUTH_ACTION_USER_MODIFY,
+    WH_AUTH_ACTION_PERMISSION_SET,
+};
```
bigbrett: duplicated in `wh_message_auth.h`. Imagine these ones can be deleted?

src/wh_server.c
```diff
@@ -316,6 +337,177 @@ static int _wh_Server_HandlePkcs11Request(whServerContext* server,
     return rc;
 }
 
+#ifdef WOLFHSM_CFG_ENABLE_AUTHENTICATION
+/* Helper to format an authorization error response for any group/action.
+ * All response structures have int32_t rc as the first field.
+ * Returns the response size to send. */
+static uint16_t _wh_Server_FormatAuthErrorResponse(uint16_t magic,
```
bigbrett:
```suggestion
static uint16_t _FormatAuthErrorResponse(uint16_t magic,
```

wolfhsm/wh_auth.h
```diff
@@ -0,0 +1,306 @@
...
+#define WH_USER_ID_INVALID ((whUserId)0)
+
+/* Authentication method enumeration */
+typedef enum {
+    WH_AUTH_METHOD_NONE = 0,
+    WH_AUTH_METHOD_PIN,
+    WH_AUTH_METHOD_CERTIFICATE,
+} whAuthMethod;
+
+#define WH_NUMBER_OF_GROUPS 14
```
bigbrett: This shouldn't be hardcoded here. It should ideally be defined with the actual groups in `wh_message.h`. Perhaps can be auto-computed based on the new group ID fields you may need to create to fix the bug in the previous review?

src/wh_auth.c
```diff
@@ -0,0 +1,337 @@
...
+                    /* Get word index and bitmask for this action */
+                    uint32_t wordAndBit = WH_AUTH_ACTION_TO_WORD_AND_BIT(action);
+                    uint32_t wordIndex   = WH_AUTH_ACTION_WORD(wordAndBit);
+                    uint32_t bitmask     = WH_AUTH_ACTION_BIT(wordAndBit);
+
+                    if (wordIndex < WH_AUTH_ACTION_WORDS &&
+                        (user->permissions.actionPermissions[groupIndex]
+                                                                [wordIndex] &
+                            bitmask)) {
+                        rc = WH_ERROR_OK;
```
bigbrett: Think there is another bug here: The packing scheme here has a collision problem. The macros pack two values into overlapping bit ranges. The word index is stored in bits 16-23, but the bitmask occupies bits 0-31 - so when` action % 32 > 16`, the bitmask bleeds into  the word index bits.
             
 For example, action 49 gives word=1 and bit position=17. The packed value becomes `(1 << 16) | (1 << 17) = 0x30000`. When we extract the word index with `(0x30000 >> 16) & 0xFF`, we get 3 instead of 1. The bitmask extraction is also wrong - we get `0x30000` instead of `0x20000`

This affects actions 48-63, 80-95, 112-127, 144-159, 176-191, 208-223, and 240-255 (about half of all possible actions). Authorization checks for these actions will look at the wrong word in the permissions array and check the wrong bits.

Suggest switching to a uint64_t return type with the word index in the upper 32 bits to avoid any overlap.

wolfhsm/wh_auth.h
```diff
@@ -0,0 +1,306 @@
...
+/** Auth Manager Context and Config */
+
+/* Simple helper context structure associated with an Auth Manager instance */
+typedef struct whAuthContext_t {
+    whAuthCb*  cb;
+    whAuthUser user;
+    void*      context;
+} whAuthContext;
+
+#define WOLFHSM_MAX_CERTIFICATE_LEN 2048
```
bigbrett: this should not be defined here. Don't we already have a macro for this? It should also be `WOFHSM_CFG_XXX` if you are adding a new config value.

bigbrett (changes requested):
(part 2, sorry terminated my last review early)

port/posix/posix_auth.c
JacobBarthelmeh: 1) Yeah it could be moved to a non posix location, at one point considered placing it in the examples directory.  Showing it as a simplified example of user database. I'll look at moving it there or in demo's, I think in a different location than port makes sense for it.

2) NVM had crossed my mind and it would be great for persistence. Wanted to keep this initial users setup as simple as possible though while ironing out the authentication part of the feature in wolfHSM itself.

src/wh_auth_base.c
```diff
@@ -0,0 +1,543 @@
...
+    if (cm == NULL) {
+        return WH_ERROR_ABORTED;
+    }
+    err = wolfSSL_CertManagerLoadCABuffer(cm, found_user->credentials,
+                                          found_user->credentials_len,
+                                          WOLFSSL_FILETYPE_ASN1);
+    if (err != WOLFSSL_SUCCESS) {
+        rc = WH_ERROR_ABORTED;
+    }
+    err = wolfSSL_CertManagerVerifyBuffer(cm, certificate, certificate_len,
```
bigbrett: This should be gated on success return from the CA buffer load

src/wh_auth_base.c
```diff
@@ -0,0 +1,543 @@
...
+    }
+
+    user                 = &users[user_id - 1];
+    user->user.is_active = false;
+    (void)context;
+    return WH_ERROR_OK;
+}
+
+
+int wh_Auth_BaseUserAdd(void* context, const char* username,
```
bigbrett: do we have a strategy for managing duplicate usernames? This allows creating multiple users with identical usernames but different user IDs. Since `wh_Auth_BaseFindUser` just `strcmps` the usrename and returns the first match, later duplicates become unreachable/orphaned.

src/wh_auth_base.c
```diff
@@ -0,0 +1,543 @@
...
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+    authCheck_len = WC_SHA256_DIGEST_SIZE;
+#else
+    authCheck_len = auth_data_len;
+#endif /* WOLFHSM_CFG_NO_CRYPTO */
+
+    found_user = wh_Auth_BaseFindUser(username);
+    if (found_user != NULL && found_user->method == WH_AUTH_METHOD_PIN &&
+        found_user->credentials_len == authCheck_len &&
+        memcmp(found_user->credentials, authCheck, authCheck_len) == 0) {
```
bigbrett: I suppose this should probably be a constant-time comparison for the non-hashed case, if we are going to support it. I suppose saying that if you waive crypto then you don't care about side channels is okay too

src/wh_auth_base.c
```diff
@@ -0,0 +1,543 @@
...
+                return rc;
+            }
+            if (user->credentials_len != WC_SHA256_DIGEST_SIZE ||
+                memcmp(user->credentials, hash, WC_SHA256_DIGEST_SIZE) != 0) {
+                return WH_ERROR_ACCESS;
+            }
+#else
+            /* When crypto is disabled, compare PINs directly */
+            if (user->credentials_len != current_credentials_len ||
+                memcmp(user->credentials, current_credentials,
```
bigbrett: see above comment re: constant compare. Perhaps necessary, perhaps not, open to discussion

src/wh_auth.c
```diff
@@ -0,0 +1,463 @@
...
+        else {
+            rc = WH_ERROR_ACCESS;
+        }
+    }
+    else {
+        int groupIndex = (group >> 8) & 0xFF;
+
+        /* some operations a user logged in should by default have access to;
+         * - logging out
+         * - updating own credentials */
```
bigbrett: I might be missing something here, but seems like this is never restricted to a user's own credentials. Pretty sure this allows any authenticated user can set credentials for any other user?

bigbrett (changes requested):
Partial review - will pick back up later. But think I found some juicy things to chew on.

src/wh_auth_base.c
```diff
@@ -0,0 +1,543 @@
...
+    }
+
+    /* Set new credentials */
+    user->method = method;
+    if (new_credentials_len > 0) {
+        if (method == WH_AUTH_METHOD_PIN) {
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+            /* Hash PIN before storing */
+            unsigned char hash[WC_SHA256_DIGEST_SIZE];
+            int           rc =
```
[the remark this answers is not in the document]
JacobBarthelmeh: Great catch. Nope not intentional, is a bug after refactoring and moving from the port/posix directory to src/wh_auth_base.c.

src/wh_auth.c
```diff
@@ -0,0 +1,463 @@
...
+        if (user->permissions.keyIds[i] == key_id) {
+            rc = WH_ERROR_OK;
+            break;
+        }
+    }
+
+    (void)WH_AUTH_UNLOCK(context);
+
+    if (context->cb->CheckKeyAuthorization != NULL) {
+        rc = context->cb->CheckKeyAuthorization(context->context, rc, user_id,
```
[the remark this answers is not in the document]
JacobBarthelmeh: Yes, would make sense. The callback could have a race between other actions like setting permissions in the extreme case that would potentially lead to unexpected results. I'll adjust the unlock location for this and for the check authorization callback above it.

src/wh_server.c
```diff
@@ -320,6 +520,38 @@ int wh_Server_HandleRequestMessage(whServerContext* server)
...
+                do {
+                    rc = wh_CommServer_SendResponse(server->comm, magic, kind,
+                                                    seq, resp_size, data);
+                } while (rc == WH_ERROR_NOTREADY);
+
+                /* Log the authorization failure */
+                WH_LOG_ON_ERROR_F(
+                    &server->log, WH_LOG_LEVEL_ERROR, WH_AUTH_PERMISSION_ERROR,
+                    "Authorization failed for (group=%d, action=%d, seq=%d)",
+                    group, action, seq);
```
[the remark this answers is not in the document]
JacobBarthelmeh: at the moment this is intentional, log info should be reviewed if wanting more detail on auth permission failure reasons

src/wh_auth_base.c
```diff
@@ -0,0 +1,575 @@
...
+
+    for (i = 0; i < WH_AUTH_BASE_MAX_USERS; i++) {
+        if (users[i].user.user_id == WH_USER_ID_INVALID) {
+            break;
+        }
+
+        /* do not allow duplicate users with same name */
+        if (strcmp(users[i].user.username, username) == 0) {
+            return WH_ERROR_BADARGS;
+        }
```
bigbrett, line 295: The uniqueness loop in `UserAdd` breaks at the first empty slot instead of scanning the whole array. `UserDelete` zeroes slots in place without compacting, so any delete creates a gap that blinds the check to everything past it.

**The attack (identity confusion / lockout):**
1. State: `[0]=bob, [1]=admin`.
2. Admin deletes bob → `[0]=empty, [1]=admin`.
3. Attacker with `USER_ADD` perm calls `UserAdd(username="admin", ...)` with their own PIN.
4. Loop breaks at slot 0 (empty), never scans slot 1 → duplicate admin is inserted.
5. `FindUser("admin")` returns the first match → always the impostor.
6. Real admin's login fails (PIN doesn't match impostor's hash) = **locked out**. Attacker logs in as "admin" at their own (non-admin) permissions = **identity confusion**.

**Key insight:** The attacker doesn't gain admin rights — they gain the *name* "admin", which is enough to shadow the real account in every name-based lookup.

**Fix:** Two-pass the loop — scan all slots for duplicates, then find a free slot. Or compact the array on delete.

bigbrett (changes requested):
As discussed offline, one blocking bugfix needed, as well as a documentation update to tag this feature as experimental. I will post a list of the other issues found in a subsequent review that can then be addressed with follow-up PRs

bigbrett (approved):
Approved given that the feature is now tagged as "experimental" and individual fixes for the remaining security related issues can be introduced in subsequent PRs. 

```md
# Security Review — Distilled

---

## Vuln 1 — Cert auth has no proof-of-possession (`src/wh_auth_base.c:147-184`)

**The bug:** Certificate login only runs `CertManagerVerifyBuffer` against the stored CA. The server never checks that the client actually holds the matching private key.

**Why it matters:** Certificates are public documents. Verifying the chain proves "this cert was issued by that CA" — not "the sender owns the private key."

**The attack:**
1. Admin provisions Alice with `method=CERTIFICATE`, credential = Alice's CA.
2. Attacker grabs *any* valid cert signed by that CA (off the wire, from a public directory, from Alice's public leaf cert).
3. Attacker sends `AuthLogin(method=CERTIFICATE, username="alice", auth_data=<cert>)`.
4. Chain verifies → `is_active=true`, full permissions → attacker is now Alice.

**Key insight:** Presenting a cert should not equal proving identity. Every real cert-auth protocol pairs the cert with a signature over a server-chosen nonce.

**Fix:** Add a challenge-response action — server issues a random nonce, client returns `(cert, signature(nonce))`, server verifies *both*. Alternatively have client obtain nonce out-of-band or via earlier message.

---

## Vuln 2 — Cert login does not bind subject to username (`src/wh_auth_base.c:147-167`)

**The bug:** After the cert chain verifies, nothing compares the presented cert's subject/CN/SAN/pubkey to the user being logged into. `whAuthBase_User` doesn't even have a field to store an expected subject.

**Why it matters:** The intended model is "CA-as-credential" — multiple users share the same enterprise CA. Any legitimately-issued leaf from that CA will chain-verify against any user trusting it.

**The attack:**
1. Deployment has Alice (admin) and Bob (non-admin), both stored with `credential = EnterpriseCA`.
2. Bob holds his own genuine leaf cert signed by that CA.
3. Bob sends `LOGIN(username="alice", cert=bob's_leaf)`.
4. Server finds alice's record, loads her CA, calls `VerifyBuffer(bob's_leaf)` → chain is valid → server assigns Bob the session with Alice's `user_id` and admin permissions.

**Key insight:** Distinct from Vuln 1. Vuln 1 = *anyone with a valid signed cert wins*. Vuln 2 = *even a legitimate private-key holder can log in as a different user who trusts the same CA*. Both must be fixed.

**Fix:** Store an `expected_subject` or `expected_pubkey_hash` per user when storing credentials, OR operate on subject/CN/SAN/pubkey fields. Then on login, after chain verification, parse the leaf and compare. Fail closed on mismatch.

---

## Vuln 3 — `UserSetCredentials` missing caller authorization (`src/wh_auth.c:407-430`, `src/wh_auth_base.c:455-573`)

**The bug:** Three reinforcing defects that together let anyone with the `USER_SET_CREDENTIALS` action bit rewrite any other user's credentials:

1. **API layer:** `wh_Auth_UserSetCredentials` forwards the client-supplied `user_id` to the backend with no check that `user_id == caller` or caller is admin. Its siblings (`UserDelete`, `UserSetPermissions`) *do* pass the caller's id.
2. **Callback signature:** `UserSetCredentials`'s backend signature is missing the `current_user_id` parameter that peer callbacks include. So even if the backend wanted to check the caller, it can't — it only sees the target.
3. **Null-credential shortcut:** If the target user has `credentials_len == 0`, the backend accepts `current_credentials = NULL` without any proof-of-knowledge. This is reachable because `UserAdd` permits creating users with no credential.

**The attack (pre-provisioned admin takeover):**
1. Admin pre-provisions Bob (admin, credentials pending) via `whnvmtool` with `credentials_len=0`.
2. Non-admin Alice has only `(AUTH, USER_GET)` and `(AUTH, USER_SET_CREDENTIALS)`.
3. Alice calls `UserGet("bob")` → gets `user_id=2, admin=1`.
4. Alice calls `UserSetCredentials(user_id=2, current_len=0, new="1234")` → accepted because Bob has no credential set.
5. Alice logs in as "bob" with PIN "1234" → full admin.

**Key insight:** This is especially dangerous with bulk provisioning — any pre-created account without a credential is a landmine until an admin sets its PIN.

**Fix:** Add `current_user_id` to the callback; pass `context->user.user_id` from the API; require `caller == target` OR `WH_AUTH_IS_ADMIN(caller)`; forbid the null-credential shortcut on cross-user calls by non-admins.
**Alternative Fix:** Restrict ALL user modification to admin only

---

## Vuln 4 — Session carry-over on abrupt disconnect (`src/wh_server.c:160-173`)

**The bug:** `auth_ctx->user` is only cleared by the explicit `COMM_CLOSE` handler. A TCP RST, a client crash, or a malicious client that just drops the socket does not clear it. The disconnect callback only flips `server->connected` — it never touches `server->auth`.

**Why it matters:** The reference POSIX server uses a `Cleanup → Init → re-accept` pattern. `Cleanup` memsets the server context but does not call `wh_Auth_Cleanup` on the externally-owned auth context, and `Init` re-attaches the same auth pointer without re-zeroing `user`. So the prior session's identity survives across reconnects.

**The attack:**
1. Admin connects over TCP, authenticates with a PIN.
2. The socket closes without a `COMM_CLOSE` (RST, crash, or malicious).
3. `auth_ctx.user` still holds `{user_id=admin, is_active=true, permissions=admin}`.
4. Attacker's new TCP connection is accepted. Attacker sends no `AUTH_LOGIN`.
5. Attacker issues a privileged request. `CheckRequestAuthorization` sees `user_id != INVALID`, evaluates the stale admin permissions → authorized.

**Key insight:** A logout that only runs on the graceful path is not a logout — it's a courtesy. Abrupt disconnect is the common case on real networks.

**Fix:** On transition to `WH_COMM_DISCONNECTED`, call `wh_Auth_Logout` for any non-invalid user_id. Additionally, have `wh_Server_Init` re-zero `auth->user` so reconnect-reuse is safe by default.

---

## Vuln 5 [FIXED]  — Duplicate-username check terminates at first empty slot (`src/wh_auth_base.c:286-295`)

**The bug:** The uniqueness loop in `UserAdd` breaks at the first empty slot instead of scanning the whole array. `UserDelete` zeroes slots in place without compacting, so any delete creates a gap that blinds the check to everything past it.

**The attack (identity confusion / lockout):**
1. State: `[0]=bob, [1]=admin`.
2. Admin deletes bob → `[0]=empty, [1]=admin`.
3. Attacker with `USER_ADD` perm calls `UserAdd(username="admin", ...)` with their own PIN.
4. Loop breaks at slot 0 (empty), never scans slot 1 → duplicate admin is inserted.
5. `FindUser("admin")` returns the first match → always the impostor.
6. Real admin's login fails (PIN doesn't match impostor's hash) = **locked out**. Attacker logs in as "admin" at their own (non-admin) permissions = **identity confusion**.

**Key insight:** The attacker doesn't gain admin rights — they gain the *name* "admin", which is enough to shadow the real account in every name-based lookup.

**Fix:** Two-pass the loop — scan all slots for duplicates, then find a free slot. Or compact the array on delete. Restricting ALL modification to admin only would eliminate the possibility of a client-driven exploit, but still leave the erroneous code in place.

---

## Vuln 6 — `UserAdd` does not require caller permissions to be a superset (`src/wh_auth.c:329-338`)

**The bug:** `UserAdd` only checks one thing: non-admins can't mint admins. It never checks that the new user's requested permission bitmap is a subset of the caller's own permissions. The backend writes whatever bitmap the caller supplied straight into the new slot.

**The attack (verified end-to-end):**
1. Admin gives Alice a tiny role: `USER_ADD` + `USER_SET_CREDENTIALS` only. No crypto, NVM, key, or image perms.
2. Alice calls `UserAdd(username="victim", permissions={groupPermissions=0xFF…, admin=0}, credentials=NULL)` → server accepts the bitmap as-is.
3. Alice calls `UserSetCredentials(victim, current=NULL, new="victimpin")` → accepted via Vuln 3's null-credential shortcut.
4. Alice logs out, logs back in as "victim" → session now carries crypto/NVM/key/image permissions Alice was never granted.

**Key insight:** Without a subset check, `USER_ADD` is effectively "grant yourself any non-admin permission you want." Vulns 3 and 6 compose: Vuln 6 mints the elevated account, Vuln 3 lets the attacker log into it.

**Fix:** In `wh_Auth_UserAdd`, require the requested `groupPermissions`, `actionPermissions`, and `keyIds` be bitwise subsets of `context->user.permissions` unless the caller is admin. Reject with `WH_AUTH_PERMISSION_ERROR` otherwise. Additionally, forbid `UserAdd` with `credentials_len == 0` for non-admin callers. Fixing Vuln 3 is a prerequisite.
**Alternative Fix:** Restrict ALL user modification to admin only


## Relationships Between Vulns

- Vulns 1 & 2 are independent cert-auth defects — fixing one doesn't fix the other.
- Vulns 3 & 6 compose — Vuln 6 mints an over-privileged account, Vuln 3 lets the attacker install credentials for it.
- Vuln 4 makes any reconnect-reuse server pattern unsafe regardless of the auth method.
- Restricting ALL user modification to admin only would eliminate Vulns 3 and 6, and eliminate an exploit path for 5 (though the bug would remain).

```

billphipps (approved):
Great start!  Let's continue to refine this model so we can make it part of the mainline.

PR #278, opened 2026-01-28
Implement timeout capability. Apply timeout to crypto response

### LINKED ISSUE #130: Potential Infinite Loop in RNG Generation When Server Is Unresponsive
Description
The wh_Client_RngGenerate function contains a blocking wait for server responses that could result in an infinite loop if the server becomes unresponsive or malfunctions.

Current Behavior
In the current implementation, when waiting for a response from the server, the code enters a do-while loop that continues as long as wh_Client_RecvResponse() returns WH_ERROR_NOTREADY:
Code snippet:
/* Receive response */
if (ret == 0) {
    do {
        ret = wh_Client_RecvResponse(ctx, NULL, NULL, &res_len, (uint8_t*)packet);
    } while (ret == WH_ERROR_NOTREADY);
}
If the server stops responding or fails to provide a valid response, this loop will continue indefinitely, causing the client application to hang.
Expected Behavior
The function should implement a timeout mechanism to prevent infinite waiting. After a reasonable number of retries or a specific time period, it should exit the loop with an appropriate error code.
Few Solutions for reference:
1.	Retry count-based timeout:
int retry_count = 0;
const int max_retries = WOLFHSM_MAX_RESPONSE_RETRIES;

do {
    ret = wh_Client_RecvResponse(ctx, NULL, NULL, &res_len, (uint8_t*)packet);
    if (ret == WH_ERROR_NOTREADY && ++retry_count >= max_retries) {
        ret = WH_ERROR_TIMEOUT;
        break;
    }
} while (ret == WH_ERROR_NOTREADY);

2.	Time-based timeout:
uint32_t start_time = wh_GetCurrentTimeMs();
uint32_t timeout_ms = WOLFHSM_RESPONSE_TIMEOUT_MS;

do {
    ret = wh_Client_RecvResponse(ctx, NULL, NULL, &res_len, (uint8_t*)packet);
    if (ret == WH_ERROR_NOTREADY && 
        (wh_GetCurrentTimeMs() - start_time) > timeout_ms) {
        ret = WH_ERROR_TIMEOUT;
        break;
    }
} while (ret == WH_ERROR_NOTREADY);

Additional Context
This issue affects not only the RNG generation but potentially other client functions that use the same pattern for waiting on server responses. A comprehensive review of all client-server communication patterns should be conducted.

AlexLanzano:
fixes #130 
# Timeout Functionality: Client Perspective

## 1. Configuration at Init Time

When creating a client, you provide a `whTimeoutConfig` specifying the timeout duration and an optional callback:

```c
whTimeoutConfig timeoutCfg = {
    .timeoutUs = WH_SEC_TO_USEC(5),   /* 5-second timeout */
    .expiredCb = myTimeoutHandler,      /* optional callback on expiry */
    .cbCtx     = myAppContext,          /* context passed to callback */
};

whClientConfig clientCfg = {
    .comm              = &commConfig,
    .respTimeoutConfig = &timeoutCfg,   /* attach timeout config */
};

wh_Client_Init(&clientCtx, &clientCfg);
```

During `wh_Client_Init` (`src/wh_client.c:84-89`), the config is copied into an embedded `whTimeoutCtx respTimeout[1]` inside the client context via `wh_Timeout_Init()`. This stores the timeout duration and callback but doesn't start any timer yet.

If `respTimeoutConfig` is NULL, the timeout context is left zeroed and effectively disabled (a `timeoutUs` of 0 means "never expires").

## 2. What Happens During a Crypto Call

Before this PR, every crypto function in `wh_client_crypto.c` had this pattern after sending a request:

```c
/* Old pattern -- infinite busy-wait */
do {
    ret = wh_Client_RecvResponse(ctx, &group, &action, &res_len, dataPtr);
} while (ret == WH_ERROR_NOTREADY);
```

If the server never responded, the client would spin forever.

The PR replaces all ~30 of these with a single helper `_recvCryptoResponse()` (`src/wh_client_crypto.c:165-180`):

```c
static int _recvCryptoResponse(whClientContext* ctx,
                               uint16_t* group, uint16_t* action,
                               uint16_t* size, void *data)
{
    int ret;

#ifdef WOLFHSM_CFG_ENABLE_TIMEOUT
    ret = wh_Client_RecvResponseTimeout(ctx, group, action, size, data,
                                        ctx->respTimeout);
#else
    do {
        ret = wh_Client_RecvResponse(ctx, group, action, size, data);
    } while (ret == WH_ERROR_NOTREADY);
#endif

    return ret;
}
```

When timeout is enabled, it delegates to `wh_Client_RecvResponseTimeout`. When disabled, the old infinite-loop behavior is preserved.

## 3. The Timeout Receive Loop

`wh_Client_RecvResponseTimeout` (`src/wh_client.c:211-231`) does this:

1. **Starts the timer** -- calls `wh_Timeout_Start()` which snapshots the current time via `WH_GETTIME_US()` into `timeout->startUs`.

2. **Polls for a response** -- calls `wh_Client_RecvResponse()` in a loop.

3. **On each `WH_ERROR_NOTREADY`**, checks `wh_Timeout_Expired()`:
   - Gets the current time via `WH_GETTIME_US()`
   - Computes `(now - startUs) >= timeoutUs`
   - If expired: invokes the `expiredCb` (if set), then returns `WH_ERROR_TIMEOUT`
   - If not expired: loops again

4. **On any other return value** (success or error), returns immediately.

```
Client App                    _recvCryptoResponse                 wh_Timeout
    |                                |                                |
    |-- wh_Client_AesCbc() --------> |                                |
    |                                |-- wh_Timeout_Start --------> capture time
    |                                |                                |
    |                                |-- RecvResponse (NOTREADY)      |
    |                                |-- Expired? ------------------> no
    |                                |-- RecvResponse (NOTREADY)      |
    |                                |-- Expired? ------------------> no
    |                                |   ...                          |
    |                                |-- RecvResponse (NOTREADY)      |
    |                                |-- Expired? ------------------> YES
    |                                |                                |-- expiredCb()
    |<-- WH_ERROR_TIMEOUT -----------|                                |
```

## 4. What the Client Sees

From the application's perspective, the crypto APIs (`wh_Client_AesCbc`, `wh_Client_RsaFunction`, `wh_Client_EccSign`, etc.) now return `WH_ERROR_TIMEOUT` (-2010) instead of hanging indefinitely. The application can then decide how to handle it -- retry, log, fail gracefully, etc.

The `expiredCb` fires *before* the error is returned, so you can use it for logging or cleanup without needing to check the return code first.

## 5. Scope Limitations

A few things to note about the current design:

- **Only crypto responses are covered.** Non-crypto client calls (key management, NVM operations, comm init) still use the old infinite-wait pattern. The timeout is specifically wired into `_recvCryptoResponse`.
- **The timeout is per-client, not per-call.** All crypto operations for a given client share the same `respTimeout` context with the same duration. You can call `wh_Timeout_Set(ctx->respTimeout, newValue)` to change it between calls, but there's no per-operation override.

### DIFF
--- a/wolfhsm/wh_client.h
+++ b/wolfhsm/wh_client.h
@@ -109,9 +109,6 @@ struct whClientContext_t {
     uint16_t     last_req_id;
     uint16_t     last_req_kind;
     uint32_t     cryptoAffinity;
-#ifdef WOLFHSM_CFG_ENABLE_TIMEOUT
-    whTimeout respTimeout;
-#endif
 #ifdef WOLFHSM_CFG_DMA
     whClientDmaContext dma;
 #endif /* WOLFHSM_CFG_DMA */


src/wh_client.c
```diff
@@ -199,6 +208,31 @@ int wh_Client_RecvResponse(whClientContext *c,
...
+    int ret;
+
+    if ((c == NULL) || (timeout == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    ret = wh_Timeout_Start(timeout);
+    if (ret != WH_ERROR_OK) {
+        return ret;
+    }
```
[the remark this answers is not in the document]
AlexLanzano: This is intended. There may be use cases where the user wants to temporarily disable the response timeout and reenable it at a different time.

wolfhsm/wh_client.h
```diff
@@ -193,6 +200,23 @@ int wh_Client_SendRequest(whClientContext* c, uint16_t group, uint16_t action,
...
+ * @param c The client context.
+ * @param out_group Pointer to store the received group value.
+ * @param out_action Pointer to store the received action value.
+ * @param out_size Pointer to store the received size value.
+ * @param data Pointer to store the received data.
+ * @param timeout The timeout context to use.
+ * @return 0 if successful, WH_ERROR_TIMEOUT on expiration, or a negative value
+ * if an error occurred.
+ */
+int wh_Client_RecvResponseTimeout(whClientContext* c, uint16_t* out_group,
```
bigbrett: consider renaming `wh_Client_RecvResponseBlockingWithTimeout()` to be explicit that this function contains a blocking/polling loop that can only be exited by the timeout expiring. This is to distinguish it from `wh_Client_RecvResponse()` which would otherwise be named similarly, but is a non-blocking async function

src/wh_timeout.c
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    if (timeout->timeoutUs == 0) {
+        return 0;
+    }
+
+    nowUs = WH_GETTIME_US();
+    expired = (nowUs - timeout->startUs) >= timeout->timeoutUs;
+    if (expired && (timeout->expiredCb != NULL)) {
+        timeout->expiredCb(timeout->cbCtx);
```
bigbrett: OK so what if we allowed the callback to override the timeout here, based on the return value? So the callback could return a code that either says 1) "yes, I acknowledge the timeout, proceed with the failure return to the caller, I've done what I need to do to bring the system to a safe state" or 2) I know the timeout expired, but I want `wh_Client_RecvRequestTimeout()` to KEEP polling the request loop. 

This is necessary, since we don't really have a good way to "retry" a blocking operation like crypto, since calling the crypto function again would try and send the initial request, which would fail due to a request already being in flight. Perhaps "wait a little longer" is something we should support?


For this to work, we would need to pass the actual timeout context into the callback, not just the callback context. This would allow the callback to "restart" the timer. I think passing the message parameters for the request (group and action) for additional context would be important as well. 

Then in `wh_ClientRecvRequestTimeout()`, we could check the actual return value of `wh_Timeout_Expired()` when deciding whether or not to break from the loop.

Thoughts?

wolfhsm/wh_client.h
AlexLanzano: Fixed

wolfhsm/wh_client.h
```diff
@@ -193,6 +200,23 @@ int wh_Client_SendRequest(whClientContext* c, uint16_t group, uint16_t action,
...
+                                  uint16_t* out_action, uint16_t* out_size,
+                                  void* data, whTimeoutCtx* timeout);
```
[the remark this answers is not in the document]
AlexLanzano: Fixed

test/wh_test_timeout.c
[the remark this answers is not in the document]
AlexLanzano: Implemented this test

bigbrett (changes requested):
two more tweaks then we should be good2go

rizlik (changes requested):
Beside the a single comment on the return value of  `wh_Timeout_Expired` I've a couple of more general design questions:

1. Is ExpireCb really needed? the application can already react to the expiration after the function returns `WH_ERROR_TIMEOUT`.

2. Can timeouts be abstracted?
I would expect callbacks to be able to abstract away timeout implementation  (Init/Start/Stop/IsExpired) in similar fashion mutexes are abstracted by LockInit/LockCleanup/LockAquire/LockRelease cbs.
This avoids to commit to a specific granularity for the timer and make easier to use hardware timers in ports.

2. I do think that we should consider to bind the timeout logic in the communication layer/ctx/config: 
my rationale here is that while is good to have a built-in easy timeout implementation available,  application might need to provide their own timeout logic that avoids polling the communication channel in the first place (using interrupt + a hw timer).
Another example is posix: in a "posix transport" I imagine to set a timeout on the recv operation more that polling continuously the socket and checking for timeout expiration.
In those scenarios, the communication `recv` callback will not have a natural way to get the timeout configuration and timeout value. 
While we can start to use `container_of` tricks to get the client context from the comm context, I do think timeout belongs to communication.
If we don't want to just fully delegate the timeout logic to the port `recv` callback, we can probably lean towards a middle ground solution by adding the timeout logic in `wh_CommClient_RecvReponse`. In this case probably we want to add a new  returning error option for the `recv` callback (like `WH_ERROR_INTERRUPTED`). If the callback was interrupted we can check the timeout for expiration and either return `WH_ERR_TIMEOUT` or invoke the `recv` callback again.
In this case, abstracting away timeouts fits better imo.

port/posix/posix_timeout.c
```diff
@@ -0,0 +1,159 @@
...
+int posixTimeout_Start(void* context)
+{
+    posixTimeoutContext* ctx = (posixTimeoutContext*)context;
+
+    if (ctx == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    if (!ctx->initialized) {
+        return WH_ERROR_NOTREADY;
```
bigbrett: Recommend `WH_ERROR_BADARGS`. `WH_ERROR_NOTREADY` has a specific transport-focused meaning.

port/posix/posix_timeout.c
```diff
@@ -0,0 +1,159 @@
...
+int posixTimeout_Stop(void* context)
+{
+    posixTimeoutContext* ctx = (posixTimeoutContext*)context;
+
+    if (ctx == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    if (!ctx->initialized) {
+        return WH_ERROR_NOTREADY;
```
bigbrett: Recommend `WH_ERROR_BADARGS`. `WH_ERROR_NOTREADY` has a specific transport-focused meaning.

test/wh_test_timeout.c
```diff
@@ -97,24 +99,29 @@ static int whTest_TimeoutAesCbc(void)
     }};
 
     /* Client configuration with timeout */
-    whTimeoutConfig timeoutCfg = {
-        .timeoutUs = 1,
-        .expiredCb = NULL,
-        .cbCtx     = NULL,
+    posixTimeoutContext posixCtx                         = {0};
```
bigbrett: `git-clang-format main`

bigbrett: this is just blindly using posix stuff unprotected by `WOLFHSM_CFG_TEST_POSIX`. Protect with `WOLFHSM_CFG_TEST_POSIX` and name the test accordingly such that it is obviously a posix-only test for now.

bigbrett (changes requested):
Some nits and some bigger architectural changes to testing now that we refactored timeouts to comm layer, therefore making it fully async

test/wh_test_timeout.c
```diff
@@ -0,0 +1,489 @@
...
+        WH_TEST_ASSERT_RETURN(0);
+    }
+
+    return wh_Server_SetConnected(timeoutTestServerCtx, connected);
+}
+
+static int whTest_TimeoutAesCbc(void)
+{
+    WH_TEST_PRINT("Testing timeout AES CBC...\n");
+    int rc = 0;
```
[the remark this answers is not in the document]
AlexLanzano: Punting to later PR

PR #281, opened 2026-02-04
Implement HW/SW crypto affinity

AlexLanzano:
### Summary

This PR adds a new client API to dynamically switch between hardware and software cryptographic implementations at runtime. This allows clients to control whether the server uses crypto callbacks (hardware acceleration) or pure software wolfCrypt implementations.

### Changes

- **New Client API** (`wh_client.c`, `wh_client.h`)
  - `wh_Client_SetCryptoAffinity()` - blocking call
  - `wh_Client_SetCryptoAffinityRequest()` / `wh_Client_SetCryptoAffinityResponse()` - async API

- **New Message Types** (`wh_message_comm.c`, `wh_message_comm.h`)
  - `WH_MESSAGE_COMM_ACTION_SET_CRYPTO_AFFINITY` action
  - `WH_CRYPTO_AFFINITY_SW` / `WH_CRYPTO_AFFINITY_HW` enum values
  - Request/response structures with translation functions

- **Server Handler** (`wh_server.c`, `wh_server.h`)
  - Added `configDevId` field to preserve the original configured device ID
  - Handler switches `devId` between `INVALID_DEVID` (SW) and `configDevId` (HW)

- **Tests** (`wh_test_clientserver.c`)
  - Added `_testCryptoAffinity()` to verify SW/HW switching behavior

- **Documentation** (`docs/draft/crypto_affinity.md`)
  - API usage guide with examples

### Usage

```c
int32_t server_rc;
uint32_t affinity;

// Switch to software crypto
wh_Client_SetCryptoAffinity(client, WH_CRYPTO_AFFINITY_SW, &server_rc, &affinity);

// Switch to hardware crypto (if configured)
wh_Client_SetCryptoAffinity(client, WH_CRYPTO_AFFINITY_HW, &server_rc, &affinity);
```

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -1134,14 +1134,11 @@ int wh_Client_AesCbcResponse(whClientContext* ctx, Aes* aes, uint8_t* out,
         ret = _getCryptoResponse(dataPtr, WC_CIPHER_AES_CBC, (uint8_t**)&res);
         if (ret == WH_ERROR_OK) {
             uint8_t* res_out = (uint8_t*)(res + 1);
+            uint8_t* res_iv  = res_out + res->sz;
             memcpy(out, res_out, res->sz);
-            /* For encryption, update the IV with the last ciphertext
-             * block for CBC chaining */
-            if (res->sz >= AES_BLOCK_SIZE) {
-                uint32_t last_offset =
-                    ((res->sz / AES_BLOCK_SIZE) - 1) * AES_BLOCK_SIZE;
-                memcpy((uint8_t*)aes->reg, out + last_offset, AES_IV_SIZE);
-            }
+            /* Update the IV from the server response for CBC chaining.
+             * The server provides the updated IV after the output data. */
+            memcpy((uint8_t*)aes->reg, res_iv, AES_IV_SIZE);
             if (out_size != NULL) {
                 *out_size = res->sz;
             }
--- a/test/wh_test_clientserver.c
+++ b/test/wh_test_clientserver.c
@@ -507,69 +507,6 @@ static int _clientServerSequentialTestConnectCb(void*           context,
                                   connected);
 }
 
-#if !defined(WOLFHSM_CFG_NO_CRYPTO)
-#ifdef HAVE_AES_CBC
-static int _testAesCbcRequestResponse(whClientContext* client,
-                                      whServerContext* server)
-{
-    Aes      aes[1];
-    uint8_t  key[AES_BLOCK_SIZE]       = {0xAA, 0xBB, 0xCC, 0xDD, 0xEE, 0xFF,
-                                          0x11, 0x22, 0x33, 0x44, 0x55, 0x66,
-                                          0x77, 0x88, 0x99, 0x00};
-    uint8_t  iv[AES_BLOCK_SIZE]        = {0};
-    uint8_t  plainIn[AES_BLOCK_SIZE]   = {0x77, 0x6F, 0x6C, 0x66, 0x48, 0x53,
-                                          0x4D, 0x20, 0x41, 0x45, 0x53, 0x20,
-                                          0x74, 0x65, 0x73, 0x74};
-    uint8_t  cipherOut[AES_BLOCK_SIZE] = {0};
-    uint8_t  plainOut[AES_BLOCK_SIZE]  = {0};
-    uint32_t outSize                   = 0;
-    int      rc;
-
-    WH_TEST_PRINT("  Testing AES CBC request/response...\n");
-
-    /* Encrypt */
-    WH_TEST_RETURN_ON_FAIL(wc_AesInit(aes, NULL, INVALID_DEVID));
-    WH_TEST_RETURN_ON_FAIL(
-        wc_AesSetKey(aes, key, sizeof(key), iv, AES_ENCRYPTION));
-
-    WH_TEST_RETURN_ON_FAIL(
-        wh_Client_AesCbcRequest(client, aes, 1, plainIn, sizeof(plainIn)));
-
-    /* Response should not be ready before server processes */
-    rc = wh_Client_AesCbcResponse(client, aes, cipherOut, &outSize);
-    WH_TEST_ASSERT_RETURN(rc == WH_ERROR_NOTREADY);
-
-    /* Server processes the request */
-    WH_TEST_RETURN_ON_FAIL(wh_Server_HandleRequestMessage(server));
-
-    /* Now response should be available */
-    WH_TEST_RETURN_ON_FAIL(
-        wh_Client_AesCbcResponse(client, aes, cipherOut, &outSize));
-    WH_TEST_ASSERT_RETURN(outSize == sizeof(plainIn));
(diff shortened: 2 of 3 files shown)


wolfhsm/wh_client.h
bigbrett: Just for flexibility, should we add a `GetCryptoAffinity` too?

src/wh_server.c
```diff
@@ -247,6 +268,56 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
...
+                    resp.rc = WH_ERROR_NOTIMPL;
+                    break;
+#endif
+                default:
+                    resp.rc = WH_ERROR_BADARGS;
+                    break;
+            }
+            resp.affinity = (server->crypto->devId == INVALID_DEVID)
+                                ? WH_CRYPTO_AFFINITY_SW
+                                : WH_CRYPTO_AFFINITY_HW;
```
bigbrett: Hmmm trying to figure out whether this is best, or instead whether the assignment of `affinity` in the response should  just be an unconditional pass-through from the request.

In this case, you echo back the value the server is actually using. So if you request hardware and there is no devId available, you would respond with "Software + error"

In the other case, you echo back the value the server is actually using on success, but on fail you would echo back the failing affinity value and the corresponding error code.

Not sure which I prefer, just musing...

src/wh_server.c
```diff
@@ -247,6 +268,56 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
...
+        }
+#else
+        resp.rc       = WH_ERROR_NOTIMPL;
+        resp.affinity = WH_CRYPTO_AFFINITY_SW;
+#endif
+
+        wh_MessageComm_TranslateSetCryptoAffinityResponse(
```
bigbrett: void-cast unused return value

src/wh_server.c
```diff
@@ -247,6 +268,56 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
         *out_resp_size = sizeof(resp);
     }; break;
 
+    case WH_MESSAGE_COMM_ACTION_SET_CRYPTO_AFFINITY: {
+        whMessageCommSetCryptoAffinityRequest  req  = {0};
+        whMessageCommSetCryptoAffinityResponse resp = {0};
+
+        wh_MessageComm_TranslateSetCryptoAffinityRequest(
```
bigbrett: void-cast unused return value

wolfhsm/wh_error.h
```diff
@@ -45,6 +45,7 @@ enum WH_ERROR_ENUM {
                                       compile-time configuration */
     WH_ERROR_USAGE =
         -2009, /* Operation not permitted based on object/key usage flags */
+    WH_ERROR_BADCONFIG = -2010, /* Failed due to runtime configuration */
```
bigbrett: as mentioned elsewhere, don't think we need this. `WH_ERROR_NOTIMPL` is sufficient.

wolfhsm/wh_server.h
```diff
@@ -67,9 +67,26 @@ typedef struct whServerContext_t whServerContext;
 
 typedef struct whServerCryptoContext {
     int devId;
+    int configDevId;
```
bigbrett: Do we think this naming, or `defaultDevId` conveys the meaning best? I'm fine with either, just curious your opinion.

bigbrett (changes requested):
Nice work. Some tweaks required and some deeper architectural questions.

src/wh_server.c
AlexLanzano: Hmmm, I think it might be confusing from a user stand point to have the resp.affinity be either "What the server is using" on success or "The requested affinity" on failure.

I think it's probably best to keep that consistent whether it's a failure or success.

test/wh_test_crypto_affinity.c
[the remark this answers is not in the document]
AlexLanzano: For readability and organization sake I believe it's best to keep this separate from wh_test_clientserver.c. From a user perspective it's immediately apparent where how this specific affinity feature is tested instead of having to reason about a generic test function.

wolfhsm/wh_message_comm.h
```diff
@@ -35,12 +35,18 @@
...
-    WH_MESSAGE_COMM_ACTION_CLOSE     = 0x03,
-    WH_MESSAGE_COMM_ACTION_INFO      = 0x04,
-    WH_MESSAGE_COMM_ACTION_ECHO      = 0x05,
+    WH_MESSAGE_COMM_ACTION_NONE                = 0x00,
+    WH_MESSAGE_COMM_ACTION_INIT                = 0x01,
+    WH_MESSAGE_COMM_ACTION_KEEPALIVE           = 0x02,
+    WH_MESSAGE_COMM_ACTION_CLOSE               = 0x03,
+    WH_MESSAGE_COMM_ACTION_INFO                = 0x04,
+    WH_MESSAGE_COMM_ACTION_ECHO                = 0x05,
+    WH_MESSAGE_COMM_ACTION_SET_CRYPTO_AFFINITY = 0x06,
```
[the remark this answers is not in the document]
AlexLanzano: Yeah, completely agree. We could potentially refactor the crypto message action to not be bound to wolfCrypt algo types so we could allow the affinity message to be handled there?

benchmark/wh_bench.c
```diff
@@ -1085,7 +1083,7 @@ int wh_Bench_ClientServer_Posix(int transport, int moduleIndex)
     }
 
     /* Initialize RNG */
-    ret = wc_InitRng_ex(crypto->rng, NULL, crypto->devId);
+    ret = wc_InitRng_ex(crypto->rng, NULL, INVALID_DEVID);
```
bigbrett: per our discussion we talked about removing the global devId, but I'm now realizing that we do still need it to exist for this use case. RNG is the one thing that does make sense to keep global, since initialization is expensive. I think we need to keep the global crypto->devId so the RNG can be initialized with it, but the server stack allocated crypto can instead hold its own "active" devId that is either `INVALID_DEVID` for SW affinity or use `crypto->devId` for HW affinity. Key point being, the devId is still globally held, used for RNG, but not otherwise able to have cross-client effects for crypto requested by a client

src/wh_server.c
```diff
@@ -247,6 +268,56 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
...
+        }
+#else
+        resp.rc       = WH_ERROR_NOTIMPL;
+        resp.affinity = WH_CRYPTO_AFFINITY_SW;
+#endif
+
+        wh_MessageComm_TranslateSetCryptoAffinityResponse(
+            magic, &resp, (whMessageCommSetCryptoAffinityResponse*)resp_packet);
+        *out_resp_size = sizeof(resp);
+    }; break;
```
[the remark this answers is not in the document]
bigbrett: not necessarily. Right now it is all managed internally. Use case being, a client connects, sets affinity to SW, does crypto, then disconnects. Currently the affinity state is still implicitly held in the server devId. When the client reconnects, the server context isn't reinitialized, it just sets the connection state in the server context so a polling loop can start processing requests

PR #282, opened 2026-02-05
Improvements for AES operations

Frauschi:
This PR reworks a large part of the AES-DMA code and adds DMA support for the three remaining AES modes:
- ECB
- CBC
- CTR

The existing AES-GCM DMA functionality has also been rewritten.

## General

For all modes, only the input and output data is transferred using DMA buffers (exception: for AES-GCM, the AAD data is also transferred via DMA). All other variable data like a key, IV or other internal data with a (semi) fixed and short length are transferred by value within the request/response messages. 

For all four modes, individual request/response message structures have been created for these AES DMA operations, as mapping them to the existing `whMessageCrypto_AesDmaRequest`/`whMessageCrypto_AesGcmDmaResponse` structures would increase complexity and also the number of transmitted bytes unnecessarily. 

In the non-DMA implementations, some minor bugs have been fixed and the error handling has been cleaned up regarding usage of `WH_ERROR_OK`.

## AES-ECB

As ECB does not an IV (each block is processed independently), only the key is transferred by value. Also, there is no internal AES state that has to be stored. 

## AES-CBC

Next to the key, an IV is transferred (IV is fixed block size length). This IV is also modified during the operation and has to be saved in the caller's AES structure for subsequent calls. Hence, an IV is transferred by value both in the request and in the response. This subsequent processing is also now tested in the unit tests.

## AES-CTR

CTR mode also uses an IV (fixed block size length), which is transferred in both the request and the response. However, CTR also supports input/output data that is **not** a multiple of the AES block size. In this case, the last partial block is stored in the AES structure as internal state for subsequent calls (in the `tmp` array). Hence, this state must also be transferred between client and server. Subsequent processing for both whole blocks and partial blocks is now also tested in the unit tests to ensure this state handling is working.

## AES-GCM

The AAD is also transferred via DMA, as this is variable length. Next to the variable length IV and the key, the authentication tag is also transferred by value in the request (for decryption) and the response (for encryption). 

## Benchmark

All AES modes with DMA are now executed in the benchmark in addition to the non-DMA versions. Furthermore, the module handling of the benchmarking application has been slightly reworked to get rid of the `MAX_BENCH_OPS` define.

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -1244,6 +1244,10 @@ int wh_Client_AesGcmDma(whClientContext* ctx, Aes* aes, int enc,
         req_len    += req->keySz;
     }
 
+    if (req_len > WOLFHSM_CFG_COMM_DATA_LEN) {
+        return WH_ERROR_BADARGS;
+    }
+
     /* Copy request data not handled by DMA */
     memcpy(req_iv, iv, iv_len);
     if (enc == 0 && tag_len > 0) {


wolfhsm/wh_client_crypto.h
```diff
@@ -600,16 +600,78 @@ int wh_Client_AesGetKeyId(Aes* key, whNvmId* outId);
 #ifdef WOLFSSL_AES_COUNTER
 int wh_Client_AesCtr(whClientContext* ctx, Aes* aes, int enc, const uint8_t* in,
                      uint32_t len, uint8_t* out);
+
+#ifdef WOLFHSM_CFG_DMA
```
bigbrett: could you remove the macro protection here?

In wolfHSM we try not to conditionally declare functions in headers, but just gate their definitions in the corresponding source file.

This rule was implemented halfway through library development, so as you can see in this file, it is not consistently applied. Please just remove the DMA macro protection you added for now, and we can clean up and make more uniform in a subsequent PR>

src/wh_message_crypto.c
```diff
@@ -1215,15 +1215,189 @@ int wh_MessageCrypto_TranslateAesDmaRequest(
     }
 
     WH_T32(magic, dest, src, enc);
-    WH_T32(magic, dest, src, type);
-    WH_T32(magic, dest, src, finalize);
     WH_T32(magic, dest, src, keyId);
 
-    ret = wh_MessageCrypto_TranslateDmaBuffer(magic, &src->state, &dest->state);
+    ret = wh_MessageCrypto_TranslateDmaBuffer(magic, &src->key, &dest->key);
+    if (ret != 0) {
```
bigbrett: I know you didn't write this original code, but since we are touching it, could you pls change the `ret != 0` to `ret != WH_ERROR_OK`?

We want all wolfHSM success values to be `WH_ERROR_OK`, which helps distinguish the intent from a wolfCrypt/wolfSSL error value (of which sometimes 0 is success, and sometimes not, depending on the API).

That said, since the DMA buffer translation can't fail when you control the arguments (already null checked), I'd be okay with you void-casting the return value here just to keep the function body smaller.

benchmark/wh_bench_ops.h
```diff
@@ -26,7 +26,7 @@
 #include <stdint.h>
 
 /* Maximum number of operations that can be registered */
-#define MAX_BENCH_OPS 89
+#define MAX_BENCH_OPS 103
```
bigbrett: 89 + 12 = 101?

Regardless, this really should be parameterized and not up to the developer to update each time...sigh

src/wh_server_crypto.c
```diff
@@ -2525,6 +2721,139 @@ static int _HandleAesEcb(whServerContext* ctx, uint16_t magic,
...
+    void*  inAddr  = NULL;
+    void*  outAddr = NULL;
+    word32 outSz   = 0;
+
+    whKeyId  keyId;
+    uint32_t keyLen;
+
+    (void)inSize;
+    (void)seq;
+
```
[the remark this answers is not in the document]
Frauschi: fixed

src/wh_client_crypto.c
```diff
@@ -545,6 +698,142 @@ int wh_Client_AesEcb(whClientContext* ctx, Aes* aes, int enc, const uint8_t* in,
...
+    if (ret == WH_ERROR_OK) {
+        uint16_t resLen = 0;
+        do {
+            ret =
+                wh_Client_RecvResponse(ctx, &group, &action, &resLen, dataPtr);
+        } while (ret == WH_ERROR_NOTREADY);
+
+        if (ret == WH_ERROR_OK) {
+            /* Get response */
+            whMessageCrypto_AesCbcDmaResponse* res;
```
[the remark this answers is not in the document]
Frauschi: fixed

Frauschi:
I fixed all the Copilot findings. The failing test is related to unrelated changes in wolfssl upstream.

src/wh_server_crypto.c
```diff
@@ -2620,6 +2975,168 @@ static int _HandleAesCbc(whServerContext* ctx, uint16_t magic, const void* crypt
...
+    }
+
+    wc_AesFree(aes);
+
+    /* Set response */
+    res.outSz = outSz;
+    *outSize  = sizeof(whMessageCrypto_AesCbcDmaResponse);
+
+    /* Translate response back */
+    ret = wh_MessageCrypto_TranslateAesCbcDmaResponse(
```
bigbrett: unconditional overwrite of `ret` here, causing it to be dropped by the caller. Recommend void casting this here, since we always want to translate the response, even on error, and translation is infallible if you control the inputs.

src/wh_server_crypto.c
```diff
@@ -2402,7 +2421,206 @@ static int _HandleAesCtr(whServerContext* ctx, uint16_t magic,
...
+        /* Copy key from client if provided */
+        ret = whServerDma_CopyFromClient(ctx, tmpKey, req.key.addr, keyLen,
+                                         (whServerDmaFlags){0});
+        if (ret != WH_ERROR_OK) {
+            res.dmaAddrStatus.badAddr = req.key;
+        }
+    }
+
+    /* Handle input data */
+    if (ret == WH_ERROR_OK && req.input.sz > 0) {
```
bigbrett: should we also check input and output sizes are multiples of `WC_AES_BLOCK_SIZE` like we do for non-DMA paths? Comment applies to all new DMA functions added in this file.

Consider sanity checks on all DMA input data sizes if we know what they should be.

src/wh_server_crypto.c
Frauschi: fixed

src/wh_server_crypto.c
Frauschi: For AES-ECB and AES-CBC, the length checks are integrated on the client-side in wolfHSM as well as in the native wolfCrypt AES calls. So this is already handled. 

For AES-GCM and AES-CTR, the data length may also **not** be multiple of the block size. For AES-GCM, padding is handled internally, for AES-CTR, partial blocks are buffered internally to be processed by a consecutive call to encrypt/decrypt. This is also properly supported in our implementation now. I also added a unit test to check this behavior.

src/wh_server_crypto.c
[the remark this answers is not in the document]
Frauschi: Added

wolfhsm/wh_message_crypto.h
```diff
@@ -990,36 +993,111 @@ int wh_MessageCrypto_TranslateCmacDmaResponse(
...
+
+int wh_MessageCrypto_TranslateAesEcbDmaResponse(
+    uint16_t magic, const whMessageCrypto_AesEcbDmaResponse* src,
+    whMessageCrypto_AesEcbDmaResponse* dest);
+
+/* AES-CBC DMA Request */
+typedef struct {
+    uint32_t                  enc;      /* 1 for encrypt, 0 for decrypt */
+    uint32_t                  keyId;
+    whMessageCrypto_DmaBuffer key;      /* Key buffer */
```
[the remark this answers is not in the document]
Frauschi: Modified the DMA behavior for all four AES modes to only transfer input and output via DMA. Key (when present), IV etc. are now passed by value within the message.

src/wh_client_crypto.c
```diff
@@ -644,20 +900,159 @@ int wh_Client_AesCbc(whClientContext* ctx, Aes* aes, int enc, const uint8_t* in,
...
+            }
+        }
+    }
+    return ret;
+}
+
+int wh_Client_AesCbcDma(whClientContext* ctx, Aes* aes, int enc,
+                        const uint8_t* in, uint32_t len, uint8_t* out)
+{
+#ifdef WOLFHSM_CFG_DMA
```
bigbrett: same comment as above re: conditional compilation of function body

src/wh_client_crypto.c
```diff
@@ -802,30 +1197,25 @@ int wh_Client_AesGcm(whClientContext* ctx, Aes* aes, int enc, const uint8_t* in,
...
-    int                            ret         = WH_ERROR_OK;
-    whMessageCrypto_AesDmaRequest* req         = NULL;
-    uint8_t*                       dataPtr     = NULL;
-    uintptr_t                      inAddr      = 0;
-    uintptr_t                      outAddr     = 0;
-    uintptr_t                      keyAddr     = 0;
-    uintptr_t                      ivAddr      = 0;
-    uintptr_t                      aadAddr     = 0;
-    uintptr_t                      authTagAddr = 0;
+#ifdef WOLFHSM_CFG_DMA
```
bigbrett: same comment as above re: declaration of function body

src/wh_message_crypto.c
```diff
@@ -1212,80 +1212,153 @@ int wh_MessageCrypto_TranslateEd25519VerifyDmaResponse(
...
-    }
+    (void)wh_MessageCrypto_TranslateDmaBuffer(magic, &src->input, &dest->input);
+    (void)wh_MessageCrypto_TranslateDmaBuffer(magic, &src->output,
+                                              &dest->output);
 
-    ret = wh_MessageCrypto_TranslateDmaBuffer(magic, &src->key, &dest->key);
-    if (ret != 0) {
-        return ret;
-    }
+    return 0;
```
bigbrett: nit: `WH_ERROR_OK` (and same for all functions in this file)

src/wh_server_crypto.c
```diff
@@ -2402,7 +2404,182 @@ static int _HandleAesCtr(whServerContext* ctx, uint16_t magic,
...
+            }
+            if (ret == WH_ERROR_OK) {
+                key    = cachedKey;
+                keyLen = keyMeta->len;
+                WH_DEBUG_VERBOSE_HEXDUMP("[AesCtr] Key from HSM", key, keyLen);
+            }
+        }
+    }
+    /* Verify key size is valid for AES */
+    if (keyLen != AES_128_KEY_SIZE && keyLen != AES_192_KEY_SIZE &&
```
bigbrett: this should probably be checked earlier in the function before any keystore lookups occur? And would `WH_ERROR_BADARGS` be more appropriate here?

This comment applies to all the new DMA handlers as the same logic exists in all of them.

src/wh_client_crypto.c
```diff
@@ -433,19 +433,164 @@ int wh_Client_AesCtr(whClientContext* ctx, Aes* aes, int enc, const uint8_t* in,
...
             }
         }
     }
     return ret;
 }
+
+int wh_Client_AesCtrDma(whClientContext* ctx, Aes* aes, int enc,
+                        const uint8_t* in, uint32_t len, uint8_t* out)
+{
+#ifdef WOLFHSM_CFG_DMA
```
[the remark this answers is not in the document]
Frauschi: Changed, makes sense to handle it that way. Based on your previous comment to remove the gate in the header, I just thought that the goal is to prevent compilation errors.

src/wh_client_crypto.c
Frauschi: changed

src/wh_client_crypto.c
Frauschi: changed

src/wh_message_crypto.c
Frauschi: Fixed for all functions I created within this PR.

src/wh_server_crypto.c
bigbrett: OK, should this be protected by `WH_ERROR_OK` then? Specifically I don't want a keystore failure or lookup to be "masked" by a badargs here. We should return the first error that occurs

src/wh_server_crypto.c
```diff
@@ -2402,13 +2404,189 @@ static int _HandleAesCtr(whServerContext* ctx, uint16_t magic,
...
+        magic, (whMessageCrypto_AesCtrDmaRequest*)cryptoDataIn, &req);
+    if (ret != WH_ERROR_OK) {
+        return ret;
+    }
+
+    uint32_t enc         = req.enc;
+    uint32_t keyLen      = req.keySz;
+    uint32_t len         = req.input.sz;
+    uint32_t left        = req.left;
+    uint16_t needed_size = sizeof(whMessageCrypto_AesCtrDmaRequest) + keyLen +
```
bigbrett: I think there could be a "bug" here. Quotes because you would need to manually craft a packet to do this, not sure it is reachable through the client API or wolfCrypt, but still worth fixing.

`needed_size` needs to be `uint64_t` (like the non-DMA handlers). `req.keyLen` is a `uint32_t` so the addition can overflow `uint16_t` and bypass this bounds check. Pretty sure that no crypto is actually affected since the  later `keyLen` length check against AES sizes catches this but I believe it still could cause `WH_DEBUG_VERBOSE_HEXDUMP` to do an out-of-bounds read. Same issue in all the other handlers. I would just change these to be safe.

src/wh_client_crypto.c
```diff
@@ -963,54 +1302,42 @@ int wh_Client_AesGcmDma(whClientContext* ctx, Aes* aes, int enc,
...
              * scenarios */
-            if (ret >= 0) {
-                /* For DMA operations, data is already in client memory,
-                 * no need to copy it back */
-                ret = 0; /* Success */
+            if (ret >= WH_ERROR_OK) {
+                if (enc != 0 && res->authTagSz > 0) {
+                    uint8_t* res_tag = (uint8_t*)res +
+                                       sizeof(whMessageCrypto_AesGcmDmaResponse);
+                    memcpy(enc_tag, res_tag, tag_len);
```
bigbrett: Doesn't look like we ever check `enc_tag != NULL` here, or clamp to tag_len, like the non-DMA wh_Client_AesGcm path does. So `tag_len > 0` and `enc_tag == NULL` would lead to a null dereference.

I'd check server side too to make sure we are similarly robust

Frauschi:
Added all the missing `READ_POST` operations (unconditionally).

src/wh_server_crypto.c
```diff
@@ -2402,13 +2409,202 @@ static int _HandleAesCtr(whServerContext* ctx, uint16_t magic,
...
+    }
+
+    /* Translate request */
+    ret = wh_MessageCrypto_TranslateAesCtrDmaRequest(
+        magic, (whMessageCrypto_AesCtrDmaRequest*)cryptoDataIn, &req);
+    if (ret != WH_ERROR_OK) {
+        return ret;
+    }
+
+    uint32_t enc         = req.enc;
```
JacobBarthelmeh: Preference in wolfSSL master is all variables declared at the top of the scope. Some old compilers used with Windows VS could not handle variables not declared at the top of the scope.

JacobBarthelmeh (approved):
Thanks! I had one comment about variable declaration placement but that could be addressed in a follow up PR if wanted.

PR #290, opened 2026-02-20
Adds NVM for storing users with authentication feature

JacobBarthelmeh:
Builds on top of (https://github.com/wolfSSL/wolfHSM/pull/270) with adding NVM storage for users.

### DIFF
--- a/docs/draft/auth.md
+++ b/docs/draft/auth.md
@@ -11,12 +11,12 @@ PR #270 introduces a **PKCS11-flavored Authentication/Authorization Manager** to
 - A **permission model** of (admin flag) + (per-group allow boolean) + (per-group bitmap of 256 allowed actions) + (a small per-user list of accessible key IDs — not yet wired into crypto paths).
 - A **server-side request gate** that, on every incoming request, consults the Auth Manager and rejects messages the current session is not permitted to run.
 - A **message group** `WH_MESSAGE_GROUP_AUTH = 0x0D00` with 7 new actions (login, logout, user add/delete/get, set-permissions, set-credentials), complete with endian/magic translation functions.
-- A **pluggable backend**: everything goes through a `whAuthCb` callback vtable. A default in-memory backend lives in `src/wh_auth_base.c` (up to 5 users, credential storage up to 2 KiB per user, used by examples and tests).
+- A **pluggable backend**: everything goes through a `whAuthCb` callback vtable. A default backend lives in `src/wh_auth_base.c` (up to 5 users, credential storage up to 2 KiB per user, optional NVM persistence with credentials split into per-user objects, used by examples and tests).
 - The feature is **opt-in**: entire subsystem is guarded by `WOLFHSM_CFG_ENABLE_AUTHENTICATION`. With it compiled in but no context configured (`server->auth == NULL`), the server logs a SECEVENT and processes all requests without any authorization check — preserving backwards compatibility.
 
 Design notes called out by the author:
 - The "check key use" callback (`CheckKeyAuthorization`) is wired into the interface but **not yet invoked** on the key paths — it's a TODO placeholder.
-- The base user list is in RAM, not NVM — deliberate for an initial cut.
+- The base user list lives in a RAM index and is optionally persisted to NVM (config `whAuthBaseConfig.nvm`; NULL = RAM-only). Credentials are split into per-user NVM objects, and a corrupt/incompatible persisted database is a fatal init error; see [6.1 Persistence (NVM)](#61-persistence-nvm).
 - Logging of auth events (login success/failure, crypto actions) is another TODO, though authorization failures already log via `WH_LOG_ON_ERROR_F`.
 
 ---
@@ -57,7 +57,7 @@ Design notes called out by the author:
              ▼
      ┌─────────────────────────┐
      │   Pluggable backend     │
-     │   default: wh_auth_base │   in-memory user db, SHA-256 PIN hashing,
+     │   default: wh_auth_base │   user db (RAM index + optional NVM), SHA-256 PIN hashing,
      │   (src/wh_auth_base.c)  │   optional wolfSSL cert verification
      └─────────────────────────┘
 ```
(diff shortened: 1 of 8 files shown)


src/wh_auth_base.c
```diff
@@ -390,16 +838,40 @@ int wh_Auth_BaseUserDelete(void* context, uint16_t current_user_id,
...
+    if (rc != WH_ERROR_OK) {
+        memcpy(user, &s_auth_base_backup, sizeof(whAuthBase_User));
+    }
+    else if (s_auth_base_backup.credentials_len > 0) {
+        /* A lingering blob could resurrect the credential if the slot is reused,
+         * so a destroy failure is fatal: roll the deletion back. */
+        rc = wh_Auth_BaseDestroyCred(user_id);
+        if (rc != WH_ERROR_OK) {
+            memcpy(user, &s_auth_base_backup, sizeof(whAuthBase_User));
+            (void)wh_Auth_BasePersistIndex();
```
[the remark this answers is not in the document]
JacobBarthelmeh: Reviewed this, I think it is not necessary since rc is already returning na error. Basically if the persist index call fails too it's in a non recoverable state.

JacobBarthelmeh:
@bigbrett revived this and got it to the point where I think it's ready for review. It adds in a default NVM option for user storage, and a nice addition of separating user meta data from user credentials stored.

docs/draft/auth.md
bigbrett, line 1: please update the actual docs now. There is a whole section on auth in 5-Features.md. I should probably remove `docs/draft.md`

docs/draft/auth.md
```diff
@@ -253,14 +253,48 @@ The two "Check*" callbacks are **overrides, not gates** — see §7 below for ex
...
+- NVM configured (`cfg.nvm != NULL`): the index is loaded at init; the index and the affected credential object are rewritten on every change.
+
+> Note on `WOLFHSM_CFG_AUTH_BASE_NVM_ONLY` (RAM savings): the RAM-only mode requires a static fallback table sized to hold every user's full credential blob at once, `s_auth_base_ram_cred[WH_AUTH_BASE_MAX_USERS][WH_AUTH_BASE_MAX_CREDENTIALS_LEN]`, i.e. `5 x 2048 = 10 KiB` at the defaults. On a memory-constrained target that always configures an NVM backend, define `WOLFHSM_CFG_AUTH_BASE_NVM_ONLY` to compile that table out entirely (and drop support for the `nvm == NULL` configuration). Credentials then only ever occupy the single shared `s_auth_base_cred_buf` scratch buffer (`WH_AUTH_BASE_MAX_CREDENTIALS_LEN` = 2 KiB, force-zeroed after each use) while one credential is being read or written, rather than a per-user resident copy. That reclaims roughly `(MAX_USERS - 1) x MAX_CREDENTIALS_LEN` of RAM.
```
bigbrett, line 284: Is there any advantage to having RAM + NVM? Or even the pure RAM version at all? I def want to keep it flexible, so perhaps no action required, but just thinking out loud.

RAM only is probably good to keep around only because we do have some platforms without dedicated HSM NVM object storage, however we would want the capability for the DB objects to be "wrapped" for it to work in this case, so they can be stored client-side encrypted. But this is WAY beyond the scope of this PR so dont' worry about that for now.

However do we think that if you have NVM backing, is there a reason you would ever need the fallback table at all? Might as well save space if possible?

bigbrett (approved):
Overall, this looks good. I found a few cases where a client could still erase or replace credentials through another path, especially the certificate API. That API intentionally allows some protected objects to be updated so trust anchors can be rotated.

This branch has fallen behind `main`, which now includes changes that can help close these gaps. I recommend the following as a follow up PR

Since this is an "experimental" feature I'm going to approve/merge now, but leaving this for follow up:


1. **Rebase onto `main`.**

   This will bring in the `WH_NVM_FLAGS_SERVER_ONLY` mask. Client-provided metadata is already stripped of these flags by `wh_Nvm_AddObjectChecked`, and `wh_Nvm_CheckPolicy` rejects them for add, destroy, and read operations.

2. **Add a new `WH_NVM_FLAGS_SERVER_INTERNAL` flag using bit 13, and include it in `WH_NVM_FLAGS_SERVER_ONLY`.**

   Also update these checks that this issue made me realize I missed:

   * In `wh_Nvm_CheckPolicy`, check the full `SERVER_ONLY` mask instead of only `TRUSTED`. The "server only" bits being set should indicate that a client can never modify the object 
   * Make the same change to the keystore freeze check at `wh_server_keystore.c:215`.
   * Keep the key-encryption-key check at `wh_server_keystore.c:1484` limited to `TRUSTED`.

   The other checks already use the full mask, so they should handle the new flag automatically.

3. **Mark auth objects created by the server as `SERVER_INTERNAL`.**

   Add the flag where these objects are created in `wh_auth_base.c:154` and `wh_auth_base.c:206`.

   This prevents client-facing APIs, including the certificate API, from modifying or replacing these objects through another object type.

   It should not interfere with the server’s own work because its writes and deletes use the unchecked backend functions: `wh_Nvm_AddObjectWithReclaim` and `wh_Nvm_DestroyObjects`.

4. **Require `SERVER_INTERNAL` when loading credentials.**

   Add this check to both `wh_Auth_BaseLoadCredHeader` and `wh_Auth_BaseLoadFromNvm`, alongside the existing length and magic-value checks.

   This is important because the flags only protect objects that already exist. If an object does not exist, `wh_Nvm_CheckPolicy` returns after `GetMetadata` reports `NOTFOUND`.

   A client could therefore create an object using an unused reserved ID, such as `0xFE04` when only three users exist, or use the slot of a user that was created without credentials.

   On the next boot, the loader could treat that object as a valid credential because it currently reads the method from `meta.label[23]` and the length from `meta.len` without checking where the object came from.

   Clients cannot set `SERVER_INTERNAL`, so requiring that flag during loading confirms that the credential was created by the server.

docs/draft/auth.md
JacobBarthelmeh: No advantage to both on at the same time, could be just one or the other. I like keeping the RAM around though for its simplicity and possibly trust zone use or in simple cases where no NVM is available. Wrapping the RAM would be neat, could do something similar to the PKCS11 token wrapping here (https://github.com/wolfSSL/wolfPKCS11/pull/159) where a devices unique key is used to encrypt the RAM when "saved" and "restored".

PR #293, opened 2026-02-27
Address chunk size issue in wh_Client_SheSecureBoot

padelsbach:
Fixes finding 146 and adds associated regression test. Also, this adjusts a buffer size constant in the test code.

### DIFF
--- a/test/wh_test_she.c
+++ b/test/wh_test_she.c
@@ -74,11 +74,11 @@
 #endif
 
 enum {
-        REQ_SIZE = 32,
-        RESP_SIZE = 64,
-        BUFFER_SIZE = sizeof(whTransportMemCsr) + sizeof(whCommHeader) +
-                      WOLFHSM_CFG_COMM_DATA_LEN,
-    };
+    REQ_SIZE    = 32,
+    RESP_SIZE   = 64,
+    BUFFER_SIZE = sizeof(whTransportMemCsr) + sizeof(whCommHeader) +
+                  WOLFHSM_CFG_COMM_DATA_LEN,
+};
 
 #define FLASH_RAM_SIZE (1024 * 1024) /* 1MB */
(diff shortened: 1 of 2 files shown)


test/wh_test_she.c
```diff
@@ -74,7 +75,8 @@
 enum {
         REQ_SIZE = 32,
         RESP_SIZE = 64,
-        BUFFER_SIZE = 4096,
+        BUFFER_SIZE = sizeof(whTransportMemCsr) + sizeof(whCommHeader) +
```
padelsbach: `BUFFER_SIZE` was too small and causing a failure/hang due to overflow in build+test workflow in CI. This increases it to 5120.

src/wh_server_she.c
```diff
@@ -232,13 +232,18 @@ static int _SecureBootInit(whServerContext* server, uint16_t magic,
 
     (void)wh_MessageShe_TranslateSecureBootInitRequest(magic, req_packet, &req);
 
-    /* if we aren't looking for init return error */
-    if (server->she->sbState != WH_SHE_SB_INIT) {
+    /* Allow a new secure-boot session after a completed prior session.
```
bigbrett: As you alluded to in your PR description, the SHE spec unfortunately prohibits this. Technically the failure should be latched until a POR.

IIRC we have battled this before, and are "mostly" compliant with the SHE spec but since it is a software simulation we are somewhat in a grey area.

If there are "auxillary" SHE functions we need in order to properly and effectively test the layer, I propse adding the necessary functions behind a new config macro, enabling the library to be built for unit test compatibility with the new auxillary functions defined. SOmething like `WOLFHSM_CFG_SHE_TEST_API` or something similar.

We can also punt this if you want a wolfHSM developer to pick this task up

padelsbach: Makes sense. I'll investigate a bit further

PR #301, opened 2026-03-14
Fix Dilithium issue following WolfCrypt update

padelsbach:
Address updates in WolfCrypt regarding `wc_MlDsaKey_*` and new fields in `wc_CryptoInfo`.


bigbrett (changes requested):
Thanks for this. 


Can you just expose the new functionality and make it default? Don't worry about backwards compat for this API?

PR #310, opened 2026-03-18
Misc fixes in SHE server

padelsbach:
Fixes findings 458, 476, 768

### DIFF
--- a/src/wh_server_she.c
+++ b/src/wh_server_she.c
@@ -787,7 +787,7 @@ static int _ExportRamKey(whServerContext* server, uint16_t magic,
         /* set the counter, flags and ram key */
         memset(resp.messageTwo, 0, sizeof(resp.messageTwo));
         /* set count to 1 */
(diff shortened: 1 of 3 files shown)


bigbrett:
@padelsbach can you resolve conflicts plz

src/wh_server_she.c
```diff
@@ -1523,6 +1523,10 @@ static int _VerifyMac(whServerContext* server, uint16_t magic,
         }
         else {
             mac = message + req.messageLen;
+            /* Verify mac is allowed to fail, per SHE spec.
+               Capture status, but return success. */
+            resp.status = 1;
+            ret = 0;
```
bigbrett: I'm confused as to the intent of these lines - the comment makes it seem like you intended to have these lines after the mac verification, but it is before? What is going on here? Is the intent simply to ensure `resp.status` is initialized to 1 (failure)?

padelsbach: Fixed

bigbrett (approved):

PR #318, opened 2026-03-20
add support for wrapped certs

### LINKED ISSUE #312: Bug: wh_server_cert calls don't support local cache
The issue:
This is issue is widespread throughout wh_server_cert.c. I'm going to walk through one function but search for "server->nvm,". for other instances.
wh_Server_CertReadTrusted() needs to get a root certificate out of NVM. If key wrapping is enabled, this is only available to the server via the cache. In these builds the server will/may have two caches, a global one (accessible via server->nvm->globalCache) and a local one (accessible by server->localCache. wh_Server_CertReadTrusted() calls wh_Nvm_GetMetadata and wh_Nvm_Read on server->nvm. If the key wrap is using a local cache (the safer and default option) these calls will obviously fail as the call has no access to a local cache.

The fix:
This suggestion should obviously be verified for architectural correctness and unintended security side effects before adoption. The certs can be accessed via wh_Server_KeystoreReadKey and other general keystore calls which correctly handle caches and reduces code duplication. Alternatively, wh_Server_CertReadTrusted and other calls in the certificate manager can be rewritten to use logic more similar to that in the general keystore and accept the entire server object as a parameter.

bigbrett:
Allows clients to use trusted root certificates that are AES-GCM wrapped and cached in server RAM rather than committed to NVM. The client stores the wrapped blob cheaply and unwraps it on demand for verification.

Fixes #312 

## Changes                                                                                                                                                                                                                                                                                                                              
  - Add client-side certificate wrap/unwrap/cache API (wh_Client_CertWrap, wh_Client_CertUnwrapAndExport, wh_Client_CertUnwrapAndCache) as convenience wrappers around the existing keywrap functions, guarded by WOLFHSM_CFG_KEYWRAP                                                                                         
  - Modify server cert layer (wh_Server_CertReadTrusted, wh_Server_CertVerify, wh_Server_CertVerifyAcert) to support routing to keystore cache when the key type is WH_KEYTYPE_WRAPPED 
  - Update all server request handlers (READTRUSTED, VERIFY, READTRUSTED_DMA, VERIFY_DMA, VERIFY_ACERT, VERIFY_ACERT_DMA) to translate the WH_KEYID_CLIENT_WRAPPED_FLAG and branch between cache and NVM paths                                                                                                                
  - Add tests for cache-based cert verify, ReadTrusted, and DMA variants

### DIFF
--- a/docs/draft/wrapped-certs.md
+++ b/docs/draft/wrapped-certs.md
@@ -4,7 +4,7 @@
 
 This documents the "wrapped certificate" use case, showing how to leverage the
 certificate manager to use trusted root certificates that live in the server's
-**keystore cache** (RAM) after being unwrapped via keywrap funcitonality,
+**keystore cache** (RAM) after being unwrapped via keywrap functionality,
 rather than exclusively in **NVM** (flash). A root certificate is wrapped
 (AES-GCM encrypted) by the server, handed back to the client as an opaque blob,
 and later unwrapped into the server's key cache on demand. Once cached, it can
@@ -196,7 +196,7 @@ updated with the same pattern:
 
 1. **Translate the client ID**: If the incoming `req.id` (or
    `req.trustedRootNvmId`) has `WH_KEYID_CLIENT_WRAPPED_FLAG` set, call
-   `wh_KeyId_TranslateFromClient(WH_KEYTYPE_NVM, server->comm->client_id, req.id)`
+   `wh_KeyId_TranslateFromClient(WH_KEYTYPE_WRAPPED, server->comm->client_id, req.id)`
    to produce a full server-internal key ID with `TYPE=WH_KEYTYPE_WRAPPED`,
    `USER=client_id`, and the bare key `ID` in the low byte.
 
--- a/docs/draft/wrapped-certs.md
+++ b/docs/draft/wrapped-certs.md
@@ -35,6 +35,10 @@ wh_Client_KeyCache(client,
                    kek, sizeof(kek), &kekId);
 ```
 
+> **Note:** This is an example only. In production, the KEK is normally
+> provisioned and protected by target-specific hardware and should never be
+> hardcoded.
+
 The KEK is now sitting in the server's `localCache` (or `globalCache` if marked
 global), indexed by `kekId`.
 
--- a/docs/draft/wrapped-certs.md
+++ b/docs/draft/wrapped-certs.md
@@ -1,0 +1,290 @@
+# Wrapped Cached Certificates
+
+## Overview
+
+This documents the "wrapped certificate" use case, showing how to leverage the
+certificate manager to use trusted root certificates that live in the server's
+**keystore cache** (RAM) after being unwrapped via keywrap funcitonality,
+rather than exclusively in **NVM** (flash). A root certificate is wrapped
+(AES-GCM encrypted) by the server, handed back to the client as an opaque blob,
+and later unwrapped into the server's key cache on demand. Once cached, it can
+be used in all certificate verification paths — standard, DMA, and ACERT —
+exactly like an NVM-resident root certificate.
+
+This is useful when a client needs to use a trusted root for verification but
+does not want to (or cannot) commit it to NVM. The wrapped blob can be stored
+cheaply on the client side, while the server only holds the unwrapped plaintext
+in its volatile cache for as long as it is needed.
+
+## High-Level Usage
+
+The lifecycle has three stages: **wrap**, **unwrap-and-cache**, and **use**.
+
+### 1. Provision a wrapping key (KEK)
+
+Before wrapping anything the server needs an AES-256 key to use as the
+key-encryption key. Cache it on the server with the `WH_NVM_FLAGS_USAGE_WRAP`
+flag:
+
+```c
+whKeyId kekId = 10;
+uint8_t kek[32] = { /* 256-bit AES key */ };
+
+wh_Client_KeyCache(client,
+                   WH_NVM_FLAGS_USAGE_WRAP, NULL, 0,
+                   kek, sizeof(kek), &kekId);
+```
+
+The KEK is now sitting in the server's `localCache` (or `globalCache` if marked
+global), indexed by `kekId`.
+
+### 2. Wrap the certificate
+
+Call `wh_Client_CertWrap` with the raw certificate DER and the KEK's ID. The
+server encrypts the certificate using AES-GCM and returns the wrapped blob:
+
+```c
+uint8_t  wrappedCert[2048];
+uint16_t wrappedCertSz = sizeof(wrappedCert);
+
+/* Build metadata: id embeds TYPE=WRAPPED and the client's USER id;
+ * caller controls flags, access, and optionally label */
+whNvmMetadata certMeta = {0};
+certMeta.id     = WH_CLIENT_KEYID_MAKE_WRAPPED_META(
+    client->comm->client_id, 5);
+certMeta.flags  = WH_NVM_FLAGS_USAGE_ANY;
+certMeta.access = WH_NVM_ACCESS_ANY;
+
+wh_Client_CertWrap(client, WC_CIPHER_AES_GCM, kekId,
+                    rootCaCert, rootCaCertLen,
+                    &certMeta,
+                    wrappedCert, &wrappedCertSz);
+```
+
+After this call:
+
+| Data | Location |
+|---|---|
+| KEK | Server key cache (`localCache[kekId]`) |
+| Wrapped cert blob (ciphertext + GCM tag + IV + metadata) | Client memory (`wrappedCert` buffer) |
+| Raw certificate | Nowhere on the server — only the client supplied it transiently |
+
+The client can now persist `wrappedCert` to its own storage (file, flash,
+external memory, etc.).
+
+### 3. Unwrap and cache the certificate on the server
+
+When the client needs the root for verification, it pushes the wrapped blob back
+to the server:
+
+```c
+whKeyId cachedCertId = WH_KEYID_ERASED;
+
+wh_Client_CertUnwrapAndCache(client, WC_CIPHER_AES_GCM, kekId,
+                              wrappedCert, wrappedCertSz,
+                              &cachedCertId);
+```
+
+The server decrypts the blob using the KEK, verifies the GCM authentication
+tag, and places the plaintext certificate into its key cache. The returned
+`cachedCertId` is the server-internal key ID (with `TYPE=WH_KEYTYPE_WRAPPED`
+already encoded).
+
+After this call:
+
+| Data | Location |
+|---|---|
+| KEK | Server key cache |
+| Plaintext certificate | Server key cache (`localCache[cachedCertId]`) |
+| Wrapped cert blob | Still in client memory (unchanged) |
+
+### 4. Use the cached cert for verification
+
+Pass the cached cert's ID — decorated with the wrapped flag — as the trusted
+root to any verify API:
+
+```c
+int32_t verifyResult;
+
+wh_Client_CertVerify(client,
+                      intermediateCert, intermediateCertLen,
+                      WH_CLIENT_KEYID_MAKE_WRAPPED(cachedCertId),
+                      &verifyResult);
+```
+
+`WH_CLIENT_KEYID_MAKE_WRAPPED(cachedCertId)` sets bit 9
+(`WH_KEYID_CLIENT_WRAPPED_FLAG = 0x0200`) on the ID the client sends to the
+server. This is the signal that tells the server "this root cert is in the
+cache, not in NVM."
+
+The same pattern works for:
+- `wh_Client_CertVerifyDma` (DMA path)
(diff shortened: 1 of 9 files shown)


src/wh_server_cert.c
```diff
@@ -238,6 +238,21 @@ int wh_Server_CertReadTrusted(whServerContext* server, whNvmId id,
...
+            *inout_cert_len = sz;
+        }
+        else if (rc == WH_ERROR_NOSPACE) {
+            /* Map keystore error to cert API's documented error and report the
+             * required size so the caller can retry with a larger buffer */
+            *inout_cert_len = meta.len;
+            rc              = WH_ERROR_BUFFER_SIZE;
+        }
+        return rc;
+    }
```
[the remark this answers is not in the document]
bigbrett: this is how the code originally worked. Any raw NVM type should be useable for now and fill fail up the stack

src/wh_server_cert.c
```diff
@@ -473,21 +488,45 @@ int wh_Server_HandleCertRequest(whServerContext* server, uint16_t magic,
...
+                    rc = wh_Server_KeystoreReadKey(server, certId, &meta, NULL,
+                                                   &cert_len);
+                    if (rc == WH_ERROR_OK) {
+                        if (meta.flags & WH_NVM_FLAGS_NONEXPORTABLE) {
+                            rc = WH_ERROR_ACCESS;
+                        }
+                        else {
+                            cert_len = buf_capacity;
+                            rc = wh_Server_KeystoreReadKey(
+                                server, certId, NULL, cert_data, &cert_len);
```
AlexLanzano, line 507: This is fine for now, but it would be a good idea to find a way to cleanly do a single wh_Server_KeystoreReadKey call to cut down on the cache lookup.

docs/draft/wrapped-certs.md
```diff
@@ -0,0 +1,290 @@
...
+
+## High-Level Usage
+
+The lifecycle has three stages: **wrap**, **unwrap-and-cache**, and **use**.
+
+### 1. Provision a wrapping key (KEK)
+
+Before wrapping anything the server needs an AES-256 key to use as the
+key-encryption key. Cache it on the server with the `WH_NVM_FLAGS_USAGE_WRAP`
+flag:
```
billphipps, line 27: Maybe add a note like:
This is an example ONLY.  The "real" KEK is normally securely handled by target-specific hardware.

docs/draft/wrapped-certs.md
```diff
@@ -0,0 +1,290 @@
...
+
+All functions accept an `enum wc_CipherType cipherType` parameter (e.g.
+`WC_CIPHER_AES_GCM`) to specify the wrapping cipher. The blocking variants
+call their respective Request/Response functions in a do-while-NOTREADY loop.
+
+These are pure convenience; a caller could use `wh_Client_KeyWrap*` /
+`wh_Client_KeyUnwrapAndExport*` / `wh_Client_KeyUnwrapAndCache*` directly if
+it needed custom metadata.
+
+### Server-side routing (the key change)
```
billphipps, line 173: The "key change"?  Boo!  Hiss!  Terrible joke.  :)

src/wh_client_cert.c
```diff
@@ -958,6 +958,171 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
 
 #endif /* WOLFHSM_CFG_CERTIFICATE_MANAGER_ACERT */
 
+#ifdef WOLFHSM_CFG_KEYWRAP
```
billphipps, line 957: I assume we are also in a !NO_CRYPTO area?

src/wh_client_cert.c
```diff
@@ -958,6 +958,171 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
...
+
+int wh_Client_CertWrapRequest(whClientContext*   ctx,
+                              enum wc_CipherType cipherType,
+                              whKeyId serverKeyId, const uint8_t* cert,
+                              uint32_t certSz, whNvmMetadata* meta)
+{
+    if ((ctx == NULL) || (cert == NULL) || (certSz == 0) || (meta == NULL) ||
+        (certSz > UINT16_MAX)) {
```
billphipps: Do we have a different max limit for cert size?  uint16max would certainly overflow when trying to wrap it.  This may be limited by the MTU size for non-DMA.

src/wh_server_cert.c
```diff
@@ -522,12 +561,19 @@ int wh_Server_HandleCertRequest(whServerContext* server, uint16_t magic,
                 whKeyId keyId = wh_KeyId_TranslateFromClient(
                     WH_KEYTYPE_CRYPTO, server->comm->client_id, req.keyId);
 
+                /* Translate trustedRootNvmId if client set wrapped flag */
+                whKeyId trustedRootId = req.trustedRootNvmId;
```
AlexLanzano: `req.trustedRootNvmId` should be renamed to `req.trustedRootId`

src/wh_server_cert.c
```diff
@@ -612,19 +658,46 @@ int wh_Server_HandleCertRequest(whServerContext* server, uint16_t magic,
...
+                         * reading cert data into client DMA buffer */
+                        whKeyId certId = wh_KeyId_TranslateFromClient(
+                            WH_KEYTYPE_WRAPPED, server->comm->client_id,
+                            req.id);
+                        cert_len = req.cert_len;
+                        resp.rc  = wh_Server_KeystoreReadKey(
+                            server, certId, &meta, NULL, &cert_len);
+                        if (resp.rc == WH_ERROR_OK) {
+                            if ((meta.flags & WH_NVM_FLAGS_NONEXPORTABLE) !=
+                                0) {
```
AlexLanzano, line 684: Nit: code formatting here is a bit odd.

src/wh_server_cert.c
```diff
@@ -612,19 +658,46 @@ int wh_Server_HandleCertRequest(whServerContext* server, uint16_t magic,
...
-                            resp.rc  = wh_Server_CertReadTrusted(
-                                server, req.id, cert_data, &cert_len);
+                    }
+                    else {
+                        /* NVM path */
+                        resp.rc =
+                            wh_Nvm_GetMetadata(server->nvm, req.id, &meta);
+                        if (resp.rc == WH_ERROR_OK) {
+                            if ((meta.flags & WH_NVM_FLAGS_NONEXPORTABLE) !=
+                                0) {
```
AlexLanzano, line 703: See above comment

billphipps (dismissed):
Minor nits.  Looks great!  Ty!

test/wh_test_cert.c
AlexLanzano, line 1: Test for ACERT?

AlexLanzano (changes requested):
Looks great! Some minor nits and possibly need a test for ACERT stuff

test/wh_test_cert.c
bigbrett: added

src/wh_server_cert.c
bigbrett: clang-format does as clang-format does. Sometimes you gotta just live with it.

src/wh_server_cert.c
bigbrett: clang-format does as clang-format does. Sometimes you gotta just live with it.

src/wh_server_cert.c
bigbrett: big change across message layer too but fixed in last commit

src/wh_client_cert.c
bigbrett: fixed - changed to the max cert size

docs/draft/wrapped-certs.md
bigbrett: done

src/wh_client_cert.c
bigbrett: no just following pattern from client keywrap API. Server keywrap code is gated on this but it will jsut return a runtime "not implemented error". Keeping for now since it doesnt actually use any crypto code on client side

src/wh_server_cert.c
```diff
@@ -522,12 +561,19 @@ int wh_Server_HandleCertRequest(whServerContext* server, uint16_t magic,
                 whKeyId keyId = wh_KeyId_TranslateFromClient(
                     WH_KEYTYPE_CRYPTO, server->comm->client_id, req.keyId);
 
+                /* Translate trustedRootId if client set wrapped flag */
+                whKeyId trustedRootId = req.trustedRootId;
+                if (req.trustedRootId & WH_KEYID_CLIENT_WRAPPED_FLAG) {
+                    trustedRootId = wh_KeyId_TranslateFromClient(
+                        WH_KEYTYPE_WRAPPED, server->comm->client_id,
+                        req.trustedRootId);
+                }
```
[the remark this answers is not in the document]
bigbrett: false positive, The lookup always happens in the current server context's own cache

src/wh_server_cert.c
```diff
@@ -522,12 +561,19 @@ int wh_Server_HandleCertRequest(whServerContext* server, uint16_t magic,
...
 
                 rc = WH_SERVER_NVM_LOCK(server);
                 if (rc == WH_ERROR_OK) {
                     /* Process the verify action */
                     rc = wh_Server_CertVerify(server, cert_data, req.cert_len,
-                                              req.trustedRootNvmId, req.flags,
+                                              trustedRootId, req.flags,
```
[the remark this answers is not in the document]
bigbrett: false positive, The lookup always happens in the current server context's own cache

docs/draft/wrapped-certs.md
```diff
@@ -0,0 +1,294 @@
...
+`WH_CLIENT_KEYID_MAKE_WRAPPED(cachedCertId)` sets bit 9
+(`WH_KEYID_CLIENT_WRAPPED_FLAG = 0x0200`) on the ID the client sends to the
+server. This is the signal that tells the server "this root cert is in the
+cache, not in NVM."
+
+The same pattern works for:
+- `wh_Client_CertVerifyDma` (DMA path)
+- `wh_Client_CertReadTrusted` / `wh_Client_CertReadTrustedDma` (read-back)
+- `wh_Client_CertVerifyAcert` / `wh_Client_CertVerifyAcertDma` (attribute certs)
+
```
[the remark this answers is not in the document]
bigbrett: about to be refactored anyway. They are the same underlying type.

src/wh_client_cert.c
```diff
@@ -958,6 +954,171 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
...
+                                         whKeyId            serverKeyId,
+                                         const uint8_t*     wrappedCert,
+                                         uint16_t           wrappedCertSz)
+{
+    if ((ctx == NULL) || (wrappedCert == NULL) || (wrappedCertSz == 0)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    return wh_Client_KeyUnwrapAndExportRequest(
+        ctx, cipherType, serverKeyId, (void*)wrappedCert, wrappedCertSz);
```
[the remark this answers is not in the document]
bigbrett: out of scope

src/wh_client_cert.c
```diff
@@ -958,6 +954,171 @@ int wh_Client_CertVerifyAcertDma(whClientContext* c, const void* cert,
...
+                                        whKeyId            serverKeyId,
+                                        const uint8_t*     wrappedCert,
+                                        uint16_t           wrappedCertSz)
+{
+    if ((ctx == NULL) || (wrappedCert == NULL) || (wrappedCertSz == 0)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    return wh_Client_KeyUnwrapAndCacheRequest(
+        ctx, cipherType, serverKeyId, (void*)wrappedCert, wrappedCertSz);
```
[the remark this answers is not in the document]
bigbrett: out of scope

src/wh_client_cert.c
```diff
@@ -807,23 +806,21 @@ int wh_Client_CertVerifyDmaAndCacheLeafPubKey(
...
     whMessageCert_VerifyAcertRequest req = {0};
     uint8_t                          buffer[WOLFHSM_CFG_COMM_DATA_LEN];
     size_t                           hdr_len = sizeof(req);
     uint8_t*                         payload = buffer + hdr_len;
 
 
-    if ((c == NULL) || (trustedRootNvmId == WH_NVM_ID_INVALID) ||
-        (cert == NULL) || (cert_len == 0) ||
-        (cert_len > (sizeof(buffer) - hdr_len))) {
+    if ((c == NULL) || (trustedRootId == WH_NVM_ID_INVALID) || (cert == NULL) ||
```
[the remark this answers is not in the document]
bigbrett: the sentinel check should still hold in this case

bigbrett:
This actually won't work due to it requiring a single ID field to work for both a keyId + flags and a raw NVM object ID. Until we refactor NVM layer to also do translation like the keystore, the wrapped flag is indistinguishable from a valid NVM ID. Reverting to draft until this can be resolved.

PR #326, opened 2026-03-23
Use ConstantCompare instead of memcmp in wh_server_she.c

padelsbach:
Fixes F-127

### DIFF
--- a/src/wh_server_she.c
+++ b/src/wh_server_she.c
@@ -43,6 +43,7 @@
 #include "wolfssl/wolfcrypt/wc_port.h"
 #include "wolfssl/wolfcrypt/aes.h"
 #include "wolfssl/wolfcrypt/cmac.h"
+#include "wolfssl/wolfcrypt/misc.h"
 #include "wolfhsm/wh_server_keystore.h"
 #endif /* !WOLFHSM_CFG_NO_CRYPTO */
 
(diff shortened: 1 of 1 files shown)


billphipps:
> @billphipps thoughts on side channel mitigations via pulling in wolfCrypt? Or do you think we should have our own wolfHSM internal constant compare?

In this case, server_she already requires wolfCrypt, so I don't see any reason not to leverage the ConstantCompare.  I don't think we'll need a wolfHSM specific version of this as anything I can think of that would need ConstantCompare would already depend on wolfCrypt.

Can we take a look at any other memcpy/memcmp calls that are in crypto sensitive functions?

padelsbach:
> Can we take a look at any other memcpy/memcmp calls that are in crypto sensitive functions?

The only remaining use of `memcmp` in the src directory is in `whFlashRamsim_Verify()`. Is that a concern?

Almost 400 instances of `memcpy`. Maybe I can narrow in on the const time crypto routines.

bigbrett (approved):

bigbrett:
@padelsbach no that memcmp is fine, its just in the ram simulator. 

I dont think `memcpy` is a concern (yet) when it comes to crypto. wolfCrypt just lets you override this, as well as all string functions, for purposes of portability. The compare is obci a cryptographically sensitive operation from a length-dependent sidechannel perspective.

memcpy could still have side-channel leakages due to things like memory access time and page-faults but wolfCrypt does not to my knowledge have any mitigations surrounding this as it is not typically a remote exploit. This is not the purpose of XMEMCPY - that is just for portability sake.

TL;DR open season on replacement of `memcmp` with `ConstantCompare()` in cryptographically sensitive buffer comparisons (e.g. where a side channel exists for extraction of sensitive data). The fact that `ConstantCompare()` is part of wolfCrypt will help limit the scope to only `!WOLFHSM_CFG_NO_CRYPTO` code as well. No need to touch memcpy for now.

PR #331, opened 2026-04-08
Initial prototype of test refactor

padelsbach:
Refactor top level test modules with the following:

- New test runner
- Unified connection logic (helper)
- Cleaner integration of sequenced tests

See more in the README.md

### DIFF
--- a/.github/workflows/code-coverage-refactor.yml
+++ b/.github/workflows/code-coverage-refactor.yml
@@ -1,4 +1,4 @@
-name: Code Coverage
+name: Code Coverage Refactor
 
 on:
   push:
@@ -44,15 +44,15 @@ jobs:
 
       - name: Build, run, and emit tracefile (${{ matrix.name }})
         run: |
-          cd test && make coverage-json \
-            OUT=$GITHUB_WORKSPACE/cov-json/legacy-${{ matrix.name }}.json \
-            WOLFSSL_DIR=../wolfssl ${{ matrix.flags }}
+          cd test-refactor/posix && make coverage-json \
+            OUT=$GITHUB_WORKSPACE/cov-json/refactor-${{ matrix.name }}.json \
+            WOLFSSL_DIR=../../wolfssl ${{ matrix.flags }}
 
       - name: Upload tracefile
         uses: actions/upload-artifact@v4
         with:
-          name: legacy-tracefile-${{ matrix.name }}
-          path: cov-json/legacy-${{ matrix.name }}.json
+          name: refactor-tracefile-${{ matrix.name }}
+          path: cov-json/refactor-${{ matrix.name }}.json
           retention-days: 7
 
   merge:
@@ -66,10 +66,10 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y gcovr
 
-      - name: Download legacy tracefiles
+      - name: Download refactor tracefiles
         uses: actions/download-artifact@v4
         with:
-          pattern: legacy-tracefile-*
+          pattern: refactor-tracefile-*
           path: cov-json
           merge-multiple: true
 
@@ -78,7 +78,7 @@ jobs:
           # nullglob so an empty match returns 0 files instead of the
           # literal pattern, which would slip past the guard below.
           shopt -s nullglob
-          files=(cov-json/legacy-*.json)
+          files=(cov-json/refactor-*.json)
           if [ ${#files[@]} -eq 0 ]; then
             echo "No tracefiles found to merge" >&2
             exit 1
@@ -88,14 +88,14 @@ jobs:
           for f in "${files[@]}"; do
             args+=(--add-tracefile "$f")
           done
-          mkdir -p coverage-merged
+          mkdir -p coverage-refactor-merged
           gcovr "${args[@]}" \
-            --html-details coverage-merged/index.html \
+            --html-details coverage-refactor-merged/index.html \
             --print-summary
 
       - name: Upload merged coverage report
         uses: actions/upload-artifact@v4
         with:
-          name: coverage-report
-          path: coverage-merged/
+          name: coverage-refactor-report
+          path: coverage-refactor-merged/
           retention-days: 30
--- a/test-refactor/README.md
+++ b/test-refactor/README.md
@@ -1,0 +1,104 @@
+# wolfHSM unit tests
+
+## Overview
+The wolfHSM unit tests are designed to run on separate cores, or separate processes to model a real HSM client-server.
+
+Tests can be run on either a POSIX host (ie a development laptop), or on the embedded target.
+
+The unit tests are split into 3 groups:
+
+- Client, for tests that run on the client core/process
+- Server, for tests that run on the server core/process
+- Misc, for tests that can run on either core/process
+
+The groups are organized in `wh_test_list.c`.
+
+## Running the tests from a dev machine
+To run the unit tests on a POSIX system (Linux/Mac):
+
+```
+cd test-refactor
+make check
+```
+
+The top-level `make` forwards to the POSIX port; `cd test-refactor/posix && make check` still works if you want to invoke the port directly.
+
+Results are printed via `WOLFHSM_CFG_PRINTF` from the wolfHSM build. `test-suite.log` contains the detailed output.
+
+## Running the tests from an embedded target
+To run the tests on a target device, create an application running on the client or server that runs the tests from `main()`. See sections on adding ports and tests.
+
+## Adding a port
+The unit tests run within the port's main application. As a prerequiste, setup a new port application as described in the [porting guide](docs/src/chapter08.md).
+
+For the unit test port, see `wh_test_posix_main.c` and the two `wh_test_posix_*.c` sources as a reference implementation.
+
+### Client
+1. Implement `main()` which creates a client context and initializes it with a config such that it can establish communication to a listening server
+2. Implement `whTestPort_ResetClient` which resets the context between tests. Can be empty.
+3. Call `whTestGroup_Client(&clientCtx)`
+4. Optionally call `whTestGroup_Misc()`
+
+For running the more substantial client tests only, no server modifications are required.
+
+### Server
+1. Implement main() which declares a server context and initializes it with a config such that it can access basic platform functionality (NVM, etc.)
+2. Implement `whTestPort_ResetServer` which resets the context between tests. Can be empty.
+3. Call `whTestGroup_Server(&serverCtx)` prior to entering the main request handling loop.
+4. Optionally call `whTestGroup_Misc()`
+
+## Adding a multi-port test
+1. Create a new function which returns `int` (0 for success) with a context argument (`whClientContext*` for client tests, `whServerContext*` for server tests, or none for misc tests).
+2. In wh_test_list.c, add a line with `WH_TEST_DECL(<function>)`
+3. In wh_test_list.c, add the function to the appropriate `whTestCase` array. 
+
+**Note**: if the test is specific to a platform, do not add it to the common list as shown above. Port-specific tests live within the port (not this directory), and are called from the port-specific code.
+
+## Adding a port-specific test
+For tests that exist within a specific port, call `whTestGroup_RunOne()` to utilize the error checking and log formatting from the test framework.
+
+## Migration from `test/`
+The legacy unit suite in `wolfHSM/test/` is being incrementally translated to `wolfHSM/test-refactor/` while preserving the original during the transition. Code coverage Github workflows will be used for confirmation.
+
+### Key differences
+- Tests are registered and called based on a list in `wh_test_list.c` rather than manual inline code, making it simpler to add tests.
+- Tests accept a client or server context which is initialized outside the test itself, reducing copypasta within the test code, and improving portability.
+- Tests are divided into groups, which clarifies the origin and environment of the test.
+- Tests are always run against a running server process or core, no sequencing code for single-thread simulation.
+
+### Test mapping
+Translated tests:
+
+| Legacy (`wolfHSM/test/`) | New location | Group | Notes |
+|---|---|---|---|
+| `wh_test_dma.c::whTest_Dma` | `misc/wh_test_dma.c::whTest_Dma` | Misc | |
+| `wh_test_cert.c::whTest_CertRamSim` | `server/wh_test_cert.c::whTest_CertVerify` | Server | remove ramsim coupling and migrate to server group |
+| `wh_test_crypto.c::whTest_Crypto` | `client-server/wh_test_crypto.c::{whTest_CryptoSha256, whTest_CryptoAes, whTest_CryptoEcc256}` | Client | Subset only; remaining cases listed below |
+| `wh_test_clientserver.c` (echo and server-info paths) | `client-server/wh_test_echo.c::whTest_Echo`, `client-server/wh_test_server_info.c::whTest_ServerInfo` | Client | pthread test ported, sequential test dropped |
+| `wh_test_wolfcrypt_test.c::whTest_WolfCryptTest` | `client-server/wh_test_wolfcrypt.c::whTest_WolfCryptTest` | Client | |
+| `wh_test_flash_ramsim.c::whTest_Flash_RamSim` | `posix/wh_test_flash_ramsim.c::{whTest_FlashWriteLock, whTest_FlashEraseProgramVerify, whTest_FlashUnitOps}` | POSIX port-specific (`whTestGroup_RunOne`) | remove ramsim coupling and migrate to server group |
+| `wh_test_nvm_flash.c::whTest_NvmFlash` | `posix/wh_test_nvm_flash.c::whTest_NvmAddOverwriteDestroy` | POSIX port-specific (`whTestGroup_RunOne`) | remove ramsim coupling and migrate to server group |
+| `wh_test_posix_threadsafe_stress.c::whTest_ThreadSafeStress` | called directly from `posix/wh_test_posix_main.c` | POSIX port-specific (direct call) | |
+
+Not yet migrated (still live in `wolfHSM/test/`):
+
+| Legacy (`wolfHSM/test/`) | Notes |
+|---|---|
+| `wh_test_comm.c::whTest_Comm` | |
+| `wh_test_clientserver.c::whTest_ClientServer` | Pthread variant: remaining client-side coverage (NVM ops, etc.) still needs to be split out as new tests. The sequential test is dropped |
+| `wh_test_crypto.c::whTest_Crypto` | RNG, key cache, key-cache enforcement, RSA, CMAC, Curve25519, ML-DSA, key usage policies, key revocation |
+| `wh_test_crypto_affinity.c::whTest_CryptoAffinity` | |
+| `wh_test_keywrap.c::whTest_KeyWrapClientConfig` | |
+| `wh_test_multiclient.c::whTest_MultiClient` | |
+| `wh_test_lock.c::whTest_LockConfig`, `whTest_LockPosix` | `whTest_LockConfig` to be reworked to fit the Misc group, likely with a context param. |
+| `wh_test_log.c::whTest_Log`, `whTest_LogBackend_RunAll` | `whTest_LogBackend_RunAll` to be reworked to fit the Misc group, likely with a context param. |
+| `wh_test_she.c::whTest_She` | |
+| `wh_test_timeout.c::whTest_TimeoutPosix` | |
+| `wh_test_auth.c::whTest_AuthMEM`, `whTest_AuthTCP` | |
+| `wh_test_server_img_mgr.c::whTest_ServerImgMgr` | |
+| `wh_test_nvmflags.c::whTest_NvmFlags` | |
+| `wh_test_flash_fault_inject.c` | |
+| `wh_test_check_struct_padding.c` | |
+
+### Other improvements
+- Add callback from `wh_Server_HandleRequestMessage` to allow sleep and avoid a busy loop
--- a/test-refactor/client-server/wh_test_crypto.c
+++ b/test-refactor/client-server/wh_test_crypto.c
@@ -1,0 +1,220 @@
+/*
+ * Copyright (C) 2026 wolfSSL Inc.
+ *
+ * This file is part of wolfHSM.
+ *
+ * wolfHSM is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfHSM is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with wolfHSM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/*
+ * test-refactor/wh_test_crypto.c
+ *
+ * Basic crypto test suite. Minimal SHA256 and AES-CBC
+ * round-trips routed through the server via WH_DEV_ID.
+ */
+
+#include "wolfhsm/wh_settings.h"
+
+#if !defined(WOLFHSM_CFG_NO_CRYPTO)
+
+#include <stdint.h>
+#include <string.h>
+
+#include "wolfssl/wolfcrypt/settings.h"
+#include "wolfssl/wolfcrypt/types.h"
+#include "wolfssl/wolfcrypt/aes.h"
+#include "wolfssl/wolfcrypt/ecc.h"
+#include "wolfssl/wolfcrypt/random.h"
+#include "wolfssl/wolfcrypt/sha256.h"
+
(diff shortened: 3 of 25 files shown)


test-refactor/wh_test_port.h
```diff
@@ -0,0 +1,40 @@
...
+#include "wolfhsm/wh_server.h"
+
+/*
+ * Caller-implemented reset hooks. The group functions invoke
+ * these between suites so the caller can reset any persistent
+ * state (NVM objects, key cache, connection state, ...) between
+ * tests. Can be empty, simply return 0.
+ */
+int whTestGroup_ResetServer(whServerContext* server);
+int whTestGroup_ResetClient(whClientContext* client);
```
bigbrett: should this be `whTestPort_ResetXXX()` following naming convention based on file/module?

test-refactor/wh_test_list.h
```diff
@@ -0,0 +1,104 @@
...
+ * skip implementation. wh_test_list.c contains a `WH_TEST_DECL(name);`
+ * line for every test in the registry. If the real test's feature
+ * gate is on, its strong definition in the test's .c file overrides
+ * this stub at link time; otherwise the stub wins and the test
+ * surfaces as SKIPPED.
+ *
+ * The trailing "struct ... whTest_decl_dummy_" lets callers end
+ * the invocation with a ';' (a bare "};" after a function body
+ * is ill-formed under strict C90); the dummy tag swallows it.
+ */
```
bigbrett, line 73: Read this 3 times and it still wasn't clear what it is for or what it does. How about this:
```suggestion
/*
 * Declares a test prototype and provides a weak default stub.
 * wh_test_list.c uses WH_TEST_DECL(name) for every registered test.
 *
 * If the real test is compiled in, its strong definition overrides
 * this stub at link time. Otherwise, the stub remains and the test
 * is reported as SKIPPED.
 *
 * The trailing dummy struct lets callers safely end the macro with
 * a semicolon, which avoids issues under strict C90.
 */
```

test-refactor/wh_test_list.h
```diff
@@ -0,0 +1,104 @@
...
+typedef struct whTestCase {
+    const char* name;
+    whTestFn    fn;
+} whTestCase;
+
+/*
+ * Per-group registries, defined in wh_test_list.c. Each group gets
+ * its own array so the runner walks only the relevant tests and
+ * doesn't carry a group tag per row.
+ */
```
bigbrett, line 92:
```suggestion
/*
 * Per-group registries, defined in wh_test_list.c. Each group gets
 * its own array so the runner walks only the relevant tests.
 */
```

test-refactor/wh_test_list.h
```diff
@@ -0,0 +1,104 @@
...
+ * ignore it. Spellings covered:
+ *   GCC, Clang, armclang, armcc, TI  -> __attribute__((weak))
+ *   IAR                               -> __weak
+ *   Renesas CC-RH / CC-RL / CC-RX    -> _Pragma("weak <name>")
+ *
+ * If a port's toolchain isn't covered here, add it rather than
+ * silently falling back -- a no-op WH_TEST_WEAK makes the stub
+ * strong and will cause a multiple-definition link error (or,
+ * worse, the stub wins over the real test).
+ */
```
bigbrett, line 43:
```suggestion
/*
 * Portable weak-symbol attribute.
 *
 * wh_test_list.c defines a weak stub for every test. If the real test
 * is compiled in, its strong definition overrides the stub at link time.
 * If not, the stub remains and returns WH_TEST_SKIPPED.
 *
 * Usage:
 *   WH_TEST_WEAK(foo) int foo(void* ctx) { ... }
 *
 * The name is only needed for pragma-based toolchains; others ignore it.
 *
 * Supported toolchains:
 *   GCC, Clang, ARM, TI   -> __attribute__((weak))
 *   IAR                   -> __weak
 *   Renesas (CC-RH/RL/RX) -> _Pragma("weak <name>")
 *
 * Add new toolchains explicitly. A missing definition here will either
 * cause link errors or let the stub override the real test.
 */
```

test-refactor/wh_test_groups.c
```diff
@@ -0,0 +1,133 @@
...
+    else {
+        WOLFHSM_CFG_PRINTF("test FAILED (rc=%d)\n", rc);
+        whTestFailed++;
+    }
+}
+
+int whTestGroup_RunOne(const char* name, int (*fn)(void*),
+    void* ctx)
+{
+    int rc = fn(ctx);
```
bigbrett: Null check on `fn`?

test-refactor/wh_test_groups.h
```diff
@@ -0,0 +1,69 @@
...
+int whTestGroup_Client(whClientContext* client);
+
+/*
+ * Run a single test outside the portable registry, print its
+ * result in the standard format, and feed the summary tally.
+ * Used by ports to invoke platform-specific tests (e.g. POSIX
+ * host-sim suites) so they appear in the final tally alongside
+ * the portable groups. Returns the test's rc unchanged.
+ */
+int whTestGroup_RunOne(const char* name, int (*fn)(void*),
```
bigbrett: I think this is should probably be static?

test-refactor/wh_test_groups.c
```diff
@@ -0,0 +1,133 @@
...
+    if (whTestFailed == 0 && whTestSkipped == 0) {
+        WOLFHSM_CFG_PRINTF("All %d tests passed!\n", total);
+    }
+    else {
+        WOLFHSM_CFG_PRINTF(
+            "%d passed, %d skipped, %d failed of %d tests\n",
+            whTestPassed, whTestSkipped, whTestFailed, total);
+    }
+
+    return whTestFailed == 0 ? 0 : -1;
```
bigbrett:
```suggestion
    return whTestFailed == 0 ? WH_ERROR_OK : WH_ERROR_ABORTED;
```

test-refactor/wh_test_groups.c
```diff
@@ -0,0 +1,133 @@
...
+
+static int whTest_Run(const whTestCase* tests, size_t count, void* ctx)
+{
+    int    overall = 0;
+    size_t i;
+
+    for (i = 0; i < count; i++) {
+        int rc = whTestGroup_RunOne(tests[i].name, tests[i].fn, ctx);
+
+        if (rc != 0 && rc != WH_TEST_SKIPPED && overall == 0) {
```
bigbrett: prefer `rc != WH_ERROR_OK` unless specifically meant to be a wolfCrypt error code, but I think these should all be intercepted by the test return values at this point

test-refactor/wh_test_groups.c
```diff
@@ -0,0 +1,133 @@
...
+            overall = rc;
+        }
```
bigbrett: currently our tests fail-fast. I think it makes sense to keep this paradigm in place for now (e.g. fail out the moment a test fails so you dont need to dig through a log). Can be changed in the future but for now it has proven valuable for quick debugging as there will likely be cascaded failures that are difficult to diagnose otherwise.

test-refactor/README.md
```diff
@@ -0,0 +1,58 @@
...
+- Misc, for tests that can run on either core/process
+
+The groups are organized in `wh_test_list.c`.
+
+## Running the tests from a dev machine
+To run the unit tests on a POSIX system (Linux/Mac):
+
+```
+cd test-refactor/posix
+make check
```
bigbrett: can we have a top level test-refactor/ make target to preserve the ergonomics of the current test runs? I'd rather just "change to test directory and run" vs explicitly navigating to the posix test app directory

test-refactor/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Running the tests from an embedded target
+To run the tests on a target device, create an application running on the client or server that runs the tests from `main()`. See sections on adding ports and tests.
+
+## Adding a port
+The unit tests run within the port's main application. As a prerequiste, setup a new port application as described in the [porting guide](docs/src/chapter08.md).
+
+For the unit test port, see `wh_test_posix_main.c` and the two `wh_test_posix_*.c` sources as a reference implementation.
+
+### Client
+1. Implement `main()` which creates a client context
```
bigbrett:
```suggestion
1. Implement `main()` which creates a client context and initializes it with a config such that it can establish communication to a listening server
```

test-refactor/README.md
```diff
@@ -0,0 +1,58 @@
...
+1. Implement `whTestGroup_ResetClient` which resets the context between tests. Can be empty.
+2. Optionally call `whTestGroup_Misc()`
+3. Call `whTestGroup_Client(&clientCtx)`
```
bigbrett, line 39: have this step before Misc just for clarity, since this is "client" section

bigbrett: fix numbering (double 1)

test-refactor/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Server
+The client unit tests can be run against the normal server application -- no special modifications required.
```
bigbrett: this line should be in the client section

test-refactor/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+To add the server side unit tests:
+1. Implement `whTestGroup_ResetServer` which resets the context between tests. Can be empty.
+2. Call `whTestGroup_Server(&serverCtx)` prior to entering the main request handling loop.
```
bigbrett: start from main and talk about context init like in client section

1. Implement `main()` which declares a server context and initializes it with a config such that it can access basic platform functionality (NVM, etc.)

test-refactor/posix/wh_test_posix_main.c
```diff
@@ -0,0 +1,316 @@
...
+ * RAM-based flash simulator, which is a host-sim component;
+ * the nvm_flash test wires the NVM stack to that simulator
+ * with a 1 MB buffer that's not realistic on embedded targets.
+ * Both run from the POSIX port directly until the NVM test is
+ * reworked to take a port-supplied flash fixture (then it can
+ * lift back into whTestGroup_Server). */
+int whTest_FlashWriteLock(void* ctx);
+int whTest_FlashEraseProgramVerify(void* ctx);
+int whTest_FlashUnitOps(void* ctx);
+int whTest_NvmAddOverwriteDestroy(void* ctx);
```
bigbrett, line 72: Hmmm so I guess you are treating this now as "port-specific" tests? I guess that is fine based on what they test. Perhaps should consider reorganizing or renaming at minimum - ok for now

test-refactor/wh_test_port.h
bigbrett: > The group functions invoke these between suites so the caller can reset any persistent state (NVM objects, key cache, connection state, ...) between tests.

Is this true? I don't see them called anywhere. IMO this should be managed internal to the runner across tests if required? Otherwise, what was the intent here? In general, tests should "clean up after themselves" if creating mutable state (NVM object, keycache entries, allocated memory, etc) but I would assume these would be called internally before running each test function as a "setup" or "teardown" operation?

.github/workflows/build-and-test.yml
```diff
@@ -36,62 +36,110 @@ jobs:
     - name: Build and test
       run: cd test && make clean && make -j WOLFSSL_DIR=../wolfssl && make run
 
+    - name: Build and test refactor
```
bigbrett: perhaps these should all just go into a new workflow to keep separate? `build-and-test-refactor.yml`? Also want coverage tracked between the two so should have a new coverage action for that

bigbrett (changes requested):
Overall,as discussed, looks great. A few things I wanted to point out and some architectural questions that should be addressed.

Also, can we get a code coverage comparison set up in CI so we can track the coverage metrics between existing test and new harness?

bigbrett:
@padelsbach oh and also, a section of the new README.md mapping old test constructs -> new test constructs would be valuable for people who need to understand the migration. Both discussing key differences at a high level, as well as mapping each individual test from its old location to its new location (this last part can be an AI table - prefer some human input for the first part)

test-refactor/wh_test_port.h
padelsbach: done

test-refactor/wh_test_list.h
padelsbach: done

test-refactor/wh_test_list.h
padelsbach: done

test-refactor/wh_test_list.h
padelsbach: done

test-refactor/wh_test_groups.c
padelsbach: done

test-refactor/wh_test_groups.h
padelsbach: Port-specific tests should call this to get the error checking (stop-on-failure behavior), print formatting. I added a note to the readme

test-refactor/wh_test_groups.c
padelsbach: done

test-refactor/wh_test_groups.c
padelsbach: done

test-refactor/wh_test_groups.c
padelsbach: This was an intentional change, but I see your point. Fixed

test-refactor/README.md
padelsbach: Added, but note this lives outside/above the posix port-specific code.

test-refactor/README.md
padelsbach: done

test-refactor/README.md
padelsbach: done

padelsbach: done

test-refactor/README.md
padelsbach: done

test-refactor/README.md
padelsbach: done

test-refactor/posix/wh_test_posix_main.c
padelsbach: These currently hardcode the ramsim backend, and decoupling is non-trivial. Planning to rename at that time. This is noted in the readme:

```
remove ramsim coupling and migrate to server group
```

test-refactor/wh_test_port.h
padelsbach: Good catch, looks like this got dropped when simplifying the macros.

You're right. Idea is to give the port an opportunity to setup/teardown the context between tests without knowledge about the internals of the context. 

I added the missing calls back. I don't feel too strongly about this -- I can remove if you do.

.github/workflows/build-and-test.yml
padelsbach: I put them here to promote equivalence between the refactor, reduce the possibility of them getting out of sync. Added `build-and-test-refactor.yml` and `code-coverage-refactor.yml` which mirror the originals.

bigbrett (dismissed):
Great start on the framework. This is at like 90% for me but dont want to delay getting it in, since it wont break anything. Opening up to others for investigation and thoughts.

test-refactor/server/wh_test_cert.c
```diff
@@ -0,0 +1,118 @@
...
+
+    /* Add trusted roots */
+    WH_TEST_RETURN_ON_FAIL(wh_Server_CertAddTrusted(
+        server, rootA, WH_NVM_ACCESS_ANY,
+        WH_NVM_FLAGS_NONMODIFIABLE,
+        NULL, 0, ROOT_A_CERT, ROOT_A_CERT_len));
+
+    WH_TEST_RETURN_ON_FAIL(wh_Server_CertAddTrusted(
+        server, rootB, WH_NVM_ACCESS_ANY,
+        WH_NVM_FLAGS_NONMODIFIABLE,
```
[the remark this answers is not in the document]
padelsbach: `wh_Server_CertEraseTrusted` calls `wh_Nvm_DestroyObjects` (not `wh_Nvm_DestroyObjectsChecked`), so the test does not fail. This test mirrors the behavior from the original `test/wh_test_cert.c`, so treating this as out of scope.

test-refactor/client-server/wh_test_crypto.c
```diff
@@ -0,0 +1,220 @@
...
+#include "wh_test_common.h"
+#include "wh_test_list.h"
+
+#ifndef NO_SHA256
+int whTest_CryptoSha256(whClientContext* ctx)
+{
+    int              devId = WH_DEV_ID;
+    int              ret   = WH_ERROR_OK;
+    wc_Sha256        sha256[1];
+    uint8_t          out[WC_SHA256_DIGEST_SIZE];
```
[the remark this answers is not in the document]
padelsbach: This is intentional and harmless with known compilers, though some pedantic tools and reviewers (ahem) may flag it. The alternatives would be:

1. Redefine test functions to take a `(void*)` and cast internally. This relies upon an assumption between test list and test itself, which is no better imho.
2. Split the test macros in `wh_test_list.h` for client, server and misc. This is more explicit, but very wordy in the common test code.

I'll add a comment around `whTestFn` about this. We can address if/when there is ever an issue.

test-refactor/client-server/wh_test_crypto.c
```diff
@@ -0,0 +1,220 @@
...
+
+
+#if defined(HAVE_ECC) && defined(HAVE_ECC_SIGN) && defined(HAVE_ECC_VERIFY)
+int whTest_CryptoEcc256(whClientContext* ctx)
+{
+    int              devId  = WH_DEV_ID;
+    int              ret    = 0;
+    WC_RNG           rng[1];
+    ecc_key          key[1];
+    uint8_t          hash[32]              = {0};
```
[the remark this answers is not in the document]
padelsbach: See comment above

test-refactor/client-server/wh_test_echo.c
```diff
@@ -0,0 +1,68 @@
...
+#define REPEAT_COUNT 10
+
+/*
+ * Echo a message to the server and verify the response
+ * matches. Repeats several times with different payloads.
+ */
+int whTest_Echo(whClientContext* ctx)
+{
+    char     send_buf[WOLFHSM_CFG_COMM_DATA_LEN];
+    char     recv_buf[WOLFHSM_CFG_COMM_DATA_LEN];
```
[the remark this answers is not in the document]
padelsbach: See comment above

test-refactor/client-server/wh_test_crypto.c
```diff
@@ -0,0 +1,220 @@
...
+#endif /* !NO_SHA256 */
+
+
+#if !defined(NO_AES) && defined(HAVE_AES_CBC)
+int whTest_CryptoAes(whClientContext* ctx)
+{
+    int              devId = WH_DEV_ID;
+    int              ret   = 0;
+    Aes              aes[1];
+    uint8_t          cipher[AES_BLOCK_SIZE]   = {0};
```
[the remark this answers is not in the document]
padelsbach: see comment above

test-refactor/client-server/wh_test_server_info.c
```diff
@@ -0,0 +1,78 @@
...
+
+
+/*
+ * Query server info and verify the response contains
+ * valid data.
+ */
+int whTest_ServerInfo(whClientContext* ctx)
+{
+    uint8_t  version[WH_INFO_VERSION_LEN + 1] = {0};
+    uint8_t  build[WH_INFO_VERSION_LEN + 1]   = {0};
```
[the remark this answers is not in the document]
padelsbach: see comment above

test-refactor/server/wh_test_cert.c
```diff
@@ -0,0 +1,118 @@
...
+#include "wh_test_common.h"
+#include "wh_test_list.h"
+#include "wh_test_cert_data.h"
+
+
+/*
+ * Add trusted roots, verify valid and invalid certs/chains,
+ * then remove roots.
+ */
+int whTest_CertVerify(whServerContext* ctx)
```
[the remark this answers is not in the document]
padelsbach: see comment above

test-refactor/client-server/wh_test_wolfcrypt.c
```diff
@@ -0,0 +1,48 @@
...
+
+#include "wolfhsm/wh_error.h"
+#include "wolfhsm/wh_client.h"
+
+#include "wh_test_common.h"
+#include "wh_test_list.h"
+
+int whTest_WolfCryptTest(whClientContext* ctx)
+{
+    (void)ctx;
```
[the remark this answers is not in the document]
padelsbach: see comment above

test-refactor/wh_test_groups.c
```diff
@@ -0,0 +1,165 @@
...
+}
+
+int whTestGroup_RunOne(const char* name, int (*fn)(void*), void* ctx)
+{
+    int rc;
+
+    if (name == NULL || fn == NULL) {
+        return WH_TEST_SKIPPED;
+    }
+
```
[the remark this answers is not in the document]
padelsbach: done

test-refactor/posix/wh_test_posix_server.c
```diff
@@ -0,0 +1,166 @@
...
+#include "wh_test_common.h"
+#include "wh_test_posix_server.h"
+
+#define POSIX_FLASH_SIZE       (1024 * 1024)
+#define POSIX_FLASH_SECTOR_SZ  (128 * 1024)
+#define POSIX_FLASH_PAGE_SZ    (8)
+/* Sized to hold a full WOLFHSM_CFG_COMM_DATA_LEN payload --
+ * the wolfCrypt test (TESTWOLFCRYPT=1) needs the full window
+ * for ML-DSA/Dilithium keys and signatures. */
+#define POSIX_TRANSPORT_BUF_SZ WH_COMM_MTU
```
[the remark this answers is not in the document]
padelsbach: done

test-refactor/posix/wh_test_posix_server.c
```diff
@@ -0,0 +1,166 @@
...
+
+    memset(&sCfg, 0, sizeof(sCfg));
+    sCfg.comm_config = &_commCfg;
+    sCfg.nvm         = &_nvm;
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+    sCfg.crypto      = &_crypto;
+#endif
+
+    return wh_Server_Init(server, &sCfg);
+}
```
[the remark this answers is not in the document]
padelsbach: done

test-refactor/posix/wh_test_posix_main.c
```diff
@@ -0,0 +1,316 @@
...
+    while (1) {
+        rc = wh_Server_HandleRequestMessage(&_server);
+        if (rc != WH_ERROR_OK && rc != WH_ERROR_NOTREADY) {
+            _serverRc = rc;
+            break;
+        }
+        (void)wh_Server_GetConnected(&_server, &connected);
+        if (connected == WH_COMM_DISCONNECTED) {
+            break;
+        }
```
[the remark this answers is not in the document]
padelsbach: This is a future improvement. Noted in the readme

test-refactor/posix/wh_test_nvm_flash.c
```diff
@@ -0,0 +1,302 @@
...
+    const whFlashCb    initFlashCb[1] = {WH_FLASH_RAMSIM_CB};
+    whNvmCb            initNvmCb[1]   = {WH_NVM_FLASH_CB};
+
+    memset(c, 0, sizeof(*c));
+
+    c->flashCb = initFlashCb[0];
+    c->flashCfg.size       = NVM_FLASH_SIZE;
+    c->flashCfg.sectorSize = NVM_FLASH_SECTOR_SZ;
+    c->flashCfg.pageSize   = NVM_FLASH_PAGE_SZ;
+    c->flashCfg.erasedByte = 0;
```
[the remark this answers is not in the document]
padelsbach: Mirroring original test code here

.github/workflows/code-coverage.yml
```diff
@@ -12,46 +12,78 @@ permissions:
...
+        with:
+          pattern: legacy-tracefile-*
+          path: cov-json
+          merge-multiple: true
+
+      - name: Merge tracefiles into HTML report
+        run: |
+          mkdir -p coverage-merged
+          gcovr \
+            $(for f in cov-json/legacy-*.json; do echo --add-tracefile $f; done) \
```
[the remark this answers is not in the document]
padelsbach: added defensive logic

.github/workflows/code-coverage-refactor.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          pattern: refactor-tracefile-*
+          path: cov-json
+          merge-multiple: true
+
+      - name: Merge tracefiles into HTML report
+        run: |
+          mkdir -p coverage-refactor-merged
+          gcovr \
+            $(for f in cov-json/refactor-*.json; do echo --add-tracefile $f; done) \
```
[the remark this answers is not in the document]
padelsbach: added defensive logic

bigbrett (approved):
Merging since I don't want this to linger, and it lives along-side the real tests until we move everything over.

@billphipps @AlexLanzano @rizlik I recommend you review this retroactively and keep abreast of the subsequent additions as they come in so that you dont have to do it all at the last minute when the migration decision happens.

PR #333, opened 2026-04-10
Fix sha512 buffer copy

padelsbach:
Fixes F-2006 and F-2007. See related but independent [wolfssl PR](https://github.com/wolfSSL/wolfssl/pull/10193).

wh_Client_Sha512 unconditionally copied WC_SHA512_DIGEST_SIZE (64) bytes on finalize, regardless of SHA512 variant. For SHA512/224 (28-byte digest) and SHA512/256 (32-byte digest), this overwrites caller memory past the output buffer.

Today this is masked because the cryptocb dispatcher only handled WC_HASH_TYPE_SHA512, so variants fell through to CRYPTOCB_UNAVAILABLE and wolfSSL's software fallback handled the truncation safely. But the variants were never HSM-accelerated, and any future dispatcher fix would expose the overflow.

### Changes
* Dispatcher (wh_client_cryptocb.c): Add WC_HASH_TYPE_SHA512_224 and WC_HASH_TYPE_SHA512_256 cases, routing to the same wh_Client_Sha512/wh_Client_Sha512Dma functions with info->hash.type passed through
* wh_Client_Sha512 (wh_client_crypto.c): New hashType parameter selects the correct digest size for memcpy and the correct wc_InitSha512*_ex variant for re-init
* wh_Client_Sha512Dma: Same treatment -- hashType drives digestSz, req->output.sz, and DMA post-processing size
* Tests (wh_test_crypto.c): New test cases for SHA512/224 and SHA512/256 with canary bytes to detect overwrite

### Design note
The hashType parameter comes from info->hash.type, which is set by wolfSSL's cryptocb dispatch layer based on digestSz -- not from sha512->hashType, which depends on the wolfSSL port's init code setting it. This makes the fix independent of wolfSSL port behavior.

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -5162,7 +5162,8 @@ int wh_Client_Sha384Dma(whClientContext* ctx, wc_Sha384* sha, const uint8_t* in,
 
 static int _xferSha512BlockAndUpdateDigest(whClientContext* ctx,
                                            wc_Sha512*       sha512,
-                                           uint32_t         isLastBlock)
+                                           uint32_t         isLastBlock,
+                                           int              hashType)
 {
     uint16_t                       group   = WH_MESSAGE_GROUP_CRYPTO;
     uint16_t                       action  = WH_MESSAGE_ACTION_NONE;
@@ -5199,10 +5200,13 @@ static int _xferSha512BlockAndUpdateDigest(whClientContext* ctx,
 
     /* Send the hash state - this will be 0 on the first block on a properly
      * initialized sha512 struct */
+    /* Source hashType from the dispatcher (info->hash.type) rather
+     * than sha512->hashType, since some wolfSSL ports do not
+     * populate the struct field. */
     memcpy(req->resumeState.hash, sha512->digest, WC_SHA512_DIGEST_SIZE);
     req->resumeState.hiLen    = sha512->hiLen;
     req->resumeState.loLen    = sha512->loLen;
-    req->resumeState.hashType = sha512->hashType;
+    req->resumeState.hashType = hashType;
     uint32_t req_len =
         sizeof(whMessageCrypto_GenericRequestHeader) + sizeof(*req);
 
@@ -5232,7 +5236,7 @@ static int _xferSha512BlockAndUpdateDigest(whClientContext* ctx,
         /* wolfCrypt allows positive error codes on success in some scenarios */
         if (ret >= 0) {
             WH_DEBUG_CLIENT_VERBOSE("Client SHA512 Res recv: ret=%d", ret);
-            WH_DEBUG_CLIENT_VERBOSE("hashType: %d\n", sha512->hashType);
+            WH_DEBUG_CLIENT_VERBOSE("hashType: %d\n", hashType);
             /* Store the received intermediate hash in the sha512
              * context and indicate the field is now valid and
              * should be passed back and forth to the server */
@@ -5248,12 +5252,18 @@ static int _xferSha512BlockAndUpdateDigest(whClientContext* ctx,
     return ret;
 }
 
-int wh_Client_Sha512(whClientContext* ctx, wc_Sha512* sha512, const uint8_t* in,
-                     uint32_t inLen, uint8_t* out)
+/* Shared implementation for SHA-512 and its truncated variants
+ * (SHA-512/224, SHA-512/256). The hashType parameter selects the
+ * output digest size and the variant-appropriate re-init on
+ * finalize. Sourcing hashType from the caller (the dispatcher)
+ * rather than sha512->hashType avoids a dependency on the wolfSSL
+ * port populating that struct field. */
+static int _doSha512(whClientContext* ctx, wc_Sha512* sha512,
+                     const uint8_t* in, uint32_t inLen, uint8_t* out,
+                     int hashType)
 {
     int      ret               = 0;
     uint8_t* sha512BufferBytes = (uint8_t*)sha512->buffer;
-    int      hashType          = WC_HASH_TYPE_SHA512;
 
     /* Caller invoked SHA Update:
      * wc_CryptoCb_Sha512Hash(sha512, data, len, NULL) */
@@ -5268,7 +5278,8 @@ int wh_Client_Sha512(whClientContext* ctx, wc_Sha512* sha512, const uint8_t* in,
                 sha512BufferBytes[sha512->buffLen++] = in[i++];
             }
             if (sha512->buffLen == WC_SHA512_BLOCK_SIZE) {
-                ret = _xferSha512BlockAndUpdateDigest(ctx, sha512, 0);
+                ret = _xferSha512BlockAndUpdateDigest(ctx, sha512, 0,
+                                                      hashType);
                 sha512->buffLen = 0;
             }
         }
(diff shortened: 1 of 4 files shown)


wolfhsm/wh_client_crypto.h
```diff
@@ -998,33 +998,37 @@ int wh_Client_Sha384Dma(whClientContext* ctx, wc_Sha384* sha, const uint8_t* in,
...
  * @param[in] ctx Pointer to the client context structure.
  * @param[in] sha Pointer to the SHA-512 context structure.
  * @param[in] in Pointer to the input data.
  * @param[in] inLen Length of the input data in bytes.
  * @param[out] out Pointer to the output buffer.
+ * @param[in] hashType The hash variant type (e.g. WC_HASH_TYPE_SHA512,
+ *            WC_HASH_TYPE_SHA512_224, or WC_HASH_TYPE_SHA512_256).
  * @return int Returns 0 on success or a negative error code on failure.
  */
 int wh_Client_Sha512(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
```
rizlik, line 1011: This is an API break. I would prefer to add new specific function wh_Client_Sha512_224/wh_Client_Sha512_384

src/wh_client_crypto.c
```diff
@@ -5341,12 +5370,12 @@ int wh_Client_Sha512Dma(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
 
     /* Setup generic header and get pointer to request data */
     req = (whMessageCrypto_Sha2DmaRequest*)_createCryptoRequest(
-        dataPtr, WC_HASH_TYPE_SHA512, ctx->cryptoAffinity);
+        dataPtr, hashType, ctx->cryptoAffinity);
```
rizlik: neither `wh_Server_HandleCryptoDmaRequest` nor `wh_Server_HandleCryptoRequest` hanlde SHA512 length variants (https://github.com/wolfSSL/wolfHSM/blob/752ca63dd210636296a9fd93dbe91f87808ba8d3/src/wh_server_crypto.c#L4700-L4711).
This request is never handled by the server.

Test fallback to client side, that's the reason tests are not failing.

test/wh_test_crypto.c
```diff
@@ -1950,6 +1950,149 @@ static int whTest_CryptoSha512(whClientContext* ctx, int devId, WC_RNG* rng)
     }
     return ret;
 }
+
+#if !defined(WOLFSSL_NOSHA512_224)
+static int whTest_CryptoSha512_224(whClientContext* ctx, int devId,
```
rizlik, line 1955: a multiblock test is missing

test/wh_test_crypto.c
```diff
@@ -1950,6 +1950,149 @@ static int whTest_CryptoSha512(whClientContext* ctx, int devId, WC_RNG* rng)
...
+                                   WC_RNG* rng)
+{
+    (void)ctx;
+    (void)rng;
+    int       ret = WH_ERROR_OK;
+    wc_Sha512 sha512[1];
+    /* Buffer sized for the variant, plus canary to detect overwrite */
+    uint8_t   out[WC_SHA512_224_DIGEST_SIZE + 8];
+    const uint8_t canary = 0xA5;
```
rizlik, line 1964: any reason why to rely on canary and not ASAN?

rizlik (changes requested):
As per PR description this was a false positive, but it's probably a better way to handle after the fixes.
Also, will hastype still needed after this PR?
A note:
- the fact that variants were never HSM-accellerated is a false claim, as the server port looks into the hashType and do accelleration if needed.

The fact that tests do not catch the bug because of the fallback mechanism is sad, we should probably address this in wolfHSM (@bigbrett ).

wolfhsm/wh_client_crypto.h
padelsbach: done

src/wh_client_crypto.c
padelsbach: Good catch, fixed

test/wh_test_crypto.c
padelsbach: added

test/wh_test_crypto.c
padelsbach: I felt the programmatic approach was more direct. Can change if you feel strongly

src/wh_client_crypto.c
```diff
@@ -5332,21 +5382,39 @@ int wh_Client_Sha512Dma(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
...
+        case WC_HASH_TYPE_SHA512_224:
+            digestSz = WC_SHA512_224_DIGEST_SIZE;
+            break;
+        case WC_HASH_TYPE_SHA512_256:
+            digestSz = WC_SHA512_256_DIGEST_SIZE;
+            break;
+        default:
+            digestSz = WC_SHA512_DIGEST_SIZE;
+            break;
+    }
```
rizlik, line 5409: I think here we need to add a check that error out if `hashType == WC_HASH_TYPE_SHA512 && sha512->hashType == hashType`).
This protects us from wolfcrypt passing bad wc_sha512 objects and more importantly to avoid the server to write more bytes than the  ones used in the `wh_Client_DmaProcessClientAddress` and related APIs

test/wh_test_crypto.c
```diff
@@ -1950,6 +1950,329 @@ static int whTest_CryptoSha512(whClientContext* ctx, int devId, WC_RNG* rng)
...
+    }
+    else {
+        /* Simulate a wolfSSL port that does not populate hashType,
+         * to verify the wolfHSM client no longer reads it.
+         * Note: this does not catch a broken hashType propagation
+         * to the server because SHA512/224 and /256 are truncations
+         * of SHA512, so software server behavior is unchanged.
+         * It does verify the client uses the dispatcher-supplied
+         * hashType for digest sizing and re-init. */
+        sha512->hashType = 0;
```
rizlik: this causes the ASAN trigger in test. This test shuold fail after the check proposed.

rizlik (changes requested):

src/wh_client_crypto.c
padelsbach: Many existing wolfcrypt ports do not set `hashType` so this check would go against the intent of this PR.

This wolfcrypt PR addresses that, but this wolfHSM change should be independent, imho. https://github.com/wolfSSL/wolfssl/pull/10193

padelsbach: Added a defensive check for the truncated 256 and 224 paths to protect against the overflow. Existing callers should fall into the `default` case -- likely an incorrect hash, but at least no overflow

test/wh_test_crypto.c
padelsbach: removed

src/wh_client_crypto.c
rizlik: if we really want to allow port that do not set hashType we can do `sha512->hashType = hashType` but I probably leaner to be stricter than forgiving. 
If you think that "fixing" `sha512->hashType` for the ports is better, go ahead. 
I don't think this will generate a bad hash btw, the init is not done here and update operation is the same, final is just a truncation. Am I missing something?

rizlik (approved):

rizlik:
@bigbrett ok for me

src/wh_client_crypto.c
padelsbach: Ah, you're right, hash should be fine.

bigbrett:
@padelsbach I'm about to open #337  that is going to change all of this code. My apologies :) 

Perhaps we should just close this?

PR #336, opened 2026-04-15
Add ML-KEM (FIPS 203) support

Frauschi:
Implement full client-server ML-KEM support across all wolfHSM layers, enabling post-quantum key exchange operations to be offloaded to the HSM.

### Client API (wh_client_crypto):
- Key management: import, export, set/get key ID
- Key generation: MakeExportKey (ephemeral) and MakeCacheKey (server-cached)
- Encapsulation and decapsulation operations
- DMA variants for all operations

### Server handling (wh_server_crypto):
- Request handlers for ML-KEM keygen, encapsulate, and decapsulate
- Auto-import with evict-after-use for uncached keys
- DMA request handlers

### Crypto callback integration (wh_client_cryptocb):
- Register PQC KEM keygen/encaps/decaps handlers so wolfCrypt ML-KEM calls are transparently forwarded to the HSM via WH_DEV_ID

### Message layer (wh_message_crypto):
- Define request/response structures for keygen, encapsulate, decapsulate
- Endian translation functions for cross-platform support

### Shared utilities (wh_crypto):
- ML-KEM key serialization/deserialization with automatic level probing

Supports all three ML-KEM parameter sets (512, 768, 1024). Includes tests for all operations and DMA paths, and benchmarks for keygen, encaps, and decaps at each security level.

Also fixes key export response to use actual stored key length from NVM metadata instead of the request size. This is required for the ML-KEM deserialization, as the exact size is needed (all other deserialization methods allow oversized buffers, as DER encoding is used; Hence, this bug never caused issues).
```
@@ -1806,7 +1806,7 @@ int wh_Server_HandleKeyRequest(whServerContext* server, uint16_t magic,
                }

                if (ret == WH_ERROR_OK) {
-                   resp.len = req.key.sz;
+                   resp.len = meta->len;
                    memcpy(resp.label, meta->label, sizeof(meta->label));
                }
```


src/wh_crypto.c
```diff
@@ -379,6 +380,103 @@ int wh_Crypto_MlDsaDeserializeKeyDer(const uint8_t* buffer, uint16_t size,
...
+    }
+
+    /* None of the levels worked, restore original level and devId */
+    wc_MlKemKey_Free(key);
+    (void)wc_MlKemKey_Init(key, origLevel, NULL, origDevId);
+    return ret;
+}
+#endif /* WOLFSSL_HAVE_MLKEM */
+
 #ifdef WOLFSSL_CMAC
```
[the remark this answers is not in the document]
Frauschi: fixed by adding the suggested checks

src/wh_client_crypto.c
```diff
@@ -6443,4 +6455,895 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, MlDsaKey* key,
...
+                        memcpy(ss, resp_data + out_ct_len, out_ss_len);
+                        *inout_ct_len = out_ct_len;
+                        *inout_ss_len = out_ss_len;
+                    }
+                }
+            }
+        }
+        else {
+            ret = WH_ERROR_BADARGS;
+        }
```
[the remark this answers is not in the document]
Frauschi: fixed

src/wh_server_crypto.c
```diff
@@ -4359,6 +4440,323 @@ static int _HandleMlDsaCheckPrivKey(whServerContext* ctx, uint16_t magic,
...
+
+    if (!WH_KEYID_ISERASED(key_id)) {
+        ret = wh_Server_KeystoreFindEnforceKeyUsage(ctx, key_id,
+                                                    WH_NVM_FLAGS_USAGE_DERIVE);
+        if (ret != WH_ERROR_OK) {
+            goto cleanup;
+        }
+    }
+
+    if (!_IsMlKemLevelSupported((int)req.level)) {
```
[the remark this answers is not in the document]
Frauschi: `wc_MlKemKey_Free()` zeros all sensitive key material, so no `wc_ForceZero()` calls are required.

src/wh_server_crypto.c
```diff
@@ -4359,6 +4440,323 @@ static int _HandleMlDsaCheckPrivKey(whServerContext* ctx, uint16_t magic,
...
+    }
+
+    available = inSize - sizeof(whMessageCrypto_MlKemDecapsRequest);
+    if (req.ctSz > available) {
+        ret = WH_ERROR_BADARGS;
+        goto cleanup;
+    }
+    req_ct = (byte*)cryptoDataIn + sizeof(whMessageCrypto_MlKemDecapsRequest);
+
+    ret = wc_MlKemKey_Init(key, (int)req.level, NULL, devId);
```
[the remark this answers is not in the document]
Frauschi: fixed

src/wh_server_crypto.c
```diff
@@ -5702,6 +6125,397 @@ static int _HandlePqcSigAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+            break;
+    }
+
+    return ret;
+}
+#endif /* WOLFSSL_HAVE_MLKEM */
+
 #if defined(WOLFSSL_CMAC) && !defined(NO_AES) && defined(WOLFSSL_AES_DIRECT)
 static int _HandleCmacDma(whServerContext* ctx, uint16_t magic, int devId,
                           uint16_t seq, const void* cryptoDataIn,
```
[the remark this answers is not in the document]
Frauschi: `wc_MlKemKey_Free()` zeros all sensitive key material, so no `wc_ForceZero()` calls are required.

bigbrett (changes requested):
Quick request before diving too deep into the review - can you add the new message structs to the struct padding check test?

Frauschi:
Added the new structures (and also missing ML-DSA ones) to the test.

Frauschi:
This work is orthogonal to #346, so this PR is missing support for the new ExportPublicKey API of #346. Depending on what goes into main first, the other one needs some updates.

src/wh_client_crypto.c
```diff
@@ -6443,4 +6455,920 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, MlDsaKey* key,
...
+    ret = wc_MlKemKey_PrivateKeySize(key, &allocSz);
+    if (ret != 0) {
+        /* Fall back to public key size if no private key */
+        ret = wc_MlKemKey_PublicKeySize(key, &allocSz);
+    }
+    if (ret != 0 || allocSz == 0) {
+        return WH_ERROR_BADARGS;
+    }
+
+    buffer = (byte*)XMALLOC(allocSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
```
bigbrett: We currently have a policy of no dynamic allocations in the wolfHSM library. Obviously wolfCrypt does allocations internally, but we want to limit the allocation boundary to that. Can you make this a statically allocated stack buffer plz. Look at how ML-DSA does it

src/wh_client_crypto.c
```diff
@@ -6443,4 +6455,920 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, MlDsaKey* key,
...
+{
+    int      ret = WH_ERROR_OK;
+    byte*    buffer = NULL;
+    uint16_t buffer_len = WC_ML_KEM_MAX_PRIVATE_KEY_SIZE;
+
+    if ((ctx == NULL) || WH_KEYID_ISERASED(keyId) || (key == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    buffer = (byte*)XMALLOC(WC_ML_KEM_MAX_PRIVATE_KEY_SIZE, NULL,
```
bigbrett: no dynamic allocation

bigbrett (changes requested):

wolfhsm/wh_message_crypto.h
```diff
@@ -947,6 +947,92 @@ int wh_MessageCrypto_TranslateMlDsaVerifyResponse(
...
+    whMessageCrypto_MlKemKeyGenRequest* dest);
+
+int wh_MessageCrypto_TranslateMlKemKeyGenResponse(
+    uint16_t magic, const whMessageCrypto_MlKemKeyGenResponse* src,
+    whMessageCrypto_MlKemKeyGenResponse* dest);
+
+/* ML-KEM Encapsulation Request */
+typedef struct {
+    uint32_t options;
+#define WH_MESSAGE_CRYPTO_MLKEM_ENCAPS_OPTIONS_EVICT (1 << 0)
```
padelsbach, line 1057: Can this bitmask live somewhere outside the struct?

bigbrett: @padelsbach no reason it couldn't but he is following the existing pattern where the bitmask is defined just like this. It is admittedly not common to do this but I think it is kind of nice to have it tethered to the definition of the struct in which it is relevant `¯\_(ツ)_/¯`. Recommend just keeping as-is for now unless we relocate all of them

src/wh_client_crypto.c
Frauschi: Fixed

src/wh_client_crypto.c
Frauschi: Fixed all occurrences.

wolfhsm/wh_message_crypto.h
Frauschi: I'm also not a fan of these inline definitions, but as Brett suggested, this should go into its own PR when we really want to move these in the future. 

I kept them as is for now.

Frauschi:
Rebased to main and fixed the dynamic memory allocation comments.

Frauschi:
Rebased and added support for `ExportPublicKey()` functionality after #346 has landed.

bigbrett (dismissed):

bigbrett:
LGTM. @padelsbach over to you for final review and merge

CLAUDE.md
```diff
@@ -0,0 +1,99 @@
+# CLAUDE.md
```
padelsbach: This is good, but probably belongs in a separate PR

wolfhsm/wh_client_crypto.h
```diff
@@ -2207,5 +2208,257 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, MlDsaKey* key,
...
+ * @param[in] key Pointer to the ML-KEM key (must have key ID set).
+ * @param[out] ct Buffer to receive the ciphertext.
+ * @param[in,out] inout_ct_len On input, size of ct buffer; on output, actual
+ *                ciphertext length.
+ * @param[out] ss Buffer to receive the shared secret.
+ * @param[in,out] inout_ss_len On input, size of ss buffer; on output, actual
+ *                shared secret length.
+ * @return int Returns 0 on success or a negative error code on failure.
+ */
+int wh_Client_MlKemEncapsulate(whClientContext* ctx, MlKemKey* key,
```
padelsbach, line 2846: The new ML-KEM APIs extend the pattern from ML-DSA of using `word32`, `byte`, etc rather than `uint32_t` and `uint8_t`, etc. In fact, this change uses a mix of both. 

From my reading, wolfHSM largely uses the `uint` style, so I don't think we should propagate the use of `word` and `byte`. And probably go back and fix ML-DSA (in a follow-up PR).

src/wh_client_crypto.c
```diff
@@ -8103,4 +8115,900 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, MlDsaKey* key,
...
+    ret = wh_Client_SendRequest(ctx, group, action, req_len,
+                                (uint8_t*)dataPtr);
+    WH_DEBUG_CLIENT_VERBOSE("MlKemMakeKey: Req sent:level:%d, ret:%d\n",
+                            level, ret);
+    if (ret != WH_ERROR_OK) {
+        return ret;
+    }
+
+    do {
+        ret = wh_Client_RecvResponse(ctx, &group, &action, &res_len,
```
padelsbach, line 9187: Please add checking of `res_len`. Existing pattern is: 

```
        ret = _getCryptoResponse(dataPtr, WC_CIPHER_AES_CTR, (uint8_t**)&res);
        if (ret == WH_ERROR_OK) {
            const uint32_t hdr_sz =
                sizeof(whMessageCrypto_GenericResponseHeader) + sizeof(*res) +
                (2 * AES_BLOCK_SIZE);
            /* Trailing payload is: output (res->sz) + reg (AES_BLOCK_SIZE)
             * + tmp (AES_BLOCK_SIZE) */
            if (res_len < hdr_sz || res->sz > (res_len - hdr_sz)) {
                ret = WH_ERROR_ABORTED;
            }
```

Similar checks are needed in `wh_Client_MlKemEncapsulate` `wh_Client_MlKemDecapsulate` `wh_Client_MlKemEncapsulateDma` `wh_Client_MlKemDecapsulateDma`

src/wh_client_cryptocb.c
```diff
@@ -600,6 +625,167 @@ int wh_Client_CryptoCb(int devId, wc_CryptoInfo* info, void* inCtx)
...
+
+    /* Extract info parameters */
+    int   size = info->pk.pqc_kem_kg.size;
+    void* key  = info->pk.pqc_kem_kg.key;
+    int   type = info->pk.pqc_kem_kg.type;
+
+#ifndef WOLFHSM_CFG_DMA
+    if (useDma) {
+        /* TODO: proper error code? */
+        return WC_HW_E;
```
padelsbach: Just curious: why do we not return `CRYPTO_UNAVAILABLE` here to allow the software fallback?

src/wh_crypto.c
```diff
@@ -379,6 +381,110 @@ int wh_Crypto_MlDsaDeserializeKeyDer(const uint8_t* buffer, uint16_t size,
...
+            ret = wc_MlKemKey_EncodePublicKey(key, buffer, keySize);
+        }
+    }
+
+    if (ret == WH_ERROR_OK) {
+        *out_size = (uint16_t)keySize;
+    }
+    else {
+        /* Clear buffer to avoid leaking partial key material on error */
+        wc_ForceZero(buffer, keySize);
```
padelsbach, line 421: AI told me that `keySize` can be uninitialized here. Just init it to 0 to be safe.

src/wh_crypto.c
```diff
@@ -379,6 +381,110 @@ int wh_Crypto_MlDsaDeserializeKeyDer(const uint8_t* buffer, uint16_t size,
...
+    }
+
+    return ret;
+}
+
+int wh_Crypto_MlKemDeserializeKey(const uint8_t* buffer, uint16_t size,
+                                  MlKemKey* key)
+{
+    static const int levels[] = {
```
padelsbach: Nice, I like this pattern of separating data from code. And `static const` is a nice touch as well.

Consider using `uint8_t` as the type here to save a few bytes.

src/wh_crypto.c
```diff
@@ -379,6 +381,110 @@ int wh_Crypto_MlDsaDeserializeKeyDer(const uint8_t* buffer, uint16_t size,
...
+    if (ret == WH_ERROR_OK) {
+        return ret;
+    }
+    ret = wc_MlKemKey_DecodePublicKey(key, buffer, size);
+    if (ret == WH_ERROR_OK) {
+        return ret;
+    }
+
+    /* Current level didn't work, try other levels in place */
+    for (i = 0; i < (word32)(sizeof(levels) / sizeof(levels[0])); i++) {
```
padelsbach: Consider using `XELEM_CNT(levels)` for readability

src/wh_crypto.c
```diff
@@ -379,6 +381,110 @@ int wh_Crypto_MlDsaDeserializeKeyDer(const uint8_t* buffer, uint16_t size,
...
+            continue;
+        }
+        wc_MlKemKey_Free(key);
+        ret = wc_MlKemKey_Init(key, levels[i], origHeap, origDevId);
+        if (ret != WH_ERROR_OK) {
+            continue;
+        }
+        ret = wc_MlKemKey_DecodePrivateKey(key, buffer, size);
+        if (ret == WH_ERROR_OK) {
+            return ret;
```
padelsbach, line 475: Does the caller need to be informed that we may have changed the level?

CLAUDE.md
Frauschi: oh wow, that has actually been added accidentally. Thanks for pointing that out - removed the file.

But yeah, probably worth adding one in a separate PR for the project.

wolfhsm/wh_client_crypto.h
Frauschi: Fixed for all ML-KEM related code.

src/wh_client_crypto.c
Frauschi: Added the checks.

src/wh_client_cryptocb.c
Frauschi: Great point, I just followed the pattern of the other algorithms. But your idea of triggering a fallback to software is actually the better way here I think. Changed it for the ML-KEM functions.

src/wh_crypto.c
Frauschi: Done

src/wh_crypto.c
Frauschi: moved to `uint8_t`.

src/wh_crypto.c
Frauschi: Done

src/wh_crypto.c
Frauschi: I added a comment in the code to hint that the level may change. But I am not convinced that we should actively inform that via a new return code or something, as ultimately the operation is considered "successful" in that case.

padelsbach (dismissed):
Thanks for the updates @Frauschi. I'll give @bigbrett time to take another look before merging.

src/wh_client_cryptocb.c
```diff
@@ -752,8 +752,8 @@ static int _handlePqcDecaps(whClientContext* ctx, wc_CryptoInfo* info,
 
 #ifndef WOLFHSM_CFG_DMA
     if (useDma) {
-        /* TODO: proper error code? */
-        return WC_HW_E;
+        /* DMA support not compiled in - let wolfCrypt fall back to software */
+        return CRYPTOCB_UNAVAILABLE;
```
bigbrett: You had it right the first time. If you use the DMA-specific devId for an algorithm that doesn't support it we should just hard error -- you are essentially "using the wrong devId" so falling back to software is not the appropriate solution. User should get an error in this regard. 

Related: this also reminds me that software fallback is NEVER the right option and I hate that we do that. Needs to change. IMO explicitly passing an HSM devId should "hard fail" if the algo isn't supported instead of falling back to software wolfCrypt and failing or having hard-to-debug issues since your key material isn't local. I need to change that globally. However that is outside the scope of this PR.

src/wh_client_cryptocb.c
bigbrett: I actually don't think that is correct - currently we fallback to software for unsupported algorithms (we really shouldn't but that is another story), however in this case the user has essentially passed the wrong devId in. This should be a hard failure.

wolfhsm/wh_client_crypto.h
bigbrett: Agreed for client side. wolfCrypt types need to be used at the wolfCrypt API boundary but typecasting function arguments should be okay there. May be some scenarios that have odd compiler warnings otherwise

bigbrett (changes requested):

src/wh_client_cryptocb.c
Frauschi: After thinking this through again, you are absolutely right. I reverted that change.

padelsbach: > passing an HSM devId should "hard fail" if the algo isn't supported instead of falling back to software

@bigbrett, just curious, is this true for the server side too, or just the client?

padelsbach (approved):

src/wh_client_cryptocb.c
bigbrett: @padelsbach  gr8 question, just client for now

bigbrett (approved):

PR #337, opened 2026-04-16
Async crypto v2

bigbrett:
# Async Crypto API for SHA

Introduces a non-blocking request/response split for wolfHSM crypto operations, starting with the SHA family. Every blocking call gains a `*Request()` / `*Response()` pair; `Request()` sends and returns immediately, `Response()` polls once and returns `WH_ERROR_NOTREADY` until the server replies. Existing blocking wrappers and the `wh_Client_CryptoCb` path are unchanged, so current application code keeps working.

## Summary
- New async API surface in `wh_client_crypto.h` covering SHA-1/224/256/384/512
- Refactored `wh_client_crypto.c` around shared request/response primitives
- Updated message layouts in `wh_message_crypto.{c,h}` to carry async-specific state (intermediate hash state, trailing input, DMA metadata)
- Server-side routing updates in `wh_server_crypto.c` (requests handled statelessly per design)
- Design doc: `docs/draft/async-crypto.md` (includes roadmap for remaining algorithms)
- Extensive new tests in `test/wh_test_crypto.c`
- Drive-by: fix gcov flag for gcov bug in `code-coverage.yml`

## Future work

- other algorithms
- pending request tracking in comm layer enforcing single-request in flight at a time, as well as remote server reset functionality allowing for re-synchronization in case the server locks up


src/wh_client_crypto.c
```diff
@@ -4331,564 +4347,538 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
...
+    }
+
+    /* Populate fixed request fields */
+    req->isLastBlock = 0;
+    req->inSz        = wirePos;
+    memcpy(req->resumeState.hash, sha->digest, WC_SHA256_DIGEST_SIZE);
+    req->resumeState.hiLen = sha->hiLen;
+    req->resumeState.loLen = sha->loLen;
 
+    ret = wh_Client_SendRequest(ctx, WH_MESSAGE_GROUP_CRYPTO, WC_ALGO_TYPE_HASH,
```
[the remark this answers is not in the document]
bigbrett: this is way out of scope

src/wh_server_crypto.c
```diff
@@ -5137,142 +5129,128 @@ static int _HandleSha512Dma(whServerContext* ctx, uint16_t magic, int devId,
...
 
-    /* Translate the response */
     (void)wh_MessageCrypto_TranslateSha2DmaResponse(
         magic, &res, (whMessageCrypto_Sha2DmaResponse*)cryptoDataOut);
     *outSize = sizeof(res);
 
-    /* return value populates rc in response message */
     return ret;
 }
 #endif /* WOLFSSL_SHA512 */
```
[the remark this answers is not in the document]
bigbrett: confused - it seems this check is in every sha handler
```
if ((uint32_t)req.inSz >
    (uint32_t)(inSize - sizeof(whMessageCrypto_Sha256Request))) {
    return WH_ERROR_BADARGS;
}
```

src/wh_client_crypto.c
```diff
@@ -4310,17 +4296,47 @@ int wh_Client_CmacDma(whClientContext* ctx, Cmac* cmac, CmacType type,
...
+{
+    int                            ret = 0;
+    whMessageCrypto_Sha256Request* req = NULL;
+    uint8_t*                       inlineData;
+    uint8_t*                       dataPtr = NULL;
+    uint8_t*                       sha256BufferBytes;
+    uint32_t                       capacity;
+    uint32_t                       wirePos = 0;
+    uint32_t                       i       = 0;
+    /* Snapshot of buffer state for rollback if SendRequest fails */
```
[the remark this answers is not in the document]
bigbrett: this is overkill since HMAC usage has its own cryptocb path. This information cannot be treated as sensitive w.r.t. the client.

src/wh_client_crypto.c
```diff
@@ -4310,17 +4296,47 @@ int wh_Client_CmacDma(whClientContext* ctx, Cmac* cmac, CmacType type,
...
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Empty update: nothing to send, no state to mutate. */
+    if (inLen == 0) {
+        return WH_ERROR_OK;
+    }
 
     /* Get data buffer */
     dataPtr = wh_CommClient_GetDataPtr(ctx->comm);
```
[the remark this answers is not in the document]
bigbrett: The whole point of the DMA path is that input travels by address, not through the comm buffer.

src/wh_client_crypto.c
```diff
@@ -4310,17 +4296,47 @@ int wh_Client_CmacDma(whClientContext* ctx, Cmac* cmac, CmacType type,
...
 {
-    uint16_t                        group   = WH_MESSAGE_GROUP_CRYPTO;
-    uint16_t                        action  = WH_MESSAGE_ACTION_NONE;
-    int                             ret     = 0;
-    uint16_t                        dataSz  = 0;
-    whMessageCrypto_Sha256Request*  req     = NULL;
-    whMessageCrypto_Sha2Response*   res     = NULL;
-    uint8_t*                        dataPtr = NULL;
+    return WH_MESSAGE_CRYPTO_SHA256_MAX_INLINE_UPDATE_SZ +
+           (uint32_t)(WC_SHA256_BLOCK_SIZE - 1u - sha->buffLen);
```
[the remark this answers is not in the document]
bigbrett: fixed by buflen check added for first issue

src/wh_client_crypto.c
```diff
@@ -4331,564 +4347,538 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
...
+    uint8_t*                       inlineData;
+    uint8_t*                       dataPtr;
+
+    if (ctx == NULL || sha == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    dataPtr = wh_CommClient_GetDataPtr(ctx->comm);
+    if (dataPtr == NULL) {
+        return WH_ERROR_BADARGS;
```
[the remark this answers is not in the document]
bigbrett: incorrect, can't actually overflow based on other checking logic.

src/wh_client_crypto.c
```diff
@@ -4331,564 +4347,538 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
...
-    WH_DEBUG_CLIENT_VERBOSE("send SHA256 Req:\n");
-    WH_DEBUG_VERBOSE_HEXDUMP("[client] inBlock: ", req->inBlock, WC_SHA256_BLOCK_SIZE);
-    if (req->resumeState.hiLen != 0 || req->resumeState.loLen != 0) {
-        WH_DEBUG_VERBOSE_HEXDUMP("  [client] resumeHash: ", req->resumeState.hash,
-                         (isLastBlock) ? req->lastBlockLen
-                                       : WC_SHA256_BLOCK_SIZE);
-        WH_DEBUG_CLIENT_VERBOSE("  hiLen: %u, loLen: %u\n",
-               (unsigned int)req->resumeState.hiLen,
-               (unsigned int)req->resumeState.loLen);
+    ret = wh_Client_RecvResponse(ctx, &group, &action, &dataSz, dataPtr);
```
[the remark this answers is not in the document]
bigbrett: overkill, as the SHA state is directly populated by the client, but added a defense in depth check

src/wh_client_crypto.c
```diff
@@ -4331,564 +4347,538 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
...
+            uint32_t capacity  = _Sha256UpdatePerCallCapacity(sha256);
+            uint32_t remaining = inLen - consumed;
+            uint32_t chunk     = (remaining < capacity) ? remaining : capacity;
+            bool     sent      = false;
+
+            ret = wh_Client_Sha256UpdateRequest(ctx, sha256, in + consumed,
+                                                chunk, &sent);
+            if (ret != WH_ERROR_OK) {
+                break;
             }
```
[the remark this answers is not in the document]
bigbrett: This is true, but is unfortunately a consequence of having no comm layer awareness of "request pending" as this behaves dramatically differently depending on the underlying transport. Punting this for a future feature.

wolfhsm/wh_message_crypto.h
```diff
@@ -740,48 +740,104 @@ int wh_MessageCrypto_TranslateEd25519VerifyResponse(
...
+ * updates always carry whole blocks. */
+#define WH_MESSAGE_CRYPTO_SHA256_MAX_INLINE_UPDATE_SZ           \
+    (((WOLFHSM_CFG_COMM_DATA_LEN -                              \
+       (uint32_t)sizeof(whMessageCrypto_GenericRequestHeader) - \
+       (uint32_t)sizeof(whMessageCrypto_Sha256Request)) /       \
+      64u) *                                                    \
+     64u)
+
+WH_UTILS_STATIC_ASSERT(WH_MESSAGE_CRYPTO_SHA256_MAX_INLINE_UPDATE_SZ >= 64u,
+                       "Comm buffer too small to fit a SHA256 block");
```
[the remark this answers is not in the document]
bigbrett: comm buffer cant be this small.

wolfhsm/wh_message_crypto.h
```diff
@@ -966,26 +1022,61 @@ typedef struct {
...
-    /* Since client addresses are subject to DMA checking, we can't use them to
-     * determine the requested operation (update/final). Therefore we need to
-     * indicate to the server which SHA224 operation to perform */
-    uint64_t                  finalize;
+    struct {
+        uint32_t hiLen;
+        uint32_t loLen;
+        uint8_t  hash[64]; /* WC_SHA512_DIGEST_SIZE */
+        uint32_t hashType;
+        uint8_t  WH_PAD[4];
```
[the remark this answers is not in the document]
bigbrett: that is the point

examples/posix/wh_posix_cfg.h
```diff
@@ -23,9 +23,11 @@
...
  * =========================================== */
 
-/* Request and Response Buffer Sizes */
-#define WH_POSIX_REQ_SIZE 2048
-#define WH_POSIX_RESP_SIZE 2048
+/* Request and Response Buffer Sizes. Must be at least
+ * sizeof(whTransportMemCsr) + sizeof(whCommHeader) + WOLFHSM_CFG_COMM_DATA_LEN
+ * to carry a full comm packet through the SHM transport. */
+#define WH_POSIX_REQ_SIZE (16 + (1024 * 8))
+#define WH_POSIX_RESP_SIZE (16 + (1024 * 8))
```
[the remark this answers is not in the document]
bigbrett: dont want to pull a header dependency into this config header

wolfhsm/wh_message_crypto.h
```diff
@@ -740,48 +740,104 @@ int wh_MessageCrypto_TranslateEd25519VerifyResponse(
...
 
+/* Maximum number of input bytes that can be carried inline (after the generic
+ * crypto request header and the Sha256Request struct) in a single comm-buffer
+ * message, rounded down to a multiple of WC_SHA256_BLOCK_SIZE so non-final
+ * updates always carry whole blocks. */
+#define WH_MESSAGE_CRYPTO_SHA256_MAX_INLINE_UPDATE_SZ           \
+    (((WOLFHSM_CFG_COMM_DATA_LEN -                              \
+       (uint32_t)sizeof(whMessageCrypto_GenericRequestHeader) - \
+       (uint32_t)sizeof(whMessageCrypto_Sha256Request)) /       \
+      64u) *                                                    \
```
Frauschi, line 777: Why not use `WC_SHA256_BLOCK_SIZE` here directly? Would make the code more understandable I think. Same goes for SHA512 below

docs/draft/async-crypto.md
```diff
@@ -0,0 +1,470 @@
...
+    sha->buffLen = savedBuffLen;
+    memcpy(sha->buffer, savedBuffer, savedBuffLen);
+}
+```
+
+### Finalization
+
+The Final request sends whatever partial data remains in the client's buffer
+(0 to `BLOCK_SIZE - 1` bytes) with `isLastBlock = 1`.  The server handles
+MD5-style padding and produces the final digest.  The Final response copies the
```
rizlik: drop the "MD5-style"

docs/draft/async-crypto.md
```diff
@@ -0,0 +1,470 @@
...
+available.  The DMA variant differs from the inline variant in how bulk data
+reaches the server:
+
+- **Inline (non-DMA)**: all input data is copied into the comm buffer message.
+- **DMA**: whole blocks are referenced by address via a `DmaBuffer` descriptor
+  (the server reads them directly from client memory).  Only the assembled first
+  block (from the partial buffer) or the final tail travels inline.
+
+The hash state (`resumeState`) always travels **inline**, not via DMA, for
+cross-architecture concerns (endian translation, address space isolation).
```
rizlik: imo drop "address space isolation"

docs/draft/async-crypto.md
```diff
@@ -0,0 +1,470 @@
...
+| ECDH              | `wh_Client_EccSharedSecret{Request,Response}` | Low     | Single-shot |
+| Curve25519        | `wh_Client_Curve25519SharedSecret{Request,Response}` | Low | Single-shot |
+| Ed25519 Sign      | `wh_Client_Ed25519Sign{Request,Response}`  | Low        | Single-shot |
+| Ed25519 Verify    | `wh_Client_Ed25519Verify{Request,Response}`| Low        | Single-shot |
+| CMAC              | `wh_Client_Cmac{Request,Response}`         | Low        | Already has partial split pattern |
+| ML-DSA Sign       | `wh_Client_MlDsaSign{Request,Response}`    | Low        | Post-quantum; single-shot |
+| ML-DSA Verify     | `wh_Client_MlDsaVerify{Request,Response}`  | Low        | Post-quantum; single-shot |
+| RNG Generate      | `wh_Client_RngGenerate{Request,Response}`  | Medium     | Chunking needed for large requests; async callers must handle chunking themselves |
+
+Most remaining algorithms are **single-shot** operations (one request, one
```
rizlik, line 455: is this true also for `AES-GCM`, `AES-CTR`?

src/wh_client_crypto.c
```diff
@@ -4897,554 +4887,1579 @@ static int _xferSha384BlockAndUpdateDigest(whClientContext* ctx,
...
+                     uint32_t inLen, uint8_t* out)
+{
+    int ret = WH_ERROR_OK;
+
+    /* Caller invoked SHA Update:
+     * wc_CryptoCb_Sha512Hash(sha512, data, len, NULL) */
+    if (in != NULL && inLen > 0) {
+        uint32_t consumed = 0;
+        while (ret == WH_ERROR_OK && consumed < inLen) {
+            uint32_t capacity  = _Sha512UpdatePerCallCapacity(sha512);
```
rizlik, line 6290: how the app knows which is the outbound limit? Should it relies on `WH_ERROR_BADARGS`. 
Should the app chunk the data on its own? 
I do this is a real problem, this limit can easily go unnoticed by the app in design/testing phase but it can break later in production.

rizlik (changes requested):
Thanks for the clear PR and docs.
My main feedback/concerns:

a) absence of a way for the client to track the in-flight request. PR acknowledge that, nevertheless it should be one of the main design choices of the API.  The PR should either add it here or at least sketch the design in the docs. I don't think  offloading the management of the req/resp tracking (is a request in flight? what kind? etc) fully to the application is a good design. At least we should protect the application and error out on bad API usage (eg, two request() without response(), request() and response() of mismatching type, etc).


b) redundancy of the code in `wh_client_crypto.c`: this is a pre-existing problem for sure, but can we reduce the amount of almost exact code across both sha variants and DMA vs non-DMA paths?

c) API usability: how the app knows the outbound limit? (see comments below). Should we switch on blocking behavior for large input size? or should we add a specific error message? or should we expose the function that compute the outbound limit?  


Is using a single API (*Request or a more general *Operation) instead of two (*Request/*Response) worth considering? 
It will solve also the cumbersome `requestSent`  (the API can return `WH_ERROR_OK` directly),  and it can address (c) by chunking at the wolfHSM layer big request across multiple client <-> server roundtrips.
In this case the app just keep calling the API until an `ret != WH_ERROR_NOTREADY`, following a POSIX-like semantic.

src/wh_client_crypto.c
```diff
@@ -4331,564 +4351,1641 @@ static int _xferSha256BlockAndUpdateDigest(whClientContext* ctx,
...
+    }
+    else {
+        /* Rollback buffer state and release DMA on failure */
+        sha->buffLen = savedBuffLen;
+        memcpy(sha256BufferBytes, savedBuffer, savedBuffLen);
+        if (inAddrAcquired) {
+            (void)wh_Client_DmaProcessClientAddress(
+                ctx, (uintptr_t)dmaBase, (void**)&inAddr, dmaSz,
+                WH_DMA_OPER_CLIENT_READ_POST, (whDmaFlags){0});
+        }
```
Frauschi, line 4707: Should ctx->dma.asyncCtx.sha with its variables be cleared here, too?

src/wh_server_crypto.c
```diff
@@ -3961,17 +4028,17 @@ static int _HandleSha512(whServerContext* ctx, uint16_t magic, int devId,
...
+            /* Post-condition: non-final updates MUST leave buffLen == 0,
+             * since we validated inSz is a multiple of the block size. */
+            if (sha512->buffLen != 0) {
+                ret = WH_ERROR_ABORTED;
+            }
+            else {
+                memcpy(res.hash, sha512->digest, WC_SHA512_DIGEST_SIZE);
+                res.loLen = sha512->loLen;
+                res.hiLen = sha512->hiLen;
+            }
```
Frauschi, line 4053: res.hashType is never set. Should be set to the one actually be used for the SHA object. Then the client can check the sent one in the request against the received one in the response to catch misconfiguration (e.g. client is built with SHA512_256 support enabled, server is not; this would cause a mismatch here).

src/wh_client_crypto.c
```diff
@@ -4898,552 +5995,526 @@ static int _xferSha384BlockAndUpdateDigest(whClientContext* ctx,
...
+
+    ret = wh_Client_RecvResponse(ctx, &group, &action, &dataSz, dataPtr);
+    if (ret != 0) {
+        return ret;
+    }
+
+    ret = _getCryptoResponse(dataPtr, WC_HASH_TYPE_SHA512, (uint8_t**)&res);
+    if (ret >= 0) {
+        /* keep hashtype before initialization */
+        hashType = sha->hashType;
```
Frauschi, line 6246: Match this with the hashType received in the response.

Frauschi (dismissed):
Only smaller issues building on top of Marco's comments.

docs/draft/async-crypto.md
Frauschi: `AES-GCM` is one-shot at the moment, as we currently do not support the streaming API of wolfCrypt. However, both `AES-CTR` and `AES-CBC` can be used in an incremental streaming-like manner (calling `encrypt()`/`decrypt()` multiple times in sequence to stepwise process a large chunk of data). But for that, the internal state of the `Aes` object is already properly exchanged between client and server (added in #282). So adding the *Request / *Response interface for them should be straightforward still.

examples/posix/wh_posix_client/wolfhsm_cfg.h
```diff
@@ -32,7 +32,7 @@
 /** wolfHSM settings */
 #define WOLFHSM_CFG_ENABLE_CLIENT
 #define WOLFHSM_CFG_HEXDUMP
-#define WOLFHSM_CFG_COMM_DATA_LEN 5000
+#define WOLFHSM_CFG_COMM_DATA_LEN (1024 * 8)
```
AlexLanzano, line 35: just curious about why this was increased

src/wh_client_crypto.c
AlexLanzano: Second this.  The context should be cleared in the case where the caller doesnt check the error here and calls UpdateResponse anyway.

docs/draft/async-crypto.md
```diff
@@ -0,0 +1,470 @@
...
+/* ... do other work while server computes ... */
+
+/* Poll for completion */
+do {
+    ret = wh_Client_EccSignResponse(ctx, sig, &sigLen);
+    if (ret == WH_ERROR_NOTREADY) {
+        /* yield to scheduler, do other work, etc. */
+    }
+} while (ret == WH_ERROR_NOTREADY);
+/* ret has final result, sig/sigLen are populated */
```
AlexLanzano, line 99: Out of scope for this PR but we should think about how we could allow the user to wire up an interrupt+callback approach here instead polling. I.E. somehow fire an interrupt when the client receives the response, have the ISR call the user defined callback.

wolfhsm/wh_message_crypto.h
bigbrett: because I didn't want `wh_message_crypto.h` to have to `#include` a wolfCrypt header, which blows up a bunch of stuff in our padding checks. Annoying, but necessary.

docs/draft/async-crypto.md
```diff
@@ -0,0 +1,470 @@
...
+### Design Tradeoffs
+
+| Decision | Tradeoff |
+|----------|----------|
+| **State on wire** | Larger messages (~40-84 bytes overhead), but the server is fully stateless and needs no per-client hash context |
+| **Whole-block alignment** | Wastes up to `BLOCK_SIZE - 1` bytes of comm buffer capacity per message, but guarantees the server never has a partial block (simplifies server logic and invariant checking) |
+| **Client-side partial buffering** | Requires wolfCrypt's buffer/buffLen fields, but avoids allocating separate storage and enables the `requestSent` optimization for small inputs |
+| **Per-call capacity limit** | Callers of the async API must respect the capacity and chunk large inputs themselves (the blocking wrapper handles this automatically), but each call is bounded and predictable |
+| **`requestSent` flag** | Adds a parameter to the API, but avoids unnecessary round-trips when input is absorbed entirely into the local buffer |
+| **Snapshot/rollback on send failure** | Small CPU cost to copy the partial buffer, but guarantees SHA state consistency even on transport failures |
```
AlexLanzano, line 417: May just want to sanity check any performance gain/degradation from these changes?

Out of scope for this PR but it would be great to have a CI test to run a benchmark on posix and record performance changes

docs/draft/async-crypto.md
AlexLanzano: I think it could be nice to still allow for streaming in cases where the underlying HW accelerator supports it. There might be cases where the customer would want this capability and not be bounded by the wolfssl api.

bigbrett: yeah not an exhaustive list, sorry. Streaming GCM should definitely be supported eventually

src/wh_client_crypto.c
bigbrett: this is a good point. Per our discussion earlier, I think this unfortunately might need to be a per-request client API for those that support  a variable/max capacity, since there might not be a uniform way to do it. Thankfully its probably a small API surface. I'll think it it. Might want to punt it though

docs/draft/async-crypto.md
bigbrett: fixed

docs/draft/async-crypto.md
bigbrett: fixed

src/wh_client_crypto.c
bigbrett: fixed

src/wh_server_crypto.c
bigbrett: fixed

src/wh_client_crypto.c
bigbrett: fixed

examples/posix/wh_posix_client/wolfhsm_cfg.h
bigbrett: Changes to upstream dilithium required larger messages.

docs/draft/async-crypto.md
bigbrett: agree. Currently nothing prohibits this from being deployed outside the context of wolfHSM if it is baked into the transport - e.g. wrap SHM transport with an interrupt such that client can signal IRQ to server when it "sends" something, then server can `_WFI()` then poll the transport in the ISR (or defer to background) once it has awoken. And same in reverse order for server->client, though that one would need to somehow communicate which response to poll (or have it be tracked entirely on client side, since only one in flight at a time). 

Can think of a few ways to do it. Like you said - food for a future PR.

docs/draft/async-crypto.md
bigbrett: This documentation was kinda AI slop, sry.

Imagine there would only be an improvement, since the old blocking method chunked a single SHA block at a time, whereas now you can send much more data in one go. Not sure how it would degrade.

Agreed about automated benchmarks - ideally we would turn on all the X86 optimizations too. Right now its pretty dumb and slow.

bigbrett:
@rizlik @Frauschi @AlexLanzano Thanks all for the feedback! Believe I addressed all the pending comments in my last few commits - pending the items we discussed punting. The big change is we now track request in flight and have a test for it, making sure the client APIs should error out if misused.

docs/draft/async-crypto.md
AlexLanzano: Created new issue to track this out of scope request

docs/draft/async-crypto.md
AlexLanzano: Created new issue to track this out of scope request

AlexLanzano (dismissed):

src/wh_comm.c
```diff
@@ -170,14 +184,43 @@ int wh_CommClient_RecvResponse(whCommClient* context,
...
+                }
+                if (out_seq != NULL) {
+                    *out_seq = 0;
+                }
+                return WH_ERROR_ABORTED;
+            }
+
+            /* Validate sequence number. A sequence mismatch indicates a
+             * well-formatted but stale or unsolicited message (e.g. a request
+             * the caller abandoned). Since nly one pending request is allowed
```
rizlik: typo only

src/wh_client_crypto.c
```diff
@@ -6190,6 +6238,13 @@ int wh_Client_Sha512FinalResponse(whClientContext* ctx, wc_Sha512* sha,
     if (ret >= 0) {
         /* keep hashtype before initialization */
         hashType = sha->hashType;
+        /* Detect server/client mismatch (e.g. client built with SHA512_256
+         * support but server not — server would fall back to plain SHA512 and
+         * return a digest that does not correspond to the requested variant).
+         */
+        if (res->hashType != (uint32_t)hashType) {
```
rizlik: Here we can probably just copy the truncated output instead of error out (the init state is always provided by the client on the wire)

bigbrett: fixed here and for DMA path as well

src/wh_comm.c
bigbrett: fixed

rizlik (approved):

AlexLanzano (approved):

PR #339, opened 2026-04-16
add wolfBoot image verification to image manager

bigbrett:
## Add wolfBoot image verification to server image manager

Extends the server image manager to verify wolfBoot-format images in addition to opaque blobs. wolfBoot images carry their signature (and optionally a certificate chain) in a TLV-formatted manifest header, so the manager now knows how to parse the header and source verification material from it rather than from fixed NVM IDs.

### What's new

- **`whServerImgMgrImgType` enum** distinguishes image loading behavior: `RAW` (existing — key from keystore, sig from NVM), `WOLFBOOT` (key from keystore, sig from header), and `WOLFBOOT_CERT` (root CA from NVM, cert chain + sig from header).
- **`whServerImgMgrImg`** gains `hdrAddr`/`hdrSize`/`imgType` fields; `sigNvmId` is repurposed as the root CA NVM ID in the cert-chain case.
- **Two new verify methods** for RSA4096+SHA256:
  - `wh_Server_ImgMgrVerifyMethodWolfBootRsa4096WithSha256` — standard wolfBoot verification against a keystore-provisioned public key.
  - `wh_Server_ImgMgrVerifyMethodWolfBootCertChainRsa4096WithSha256` — verifies the embedded cert chain against an HSM-provisioned root CA using the server cert API, then uses the leaf public key to verify the image signature.
- **Tests** covering positive and negative paths for both methods, using pre-generated signed payloads (`test/gen/wh_test_wolfboot_img_data.h`) produced offline by `test/scripts/gen_wolfboot_test_data.sh`.

### Future work

Only RSA4096+SHA256 is supported in this PR. Follow-ups will add additional wolfBoot-supported verification mechanisms (ECC, ed25519, etc.), reusing the same header-parsing and cert-chain plumbing.


test/wh_test_server_img_mgr.c
```diff
@@ -1192,6 +1196,240 @@ static int whTest_ServerImgMgrServerCfgRsa2048(whServerConfig* serverCfg)
...
+#ifndef NO_RSA
+static int
+whTest_ServerImgMgrServerCfgWolfBootRsa4096(whServerConfig* serverCfg)
+{
+    int                        ret          = 0;
+    whServerContext            server[1]    = {0};
+    whServerImgMgrConfig       imgMgrConfig = {0};
+    whServerImgMgrContext      imgMgr       = {0};
+    whServerImgMgrImg          testImage    = {0};
+    const whNvmId              testKeyId    = 1;
```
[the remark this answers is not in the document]
bigbrett: not a great way around this at the moment - the two are the same. Will be refactored in the future

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+        return WH_ERROR_NOTVERIFIED;
+    }
+
+    /* Read image size from header (this field is covered by the signature) */
+    memcpy(&img_size, hdr + 4, sizeof(img_size));
+
+    /* Ensure header-declared image size doesn't exceed supplied payload */
+    if ((size_t)img_size > payloadSize) {
+        return WH_ERROR_BADARGS;
+    }
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+
+    return WH_ERROR_OK;
+}
+
+/**
+ * Common wolfBoot header validation and hash computation.
+ * Validates that the header is large enough for the fixed fields, and that
+ * the header-declared image size does not exceed the supplied payload buffer.
+ * Returns WH_ERROR_OK on success and populates computed_hash.
+ * On success, sig_out and sig_sz_out point to the signature in the header.
```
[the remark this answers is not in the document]
bigbrett: this is the hash of a public key which is not secret. Not required.

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+        return ret;
+    }
+
+    /* Compare computed hash with stored hash */
+    if (memcmp(computed_hash, stored_sha, WC_SHA256_DIGEST_SIZE) != 0) {
+        return WH_ERROR_NOTVERIFIED;
+    }
+
+    /* Parse HDR_SIGNATURE */
+    sig_size = _wolfBootImgFindHeaderField(
```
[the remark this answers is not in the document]
bigbrett: there is no timing side channel here - it is not a secret dependent branch. It is just comparing the hash of the image with the computed hash.

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+
+    return WH_ERROR_OK;
+}
+
+
+/**
+ * Verify that a public key matches the hint stored in the wolfBoot header.
+ * The hint is SHA256(pubkey).
+ *
+ * NOTE: Code lifted directly from wolfBoot. Should remain as unmodified as
```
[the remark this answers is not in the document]
bigbrett: Both operands are attacker-observable: digest_out comes from RSA public-key decrypt of an attacker-submitted signature (attacker can compute it themselves), and hash is SHA256 of the      
  attacker-submitted image. No secret-dependent branch — the attacker already knows the expected comparison result. Constant-time comparison adds no security here.

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+    }
+
+    /* SEQUENCE - algoid */
+    if (_wolfBootImgDecodeAsn1Tag(input, inputSz, &idx, &algo_len,
+                                  _IMGMGR_ASN_SEQUENCE) != 0) {
+        return -1;
+    }
+    idx += algo_len; /* skip algoid */
+
+    /* OCTET STRING - digest */
```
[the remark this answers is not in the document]
bigbrett: this is how wolfBoot does it so just leaving at is. Not really an issue..

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+        htype = (uint16_t)((uint16_t)p[0] | ((uint16_t)p[1] << 8));
+        if (htype == 0) {
+            break;
+        }
+
+        /* Skip padding bytes and alignment */
+        if ((p[0] == WH_IMG_MGR_WOLFBOOT_HDR_PADDING) ||
+            ((((uintptr_t)p) & 0x01) != 0)) {
+            p++;
+            continue;
```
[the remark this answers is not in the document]
bigbrett: added even/odd check on header address

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+    hdr = (const uint8_t*)serverHdrPtr;
+
+    /* DMA pre-process payload */
+    ret = wh_Server_DmaProcessClientAddress(
+        server, img->addr, &serverPayloadPtr, img->size,
+        WH_DMA_OPER_CLIENT_READ_PRE, (whServerDmaFlags){0});
+    if (ret != WH_ERROR_OK) {
+        goto cleanup;
+    }
+    payload = (const uint8_t*)serverPayloadPtr;
```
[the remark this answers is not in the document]
bigbrett: fixed with new peek functionality. Good catch. Good bot

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,590 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_server_img_mgr.c
```diff
@@ -431,6 +459,604 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+    end_sha = stored_sha_ptr - tlv_sz;
+
+    ret = wc_InitSha256(&sha256_ctx);
+    if (ret != 0) {
+        return WH_ERROR_ABORTED;
+    }
+
+    /* Hash header portion */
+    ret = wc_Sha256Update(&sha256_ctx, hdr, (word32)(end_sha - hdr));
+    if (ret != 0) {
```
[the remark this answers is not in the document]
bigbrett: good point. Was going to say outside the scope of this PR but went ahead and just fixed

src/wh_server_img_mgr.c
```diff
@@ -431,6 +456,674 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+
+    *ptr = NULL;
+
+    if (p > max_p) {
+        return 0;
+    }
+
+    while ((p + 4) < max_p) {
+        htype = (uint16_t)((uint16_t)p[0] | ((uint16_t)p[1] << 8));
+        if (htype == 0) {
```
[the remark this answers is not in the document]
bigbrett: meh not rly possible but fixed

wolfhsm/wh_server_img_mgr.h
```diff
@@ -34,6 +34,30 @@
...
+#define WH_IMG_MGR_WOLFBOOT_HDR_SHA256 0x03
+#define WH_IMG_MGR_WOLFBOOT_HDR_IMG_TYPE 0x04
+#define WH_IMG_MGR_WOLFBOOT_HDR_PUBKEY 0x10
+#define WH_IMG_MGR_WOLFBOOT_HDR_SIGNATURE 0x20
+#define WH_IMG_MGR_WOLFBOOT_HDR_CERT_CHAIN 0x23
+#define WH_IMG_MGR_WOLFBOOT_HDR_PADDING 0xFF
+
+/* wolfBoot auth key types (from high byte of img_type) */
+#define WH_IMG_MGR_WOLFBOOT_AUTH_RSA4096 0x04
+#define WH_IMG_MGR_WOLFBOOT_HDR_IMG_TYPE_AUTH_MASK 0xFF00
```
billphipps, line 51: Consider using wolfboot header for this information.

wolfhsm/wh_server_img_mgr.h
```diff
@@ -86,10 +110,15 @@ typedef int (*whServerImgMgrVerifyAction)(whServerImgMgrContext* context,
...
 typedef struct whServerImgMgrImg {
-    uintptr_t                  addr;         /* Image address */
-    size_t                     size;         /* Image size */
-    whKeyId                    keyId;        /* Key ID for verification */
-    whNvmId                    sigNvmId;     /* NVM ID for signature */
+    uintptr_t addr;     /* RAW: image blob. WOLFBOOT: firmware payload */
+    size_t    size;     /* RAW: blob size. WOLFBOOT: max payload size, actual
+                            size obtained from manifest header */
+    uintptr_t hdrAddr;  /* wolfBoot header address (unused for RAW). Must be
+                            2-byte aligned for WOLFBOOT/WOLFBOOT_CERT. */
```
billphipps, line 117: Hmm.  Alignment requirement is from wolfBoot?  

Is this tested at runtime prior to usage or does wolfBoot test it?

test/gen/wh_test_wolfboot_img_data.h
```diff
@@ -0,0 +1,2016 @@
+/*
+ * Copyright (C) 2025 wolfSSL Inc.
```
billphipps, line 2: Nit. It's 2026.  ;)

test/scripts/gen_wolfboot_test_data.sh
```diff
@@ -0,0 +1,249 @@
+#!/bin/bash
```
billphipps, line 1: Add copyright?

test/scripts/gen_wolfboot_test_data.sh
```diff
@@ -0,0 +1,249 @@
...
+WOLFHSM_DIR="$(dirname "$TEST_DIR")"
+WOLFBOOT_DIR="${WOLFBOOT_DIR:-$(realpath "$WOLFHSM_DIR/../wolfBoot")}"
+
+KEYGEN="$WOLFBOOT_DIR/tools/keytools/keygen"
+SIGN="$WOLFBOOT_DIR/tools/keytools/sign"
+CHAIN_SCRIPT="$WOLFBOOT_DIR/tools/scripts/sim-gen-dummy-chain.sh"
+
+OUTPUT_DIR="$TEST_DIR/gen"
+WORK_DIR=$(mktemp -d)
+trap "rm -rf $WORK_DIR" EXIT
```
billphipps, line 32: Seems a bit extreme.  Consider testing/enabling overwrite to avoid stale data.

src/wh_server_img_mgr.c
```diff
@@ -106,33 +109,58 @@ int wh_Server_ImgMgrVerifyImg(whServerImgMgrContext*      context,
...
+            /* sig/sigSz passed as NULL/0 to callback */
+            break;
+
+        case WH_IMG_MGR_IMG_TYPE_WOLFBOOT_CERT:
+            /* Skip both key and sig loading - callback handles everything */
+            /* sig/sigSz and key/keySz passed as NULL/0 to callback */
+            break;
+
+        case WH_IMG_MGR_IMG_TYPE_RAW:
+        default:
```
billphipps: Consider returning BADARGS on default: instead of assuming RAW.

src/wh_server_img_mgr.c
```diff
@@ -431,6 +456,677 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
 
     return WH_ERROR_OK; /* RSA verification succeeded */
 }
+#endif /* !NO_RSA */
+
+#ifndef NO_RSA
```
billphipps, line 463: Is this the correct guard?

test/gen/wh_test_wolfboot_img_data.h
```diff
@@ -0,0 +1,2016 @@
...
+ */
+
+#ifndef WH_TEST_WOLFBOOT_IMG_DATA_H
+#define WH_TEST_WOLFBOOT_IMG_DATA_H
+
+#include <stdint.h>
+#include <stddef.h>
+
+/* wolfBoot signed header (RSA4096+SHA256, 1024 bytes) */
+static const uint8_t wolfboot_test_header[] = {
```
billphipps, line 35: There is no guarantee this will by 2-byte aligned.

src/wh_server_img_mgr.c
```diff
@@ -431,6 +456,677 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+
+/**
+ * Find a TLV field in a wolfBoot image header.
+ * Scans from hdr + WH_IMG_MGR_WOLFBOOT_HDR_OFFSET, reads type (uint16 LE)
+ * + len (uint16 LE), skips padding (0xFF), returns pointer to value and length.
```
billphipps: Returns the L of the TLV, or 0 if not found or invalid input.

src/wh_server_img_mgr.c
```diff
@@ -431,6 +456,677 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+                                      devId);
+    if (ret != WH_ERROR_OK) {
+        goto cleanup;
+    }
+
+    /* Verify pubkey hint matches provided key */
+    pubkey_hint_size = _wolfBootImgFindHeaderField(
+        hdr, img->hdrSize, WH_IMG_MGR_WOLFBOOT_HDR_PUBKEY, &pubkey_hint);
+    ret = _wolfBootImgVerifyPubKeyHint(key, (uint32_t)keySz, pubkey_hint,
+                                       pubkey_hint_size, devId);
```
billphipps, line 964: Check pubkey_hint_size != 0  before?

src/wh_server_img_mgr.c
```diff
@@ -431,6 +456,677 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+    if (ret != WH_ERROR_OK) {
+        goto cleanup;
+    }
+
+    /* Verify RSA4096 signature */
+    ret = _wolfBootImgVerifySigRsa4096(hdr_sig, hdr_sig_sz, computed_hash,
+                                       WC_SHA256_DIGEST_SIZE, key,
+                                       (uint32_t)keySz, devId);
+
+cleanup:
```
billphipps, line 974: Insert goto grumble here.

test/wh_test_server_img_mgr.c
```diff
@@ -1192,6 +1196,240 @@ static int whTest_ServerImgMgrServerCfgRsa2048(whServerConfig* serverCfg)
...
+    /* Initialize the image manager */
+    ret = wh_Server_ImgMgrInit(&imgMgr, &imgMgrConfig);
+    if (ret != WH_ERROR_OK) {
+        WH_ERROR_PRINT("Failed to initialize image manager: %d\n", ret);
+        wh_Server_Cleanup(server);
+        return ret;
+    }
+
+    /* Cache the public key in the keystore */
+    {
```
billphipps, line 1242: Excellent use of local scope!  Wish we did things like this more often.

consider making this like:
do { 
  if (ret=allocstuff()) {
    dostuff;
    freestuff();
  } else break;
  domorestuff;
} while 0; 
return ret;
so you can use break to exit early on error without doing a return.

test/wh_test_server_img_mgr.c
```diff
@@ -1192,6 +1196,240 @@ static int whTest_ServerImgMgrServerCfgRsa2048(whServerConfig* serverCfg)
...
+    if (ret != WH_ERROR_OK) {
+        WH_ERROR_PRINT("wolfBoot verify failed: %d\n", ret);
+        wh_Server_Cleanup(server);
+        return ret;
+    }
+    if (result.verifyMethodResult != WH_ERROR_OK) {
+        WH_ERROR_PRINT("wolfBoot verify method failed: %d\n",
+                       result.verifyMethodResult);
+        wh_Server_Cleanup(server);
+        return result.verifyMethodResult;
```
billphipps, line 1277: Is this a WH_ERROR_* enum?

billphipps (dismissed):
Love it!  Minor nits and questions.  Consider isolating all of the "lifted" wolfBoot code into a second C file.  Maybe we can get it modularized in wolfBoot to include it directly.

wolfhsm/wh_server_img_mgr.h
bigbrett: So I started out doing it this way but ultimately ended up explicitly NOT wanting wolfBoot as a dependency. It ended up getting really hairy since we have bidirectional dependencies, since wolfBoot pulls wolfHSM in as a dependency sometimes. Also the image format of wolfBoot at this point is set in stone, it cant be change without breaking many customers images. So safe to rely on it here.

src/wh_server_img_mgr.c
bigbrett: for now it is, due to -Wall -Werror, resulting in unused function warnings. This will be modified once another algo is brought online.

src/wh_server_img_mgr.c
bigbrett: at this point, many portions of wolfHSM code have become unreadable due to all the deferred and conditional wolfCrypt cleanups. There are only so many nested if (ret==0) one can tolerate.

Recommend we ban `goto` EXCEPT for single point cleanup at the end of the functions, which are easily autitable, and easy able to be justified for MISRA exceptions. 

Single label forward jumps for cleanup and deallocation are standard practice and should be encouraged. Any other use should be prohibited.

Just my 0.02

test/wh_test_server_img_mgr.c
bigbrett: depends where it fails. wolfCrypt errors should be bubbled up too, so didn't want to smoosh them as it is  a potentially valuable debugging tool when determining why test failed. "Failed with WH_ERROR_ABORTED" is less helpful than "Failed with <instert algo specific error code here>"

test/scripts/gen_wolfboot_test_data.sh
bigbrett: IIRC company policy is we only need for source code, not tooling. can add tho

src/wh_server_img_mgr.c
AlexLanzano: Second this. Do the case for raw then after have default: BADARGS

src/wh_server_img_mgr.c
AlexLanzano: Agree with this. Single point goto cleanups are much easier to read and manage than doing it explicitly in the if or doing the if (ret == OK) method

AlexLanzano (dismissed):

src/wh_server_img_mgr.c
bigbrett: fixed

wolfhsm/wh_server_img_mgr.h
bigbrett: yeah based on the parser logic. And it is not tested, but unaligned headers will fail verification. Just a comment explaining why. Don't want the code to diverge too far from upstream though

src/wh_server_img_mgr.c
bigbrett: updated comment

src/wh_server_img_mgr.c
bigbrett: verified in `_wolfBootImgVerifyPubKeyHint()`

AlexLanzano (approved):

src/wh_server_img_mgr.c
```diff
@@ -431,6 +458,678 @@ int wh_Server_ImgMgrVerifyMethodRsaSslWithSha256(
...
+/**
+ * Find a TLV field in a wolfBoot image header.
+ * Scans from hdr + WH_IMG_MGR_WOLFBOOT_HDR_OFFSET, reads type (uint16 LE)
+ * + len (uint16 LE), skips padding (0xFF), and returns the length field of the
+ * TLV, or 0 if not found or invalid input.
+ *
+ * NOTE: Code lifted directly from wolfBoot. Should remain as unmodified as
+ * possible for easy diffs
+ */
+static uint16_t _wolfBootImgFindHeaderField(const uint8_t* hdr, size_t hdrSize,
```
dgarske, line 474: Why was this code re-implemented instead of using wolfBoot as a library? Why was @danielinux or myself not consulted on this work?

bigbrett: @dgarske because I explicitly did not want to pull wolfBoot in as a dependency to wolfHSM, when wolfHSM is already pulled in as a dependency of wolfBoot AND the private ports will often pull in two disjoint versions of both wolfHSM and wolfBoot outside the scope of both libraries. It becomes dependency hell. Because the core wolfBoot image format is unlikely to change (it would break literally every customers build) this was determined to be a satisfactory compromise. I believe our tests should ensure compat and flag failure - it might not run against master but I could easily implement that.

bigbrett: as for why you were not consulted, it was mostly because it was a change internal to wolfHSM and not a change to wolfBoot, so didn't feel the need to bother you with it. I'd be happy to involve you on anything that remotely involves wolfBoot going internal to wolfHSM going forward if that is desired?

bigbrett: Also the wolfBoot as a library code's wasn't as portable as we required. Specifically it would require wolfHSM to have a working configuration of wolfBoot with the same feature enablement as the image it wants to verify (which might be a completely different core). This is hard if you dont know the algos you need to use. Dont recall all the specifics, it was a while ago, but I did try initially and it ended up being a PITA so decided to just re implement the relatively trivial image header parsing locally

PR #343, opened 2026-04-22
Zero stack buffers in keywrap and datawrap handlers

padelsbach:
Ensure sensitive buffers are zeroed when returning from `_HandleKeyWrapRequest`, `_HandleKeyUnwrapAndExportRequest`, and `_HandleDataWrapRequest`

Fixes F-2297 and F-2303

### DIFF
--- a/src/wh_server_keystore.c
+++ b/src/wh_server_keystore.c
@@ -1202,7 +1202,7 @@ static int _HandleKeyWrapRequest(whServerContext*                  server,
                                  uint8_t* respData, uint32_t respDataSz)
 {
 
-    int           ret;
+    int           ret = WH_ERROR_OK;
     uint8_t*      wrappedKey;
     whNvmMetadata metadata;
     uint8_t       key[WOLFHSM_CFG_KEYWRAP_MAX_KEY_SIZE];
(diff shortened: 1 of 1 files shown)


src/wh_server_keystore.c
```diff
@@ -1555,31 +1554,27 @@ static int _HandleDataWrapRequest(whServerContext*                   server,
...
 #endif /* HAVE_AESGCM */
 #endif /* !NO_AES */
 
         default:
-            return WH_ERROR_BADARGS;
+            ret = WH_ERROR_BADARGS;
     }
 
-    return WH_ERROR_OK;
+    wc_ForceZero(data, sizeof(data));
```
bigbrett, line 1576: we should be using `wh_Utils_ForceZero()` rather than the wolfCrypt API directly.

bigbrett (changes requested):
One nit to change the force zero API to use the wolfHSM native one.

That said, I'm somewhat unconvinced on this PR - I think the fenrir findings are false positive and aren't really necessary because this is server code and by definition we are running in a secure environment with only trusted access to RAM. So unless there was a buffer overread that somehow was able to exfiltrate that data back to the client, this fixes an essentially unexploitable path. Otherwise, we would be force zeroing pretty much any time any secret or any plaintext is present. I suppose I can leave it in as it is relatively innocuous, however I'm tempted not to, since we don't apply this scheme in a meaningful manner across the codebase. I'd almost not rather have force zero applied in an ad-hoc manner until we come up with a coherent policy of what must be zeroed and what must not.

TLDR Unless there is a real exfiltration path, it seems like noise. Until we produce a coherent security policy that can be uniformly enforced across the entire code base, I'm tempted to skip these findings (again, unless there is a REAL exfiltration path)

padelsbach:
Seemed like a nice defensive enhancement if we are not sensitive to cycle count, but I see your point. Let's revisit if/when we have a policy in place. Closing.

PR #348, opened 2026-05-01
Add stm32H5 TrustZone wolfHSM Port

aidangarske:
# Description                                                                                                                                         
  - Adds `port/stmicro/stm32-tz/wh_transport_nsc.{c,h}`: a portable
    synchronous TrustZone non-secure-callable bridge transport for                                                                                         
    ARMv8-M Cortex-M targets. Client `Send` invokes a host-supplied
    veneer (`wcs_wolfhsm_transmit`) inline and caches the response;                                                                                        
    client `Recv` consumes the cached response on the first call.                                                                                          
    Server-side callbacks consume the request the host's veneer parked                                                                                     
    in a static context and write the response back into the non-secure                                                                                    
    caller's buffer.
  - Target-agnostic. The STM32H5-specific glue (NSC veneer, flash                                                                                          
    adapter, secure-side server init, NS test exerciser) lives in the                                                                                      
    matching wolfBoot PR.
  - New `STM32_TZ_NSC=1` build flag in `test/Makefile` compiles the                                                                                        
    transport into the host test build and pulls in a new unit test                                                                                        
    `test/wh_test_transport_nsc.c` covering BADARGS, NOTREADY, happy-
    path round trip, and the `request_pending` / `rsp_size` state                                                                                          
    machine for both callback tables.
  - New CI lane in `.github/workflows/build-and-test.yml`:                                                                                                 
    `STM32_TZ_NSC=1 ASAN=1` build + run.                                                                                                                   
  - Docs: new "STM32 TrustZone (STM32H5 / NSC bridge)" section in
    `docs/src/chapter08.md`.                                                                                                                               
                  
  ## Notes                                                                                                                                                 
  - Companion wolfBoot PR adds `WOLFCRYPT_TZ_WOLFHSM=1` for STM32H5,
    which is the first consumer of this transport. [here](https://github.com/wolfSSL/wolfBoot/pull/769)                                                                                                  
   
  ## Test plan                                                                                                                                             
  - [x] `STM32_TZ_NSC=1 ASAN=1` build + `make run` (CI)
  - [x] Existing CI matrix unaffected (no changes outside the new port)                                                                                    
  - [x] End-to-end on STM32H5 NUCLEO-H563ZI via the wolfBoot PR
  - [x] Nightly test added tests both latest masters also report problem in issues like https://github.com/wolfSSL/wolfHSM/issues/396 if fails


bigbrett (changes requested):
@aidangarske thanks for this! I'm little unclear whether any of this stm32H5 specific or is just a generic TZ-M compatible transport? In the docs you say 

> The transport itself is target-agnostic; the STM32H5-specific glue (NSC veneer, `whFlashCb` flash adapter, secure-side server init, NS test exerciser) lives in the wolfBoot port.

If this is the case then I don't think you need any mention of STM32H5 in any of this? Unless I'm misunderstanding. If this is meant to be a generic trustzone M transport then I think having under `port/tzm` or something like that is better, with no mentions of STM32H5 at all. 

LMK if I'm misunderstanding

bigbrett:
@aidangarske reverting to draft after our conversation yesterday

.github/workflows/wolfboot-tz-integration.yml
```diff
@@ -0,0 +1,96 @@
+# Cross-repo CI: build wolfBoot main against this wolfHSM PR and run
+# the STM32H5 TZ demo under m33mu. Informational until m33mu's
+# data-dependent ECDH issue is fixed upstream; NUCLEO-H563ZI hardware
+# is the authoritative validation (log in PR description).
```
bigbrett: I assume the data-dependent ECDH issue is fixed upstream? Either way, pls refine this comment to be less ai-ish

.github/workflows/wolfboot-tz-integration.yml
```diff
@@ -0,0 +1,96 @@
...
+on:
+  pull_request:
+    paths:
+      - 'port/armv8m-tz/**'
+      - 'wolfhsm/wh_transport*.h'
+      - 'wolfhsm/wh_client*.h'
+      - 'src/wh_client*.c'
+      - 'src/wh_server*.c'
+      - 'test/wh_test.c'
+      - '.github/workflows/wolfboot-tz-integration.yml'
```
bigbrett: at this point I'd either remove the paths and have it run on every PR or only scope it to port/armv8m-tz

port/stmicro/stm32h5-tz/README.md
```diff
@@ -0,0 +1,27 @@
+# STM32H5 TrustZone port — lives in wolfBoot
+
+This directory is a pointer, not an implementation.
```
bigbrett: pls no

src/wh_server.c
```diff
@@ -233,6 +233,11 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
         /* Process the init action */
         server->comm->client_id = req.client_id;
 
+        /* A fresh INIT after a prior CLOSE must put the channel back into
+         * the connected state so the subsequent request loop accepts non-
+         * COMM groups again. */
+        wh_Server_SetConnected(server, WH_COMM_CONNECTED);
```
bigbrett: This is not correct and should not be set here. Exactly where this is set is on a per-transport basis. In the SHM transport this must be set via an out-of-band signal after the client has zeroed out the shared memory.

src/wh_server.c
```diff
@@ -526,9 +531,8 @@ int wh_Server_HandleRequestMessage(whServerContext* server)
     /* Use the CommServer internal buffer to avoid copies */
     data = wh_CommServer_GetDataPtr(server->comm);
 
-    /* Are we connected with a valid data pointer? */
-    if (    (server->connected == WH_COMM_DISCONNECTED) ||
-            (data == NULL) ) {
+    /* Valid data pointer? */
+    if (data == NULL) {
```
bigbrett: why did we remove the connected guard?

src/wh_server.c
```diff
@@ -539,6 +543,15 @@ int wh_Server_HandleRequestMessage(whServerContext* server)
         group = WH_MESSAGE_GROUP(kind);
         action = WH_MESSAGE_ACTION(kind);
 
+        /* When disconnected, only allow COMM-group messages through so the
+         * client can re-establish (INIT), interrogate (INFO/ECHO) or repeat
+         * a CLOSE on the channel. Any other request is rejected with
+         * NOTREADY until a fresh INIT brings the channel back up. */
```
bigbrett: unfortunately this is not the flow. Connected is a bit of a misnomer. It is more a signal that the transport is ready for use.

src/wh_server_keystore.c
```diff
@@ -317,6 +317,25 @@ static int _GetKeyCacheSlot(whKeyCacheContext* ctx, uint16_t keySz,
             }
         }
 
+        /* Last-resort fallback: every slot is occupied by an uncommitted
+         * key (typical when many ephemeral keys are cached in succession
+         * with no intervening commit). Reclaim the first non-erased slot
+         * so the cache cannot saturate on transient state alone. Drops
+         * the slot's in-flight contents; the caller's request gets the
+         * slot fresh. */
```
bigbrett: I don't think this is correct behavior. It is perfectly fine to fail if the client is trying to cache more keys than the server can handle without committing. It should be up to the client to evict manually in this case. At least that is how it was intended to work. Same issue below with the big cache.

test/wh_test_check_struct_padding.c
```diff
@@ -204,4 +204,10 @@ whMessageCert_VerifyAcertRequest whMessageCert_VerifyAcertRequest_test;
 #endif /* WOLFHSM_CFG_CERTIFICATE_MANAGER_ACERT */
 #endif /* WOLFHSM_CFG_CERTIFICATE_MANAGER */
 
+#ifdef WOLFHSM_CFG_PORT_ARMV8M_TZ_NSC
+#include "wh_transport_nsc.h"
+whTransportNscClientContext whTransportNscClientContext_test;
+whTransportNscServerContext whTransportNscServerContext_test;
+#endif /* WOLFHSM_CFG_PORT_ARMV8M_TZ_NSC */
```
bigbrett: why did we add these to the padding check? This should only hold "wire structures"

wolfhsm/wh_settings.h
```diff
@@ -129,6 +129,14 @@
  *  WOLFHSM_CFG_PORT_GETTIME - Function-like macro returning the current system
  *      time in microseconds as a uint64_t. Must be defined in wolfhsm_cfg.h for
  *      the active port UNLESS WOLFHSM_CFG_NO_SYS_TIME is defined
+ *
+ *  WOLFHSM_CFG_PORT_ARMV8M_TZ_NSC - If defined, build the ARMv8-M TrustZone
```
bigbrett: I dont think we want a separate compilation guard for port-specific stuff. Core library features, yes, but that is because the library code should support being wildcard compiled. Port stuff needs to be explicitly opted in and compiled manually so there is no sense in requiring a new core library config macro for it

docs/src/chapter08.md
```diff
@@ -46,6 +46,16 @@ The distribution of this  port is restricted by the vendor. Please contact suppo
...
+
+The port provides:
+- Single-call NSC transport (no polling, no shared-memory ring): client `Send` invokes the host-supplied veneer inline and caches the response; client `Recv` consumes the cached response on the first call (subsequent calls return `WH_ERROR_NOTREADY` until the next `Send`).
+- Server-side callbacks that consume the request the host's veneer parked in a static context and write the response back to the non-secure caller's buffer.
+
+The transport is target-agnostic. Bringing it up on a new ARMv8-M part is a porting exercise on the host side only: provide the `cmse_nonsecure_entry` veneer that fronts `wcs_wolfhsm_transmit`, plus whatever flash/NVM adapter and server init the deployment needs.
```
bigbrett: This line doesnt make sense. "Host side only"? Is that meant to indicate non-secure world only? But flash/NVM should live in secure world. This doc could use a human touch...

bigbrett (changes requested):

.github/workflows/wolfboot-tz-integration.yml
aidangarske: Reworded it, dropped the AI-ish bit. The ECDH thing was an m33mu emulator quirk, not wolfHSM

.github/workflows/wolfboot-tz-integration.yml
aidangarske: Scoped it to just port/armv8m-tz/**

port/stmicro/stm32h5-tz/README.md
aidangarske: dropped wording

src/wh_server.c
aidangarske: you're right. Reverted it. Moved the reconnect into the NSC veneer instead, it re-asserts connected before each dispatch since the bridge is always ready once linked.

src/wh_server.c
aidangarske: Reverted, guard's back.

src/wh_server.c
aidangarske: Got it, reverted. The veneer now signals transport-ready per call, which matches how you described it.

src/wh_server_keystore.c
aidangarske: Test just needed the cache sized to its working set the wolfBoot port sets 9 + 3 now to match the test config

test/wh_test_check_struct_padding.c
aidangarske: reverted

wolfhsm/wh_settings.h
aidangarske: removed

docs/src/chapter08.md
aidangarske: reworked wording should be better now

port/armv8m-tz/README.md
```diff
@@ -0,0 +1,44 @@
+# ARMv8-M TrustZone NSC bridge transport
+
+A synchronous wolfHSM transport for ARMv8-M TrustZone targets that
+bridges a non-secure client to a secure-world server through a single
+Non-Secure Callable (NSC) veneer. See `wh_transport_nsc.h` for the C
+API and `docs/src/chapter08.md` for the narrative description.
```
bigbrett: stale doc reference. But this README has way more info than the referenced doc page anyway, so I'd skip the cross reference tbh

port/stmicro/stm32h5-tz/README.md
```diff
@@ -0,0 +1,11 @@
+# STM32H5 TrustZone port
+
+The STM32H5 wolfHSM TrustZone port (HAL adapter, secure-side server,
+flash/NVM backend, non-secure demo app, linker layout, and test
+harness) is maintained in
+[wolfBoot](https://github.com/wolfSSL/wolfBoot) under
+`port/stmicro/stm32h5-tz-wolfhsm/` and `test-app/wcs/`.
+
+It consumes the generic ARMv8-M TrustZone NSC bridge transport in
+`port/armv8m-tz/` of this repository. See `docs/src/chapter08.md` for
```
bigbrett: stale doc reference, perhaps refer to transport readme instead.

port/armv8m-tz/wh_transport_nsc.h
```diff
@@ -0,0 +1,91 @@
...
+    uint8_t        WH_PAD[1];
+} whTransportNscServerContext;
+
+typedef struct {
+    uint8_t WH_PAD[1];
+} whTransportNscServerConfig;
+
+/* Pre-populated tables; callbacks are file-local in wh_transport_nsc.c */
+extern const whTransportClientCb whTransportNscClient_Cb;
+extern const whTransportServerCb whTransportNscServer_Cb;
```
bigbrett: Is there a reason you made these const extern and statically allocated by the transport source file? 

Were you trying to make the transport essentially a singleton style thing, where only one instance is ever allowed on a system? If that was the case, I would think the context itself is what would need the singleton global scope, not necessarily the callback table.

Otherwise, you could just provide a "helper macro" struct initializer and bind it to the context via the config struct. That is how we do it elsewhere in wolfHSM. See how the [POSIX SHM transport does it](https://github.com/wolfSSL/wolfHSM/blob/9d141e9427eaffa8939b07203ef0f37d7e45303c/port/posix/posix_transport_shm.h#L150-L164), for example. Lots of other examples as well - pretty much anything with configurable back-end callbacks (all the transports, comm layer, NVM, etc).

Not a huge deal, just curious why the pattern wasn't followed here.

bigbrett (changes requested):
Nice! A few nits and a question that might require a change, depending on the answer

port/armv8m-tz/wh_transport_nsc.h
aidangarske: No i wasnt trying to do that. I think I just missed the convention you had in place. Changed to convention

bigbrett (approved):

PR #350, opened 2026-05-04
multi-root CA certificate verification 

bigbrett:
Adds multi-root certificate verification support to wolfHSM so certificate chains can be validated against a set of trusted root NVM IDs instead of just a single anchor. It extends the cert message protocol, client/server certificate APIs, and test coverage for the new verification mode.

Changes:
- Added wh_Server_CertVerifyMultiRoot plus single-root delegation through the new multi-root server implementation.
- Added client-side multi-root verify APIs and corresponding server/client/unit tests.
- DMA support

### DIFF
--- a/src/wh_client_cert.c
+++ b/src/wh_client_cert.c
@@ -507,6 +507,168 @@ int wh_Client_CertVerifyAndCacheLeafPubKey(
                        inout_keyId, out_rc);
 }
 
+/* Helper: send a multi-root verify request */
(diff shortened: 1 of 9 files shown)


AlexLanzano (dismissed):
Looks great! Just need to fix the test case in test/wh_test_cert.c:172
You're only passing one root where the comments says you're passing in two.

Github wouldnt let me make a comment directly on the source file btw...

AlexLanzano (approved):

PR #351, opened 2026-05-04
Async AES

bigbrett:
Adds async API support for AES

### DIFF
--- a/docs/draft/async-crypto.md
+++ b/docs/draft/async-crypto.md
@@ -541,10 +541,13 @@ int wh_Client_RngGenerateDma(whClientContext* ctx, uint8_t* out, uint32_t size);
 ## AES: One-Shot with DMA Support
 
 AES modes (CBC, CTR, ECB, GCM) are all **one-shot** operations — every call
-consumes a fixed buffer of input and returns a fixed buffer of output. There
-is no streaming state to accumulate across multiple Request/Response pairs
-the way SHA does, which makes the async split significantly simpler than
-SHA:
+consumes a fixed buffer of input and returns a fixed buffer of output in a
+single round-trip. There is no client-side partial-block accumulation across
+Request/Response pairs the way SHA's `Update` family has, which makes the
+async split significantly simpler than SHA. (CBC and CTR still carry
+inter-call IV / counter state on the `Aes` struct — see *Mutable state*
+below — but that state is updated atomically by the Response, not buffered
+across calls.)
 
 - **No partial-block buffering** on the client.  The entire plaintext or
   ciphertext is handed to one Request and the full result comes back in
--- a/docs/draft/async-crypto.md
+++ b/docs/draft/async-crypto.md
@@ -416,8 +416,6 @@ int wh_Client_Sha256Dma(whClientContext* ctx, wc_Sha256* sha, const uint8_t* in,
 | **`requestSent` flag** | Adds a parameter to the API, but avoids unnecessary round-trips when input is absorbed entirely into the local buffer |
 | **Snapshot/rollback on send failure** | Small CPU cost to copy the partial buffer, but guarantees SHA state consistency even on transport failures |
 
-
-
 ## RNG: Single-Shot with Caller-Driven Chunking
 
 The RNG generate operation is the second algorithm to receive the async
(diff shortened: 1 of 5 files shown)


docs/draft/async-crypto.md
```diff
@@ -540,6 +538,85 @@ int wh_Client_RngGenerate(whClientContext* ctx, uint8_t* out, uint32_t size);
 int wh_Client_RngGenerateDma(whClientContext* ctx, uint8_t* out, uint32_t size);
 ```
 
+## AES: One-Shot with DMA Support
+
+AES modes (CBC, CTR, ECB, GCM) are all **one-shot** operations — every call
+consumes a fixed buffer of input and returns a fixed buffer of output. There
+is no streaming state to accumulate across multiple Request/Response pairs
+the way SHA does, which makes the async split significantly simpler than
+SHA:
```
Frauschi: Well, technically this "no streaming state" is wrong for CBC and CTR, as you already explain below. A user can chain multiple CBC / CTR calls without problem. Only a nit, but the comment could be relaxed here.

src/wh_client_crypto.c
```diff
@@ -411,518 +411,664 @@ int wh_Client_RngGenerateDma(whClientContext* ctx, uint8_t* out, uint32_t size)
...
     req = (whMessageCrypto_AesCtrRequest*)_createCryptoRequest(
         dataPtr, WC_CIPHER_AES_CTR, ctx->cryptoAffinity);
-    uint8_t* req_in  = (uint8_t*)(req + 1);
-    uint8_t* req_key = req_in + len;
-    uint8_t* req_iv  = req_key + key_len;
-    uint8_t* req_tmp = req_iv + iv_len;
-    uint32_t req_len = sizeof(whMessageCrypto_GenericRequestHeader) +
-                       sizeof(*req) + len + key_len + iv_len +
-                       AES_BLOCK_SIZE;
-    WH_DEBUG_CLIENT_VERBOSE("enc:%d keylen:%d ivsz:%d insz:%d reqsz:%u "
```
Frauschi, line 453: Is it on purpose that we lose like all the debug logs?

src/wh_client_crypto.c
```diff
@@ -1197,339 +1440,270 @@ int wh_Client_AesCbcResponse(whClientContext* ctx, Aes* aes, uint8_t* out,
...
+                if (res->sz > out_capacity) {
+                    return WH_ERROR_ABORTED;
+                }
+                memcpy(out, res_out, res->sz);
+            }
             if (out_size != NULL) {
                 *out_size = res->sz;
             }
+            if (enc_tag != NULL && res->authTagSz > 0 &&
+                res->authTagSz <= tag_len) {
```
Frauschi: When the provided `tag_len` is too small for the actual tag, then we do not copy any tag data, but still return `WH_ERROR_OK`. Hence, the user incorrectly thinks that the operation succeeded and the tag is populated with valid data.

src/wh_client_crypto.c
```diff
@@ -1197,339 +1440,270 @@ int wh_Client_AesCbcResponse(whClientContext* ctx, Aes* aes, uint8_t* out,
...
-    /* Send request and receive response */
-    ret = wh_Client_SendRequest(ctx, group, action, req_len, dataPtr);
+    ret = wh_Client_AesGcmRequest(ctx, aes, enc, in, len, iv, iv_len, authin,
+                                  authin_len, dec_tag, tag_len);
     if (ret == WH_ERROR_OK) {
-        uint16_t res_len = 0;
         do {
-            ret =
-                wh_Client_RecvResponse(ctx, &group, &action, &res_len, dataPtr);
+            ret = wh_Client_AesGcmResponse(ctx, aes, out, len, NULL, enc_tag,
```
Frauschi, line 1525: Skoll: The previous monolithic `wh_Client_AesGcm` tolerated `out=NULL` with `len>0` (it simply skipped the output memcpy). The new blocking wrapper forwards `out, len` straight into `wh_Client_AesGcmResponse`, where `(out == NULL && out_capacity > 0)` is rejected with `WH_ERROR_BADARGS` *before* `wh_Client_RecvResponse` is called. Because the request was already sent, a stale response is left in the comm queue and subsequent calls will see `wh_CommClient_IsRequestPending()` block them with `WH_ERROR_REQUEST_PENDING`. No in-tree caller hits this combination today, but it is a silent regression of the public blocking API.

test/wh_test_crypto.c
```diff
@@ -6631,6 +6631,1592 @@ static int whTestCrypto_Aes(whClientContext* ctx, int devId, WC_RNG* rng)
...
+            }
+        }
+        (void)wc_AesFree(aes);
+    }
+    if (ret == 0) {
+        WH_TEST_PRINT("AES CBC ASYNC DEVID=0x%X SUCCESS\n", devId);
+    }
+#endif /* HAVE_AES_CBC */
+
+#ifdef WOLFSSL_AES_COUNTER
```
Frauschi, line 6736: The fail-fast logic is not tested for CTR. Hence, regressions about preserving `aes->reg` and `aes->tmp` are not identified. For CBC, a test for this is present.

src/wh_client_crypto.c
```diff
@@ -411,518 +411,664 @@ int wh_Client_RngGenerateDma(whClientContext* ctx, uint8_t* out, uint32_t size)
...
+    }
+
+    ret = wh_Client_RecvResponse(ctx, NULL, NULL, &res_len, dataPtr);
+    if (ret == WH_ERROR_NOTREADY) {
+        return ret;
+    }
+
+    if (ret == WH_ERROR_OK) {
+        ret = _getCryptoResponse(dataPtr, WC_CIPHER_AES_ECB, (uint8_t**)&res);
+        if (ret >= 0) {
```
Frauschi: `WH_ERROR_OK`?

Frauschi (changes requested):
Two bigger issues (GCM tag buffer size and GCM stale response) and some smaller nits. But mostly LGTM.
I really like how this new request/response model simplifies the individual functions.

docs/draft/async-crypto.md
bigbrett: yeah that was intended to reference the "partial block accumulation" that needs to be held. Reworded. This is internal documentation anyway.

src/wh_client_crypto.c
bigbrett: I didn't like how they looked but it shoudl probably remain. Added back in.

src/wh_client_crypto.c
bigbrett: so looks like this was a carryover from some of the other crypto APIs where the server can propagate back a wolfcrypt error code. In general, when the error code could be wolfCrypt it is left as 0. If it is a native wolfHSM error code only it is WH_ERROR_OK. Here it is nebulous and could be mixed. Changed to `WH_ERROR_OK` just for cleanliness but I guess this is something we should sort in the future.

test/wh_test_crypto.c
bigbrett: added test

src/wh_client_crypto.c
bigbrett: fixed

src/wh_client_crypto.c
bigbrett: fixed

Frauschi (approved):
Thanks for the fixes, LGTM now.

PR #352, opened 2026-05-04
Add support for LMS and XMSS

padelsbach:
Requires https://github.com/wolfSSL/wolfssl/pull/10380 to be merged first (done).

Adds support for "stateful" PQC using crypto callbacks added to wolfssl.


padelsbach:
https://github.com/wolfSSL/wolfssl/pull/10488 is required for CI to pass

src/wh_server_crypto.c
```diff
@@ -5734,101 +5989,1061 @@ static int _HandlePqcSigAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+        if (ret != 0) {
+            res.dmaAddrStatus.badAddr = req.pub;
         }
+    }
+    if (ret == 0) {
+        memcpy(clientPubAddr, key->pub, pubLen32);
+        res.keyId   = wh_KeyId_TranslateToClient(keyId);
+        res.pubSize = pubLen32;
+    }
+    (void)wh_Server_DmaProcessClientAddress(
```
padelsbach: done

src/wh_server_crypto.c
```diff
@@ -5734,101 +5989,1061 @@ static int _HandlePqcSigAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+         * (a) leaves the old state in NVM with no signature exposed, or
+         * (b) commits the new state with the signature lost in transit -
+         * one wasted index but never an index reused with a fresh sig. */
+        ret = wc_LmsKey_Sign(key, sigAddr, &sigLen, msgAddr, (int)req.msg.sz);
+        if (ret == 0) {
+            res.sigLen = sigLen;
+        }
+    }
+
+    (void)wh_Server_DmaProcessClientAddress(
```
padelsbach: done

src/wh_server_crypto.c
```diff
@@ -5734,101 +5989,1061 @@ static int _HandlePqcSigAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+    word32                                           privLen32 = 0;
+    whKeyId                                          keyId;
+    whMessageCrypto_PqcStatefulSigKeyGenDmaRequest   req;
+    whMessageCrypto_PqcStatefulSigKeyGenDmaResponse  res;
+    _XmssSkCapture                                   sk_cap;
+    /* WC_XMSS_MAX_SK comes from the params table; sized for the largest
+     * supported XMSS variant. The variants enabled in user_settings.h all
+     * fit in 4 KiB, but use the wolfCrypt-reported priv length to be
+     * exact. */
+    byte                                             sk_buf[4096];
```
padelsbach: done

src/wh_server_crypto.c
```diff
@@ -5734,97 +5989,1067 @@ static int _HandlePqcSigAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+        }
+    }
+    if (ret == WH_ERROR_OK) {
+        ret = wc_XmssKey_Sign(key, sigAddr, &sigLen, msgAddr, (int)req.msg.sz);
+        if (ret == 0) {
+            res.sigLen = sigLen;
+        }
+    }
+
+    (void)wh_Server_DmaProcessClientAddress(
```
padelsbach: done

src/wh_server_crypto.c
```diff
@@ -5734,97 +5989,1067 @@ static int _HandlePqcSigAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+    }
+    ret = wh_MessageCrypto_TranslatePqcStatefulSigKeyGenDmaRequest(
+        magic, (whMessageCrypto_PqcStatefulSigKeyGenDmaRequest*)cryptoDataIn,
+        &req);
+    if (ret != WH_ERROR_OK) {
+        return ret;
+    }
+
+    /* xmssParamStr arrives via the request struct. The client
+     * (wh_Client_XmssMakeKeyDma) currently has a TODO for populating it; for
```
padelsbach: done

bigbrett:
@Frauschi FYI merge conflicts are irrelevant to the review content so you can ignore for now

wolfhsm/wh_client_crypto.h
```diff
@@ -3117,5 +3117,76 @@ int wh_Client_MlKemDecapsulateDma(whClientContext* ctx, MlKemKey* key,
...
+int wh_Client_LmsSignDma(whClientContext* ctx, const byte* msg, word32 msgSz,
+                         byte* sig, word32* sigSz, LmsKey* key);
+
+/* Verify sig against msg using an HSM-resident LMS key. *res is set to 1 on
+ * success, 0 on signature mismatch. */
+int wh_Client_LmsVerifyDma(whClientContext* ctx, const byte* sig, word32 sigSz,
+                           const byte* msg, word32 msgSz, int* res,
+                           LmsKey* key);
+
+/* Query remaining signatures on an HSM-resident LMS key. */
```
Frauschi: Our internal wolfCrypt API does not support querying the actual number of remaining signatures, only whether there are any remaining signatures left. Hence, the number obtained here reflects this booleaness. This should be reflected in the comment.

src/wh_client_crypto.c
```diff
@@ -10253,4 +10259,770 @@ int wh_Client_MlKemDecapsulateDma(whClientContext* ctx, MlKemKey* key,
...
+                                        (uint8_t*)dataPtr);
+        }
+        if (ret == WH_ERROR_OK) {
+            do {
+                ret = wh_Client_RecvResponse(ctx, &group, &action, &req_len,
+                                             (uint8_t*)dataPtr);
+            } while (ret == WH_ERROR_NOTREADY);
+        }
+
+        (void)wh_Client_DmaProcessClientAddress(
```
Frauschi: I think discarding the return value of a `WRITE_POST` post-processing step is not correct and should actually be passed to the user (also see the recent #403).
Within a `WRITE_POST` post-process, there could be a `MEMCPY` from the DMA buffer to the actual user buffer (when the passed buffer from the user is not in the correct DMA memory region already). Hence, in case the post-processing fails, we cannot be 100% sure that the user buffer actually contains the right data (in contrast to READ post-processing, where we are only talking about cleaning-up work).
This goes for all four of the `WRITE_POST` calls in this file.

src/wh_client_crypto.c
```diff
@@ -10253,4 +10259,770 @@ int wh_Client_MlKemDecapsulateDma(whClientContext* ctx, MlKemKey* key,
...
+
+    req = (whMessageCrypto_PqcStatefulSigSigsLeftDmaRequest*)
+        _createCryptoRequestWithSubtype(
+            dataPtr, WC_PK_TYPE_PQC_STATEFUL_SIG_SIGS_LEFT,
+            WC_PQC_STATEFUL_SIG_TYPE_LMS, ctx->cryptoAffinity);
+
+    {
+        uint16_t group  = WH_MESSAGE_GROUP_CRYPTO_DMA;
+        uint16_t action = WC_ALGO_TYPE_PK;
+        uint16_t req_len =
```
Frauschi, line 10773: Nit: all other functions do a bounds-check `req_len > WOLFHSM_CFG_COMM_DATA_LEN` except for the LMS and XMSS `SigsLeft()` methods. Will probably never be a problem as the request is so small, but maybe adding the check for consistency might be beneficial (AI flaged this actually).

wolfhsm/wh_message_crypto.h
```diff
@@ -1539,6 +1539,123 @@ int wh_MessageCrypto_TranslateMlKemDecapsDmaResponse(
...
+    whMessageCrypto_DmaBuffer pub;        /* Server writes pub key here */
+    uint32_t                  flags;
+    uint32_t                  keyId;
+    uint32_t                  access;
+    uint32_t                  labelSize;
+    uint32_t                  lmsLevels;
+    uint32_t                  lmsHeight;
+    uint32_t                  lmsWinternitz;
+    uint8_t                   label[WH_NVM_LABEL_LEN];
+    char                      xmssParamStr[32];
```
Frauschi, line 1574: Nit: Maybe not worth the effort, but can we place the `xmssParamStr[]` array and the three LMS parameters (`lmsLevels`, `lmsHeight`, `lmsWinternitz`) into a union? As mentioned in the introductory comment above, the discrimination is well defined. 
I mean we are currently talking about 12 bytes of saving, but maybe when additional algorithms are added in the future, this can grow further?

Frauschi: TBH I highly doubt presently that there will be any new stateful signature algorithm any time in the nearer future, so this is purely hypothetical for now.

wolfhsm/wh_client_crypto.h
```diff
@@ -3117,5 +3117,76 @@ int wh_Client_MlKemDecapsulateDma(whClientContext* ctx, MlKemKey* key,
...
+/* Bind / read the wolfHSM key id stored in key->devCtx. */
+int wh_Client_LmsSetKeyId(LmsKey* key, whKeyId keyId);
+int wh_Client_LmsGetKeyId(LmsKey* key, whKeyId* outId);
+
+/* Generate an LMS key on the server. The key's parameter set
+ * (levels/height/winternitz) must be bound on the in-memory key before this
+ * call (e.g. via wc_LmsKey_SetParameters). On success the key's devCtx
+ * carries the server-side keyId.
+ *
+ * If flags include WH_NVM_FLAGS_EPHEMERAL, the server returns the public key
```
Frauschi: I really love that the public key is now directly exported back to the client in the keyGen round-trip (that lack has been the main purpose of the `ExportPublicKey` API), but why is this only the case for `EPHEMERAL` keys?
So tbh, I don't see any real-life use case where a user wants an ephemeral stateful key pair in the first place. But independent of that, I see the role of the ephemeral flag solely in the distinction between saving the key in the non-volatile keystore or having it only in RAM (is that assumption correct for overall wolfHSM?). In both cases, one probably needs the public key on the client side to do something useful. So in the non-ephemeral case, the user has to use the `ExportPublicKey()` API to get the public key now. 
TL;DR: I would export the public key in both situations, as the logic is already there anyway.

test/config/user_settings.h
```diff
@@ -140,6 +140,14 @@
 
 /* ML-KEM Options */
 #define WOLFSSL_HAVE_MLKEM
+/* LMS / HSS Options (RFC 8554, NIST SP 800-208) */
+#define WOLFSSL_HAVE_LMS
+#define WOLFSSL_WC_LMS
```
Frauschi: `WOLFSSL_WC_LMS` and `WOLFSSL_WC_XMSS` are not required anymore

src/wh_server_crypto.c
```diff
@@ -879,6 +885,255 @@ int wh_Server_MlKemKeyCacheExport(whServerContext* ctx, whKeyId keyId,
...
+#if (defined(WOLFSSL_HAVE_LMS) || defined(WOLFSSL_HAVE_XMSS)) && \
+    defined(WOLFHSM_CFG_DMA)
+/* Stateful-key persistence bridge.
+ *
+ * wolfCrypt's wc_LmsKey_Sign and wc_XmssKey_Sign require write/read callbacks
+ * for the software path. We wire write_private_key directly to atomic NVM
+ * commit (wh_Nvm_AddObjectWithReclaim): wolfCrypt's contract is to advance
+ * the index, call write_cb, and only emit the signature if write_cb returned
+ * success. That gives us pre-commit-then-emit ordering for free — see
+ * doc/LMS_XMSS_CryptoCb.md and the plan for the crash-safety analysis.
```
Frauschi: the doc is not part of the PR?

Frauschi: Also I think the reference to Claude's plan should be removed.

src/wh_server_crypto.c
```diff
@@ -879,6 +885,255 @@ int wh_Server_MlKemKeyCacheExport(whServerContext* ctx, whKeyId keyId,
...
+static uint16_t _StatefulBridgePrivOffset(const whServerStatefulSigBridge* b)
+{
+    return (uint16_t)(b->hdrSz + b->paramLen + b->pubLen);
+}
+
+/* Update the slot blob's privLen field (header + 6 -> priv length). */
+static void _StatefulBridgeWritePrivLen(uint8_t* slotBuf, uint16_t privLen)
+{
+    /* See wh_crypto.c for layout: privLen is at offset +6. */
+    memcpy(slotBuf + 6, &privLen, sizeof(privLen));
```
Frauschi: I'm not a fan of the `6` magic number here in the code. Can we define a struct for the slot blob layout header (excluding the variable length array at the end)? Then we can cast the `slotBuf` pointer here to the struct and take the address of the `privLen` element.

src/wh_server_crypto.c
```diff
@@ -879,6 +885,255 @@ int wh_Server_MlKemKeyCacheExport(whServerContext* ctx, whKeyId keyId,
...
+ *
+ * The bridge keeps a pointer into the server's cache slot blob (laid out by
+ * wh_Crypto_{Lms,Xmss}SerializeKey). Each write_cb invocation overwrites the
+ * priv region of the slot in place and re-commits the entire slot. */
+typedef struct whServerStatefulSigBridge {
+    whServerContext* server;
+    whKeyId          keyId;
+    whNvmMetadata*   meta;        /* points at the cache slot's metadata */
+    uint8_t*         slotBuf;     /* points at the cache slot's data buffer */
+    uint16_t         hdrSz;       /* offset to priv region inside slotBuf */
```
Frauschi: I can't make sense of the comment at the end of the line. As far as I see the slot blob layout in `wh_crypto.c`, after the header there is the `paramDescriptor` array. Or is "priv region" covering all three arrays of the blob?

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
-    (void)seq;
+    uint16_t pubLen, paramLen;
 
-    int ret = 0;
-    whMessageCrypto_CmacAesDmaRequest  req;
-    whMessageCrypto_CmacAesDmaResponse res;
+    if ((b == NULL) || (server == NULL) || (slotBuf == NULL) || (meta == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+    memcpy(&pubLen,   slotBuf + 4, sizeof(pubLen));
```
Frauschi: Again, I would recommend to create a struct for the header to prevent these magic numbers.

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+    b->pubLen       = pubLen;
+    b->slotCapacity = slotCapacity;
+    return WH_ERROR_OK;
+}
+#endif /* WOLFSSL_HAVE_LMS || WOLFSSL_HAVE_XMSS */
+
+#ifdef WOLFSSL_HAVE_LMS
+/* Dummy cbs used during keygen. wc_LmsKey_MakeKey requires both cbs to be
+ * set; we don't actually persist via cb during keygen because the slot blob
+ * (including pub) is assembled after MakeKey populates key->pub and
```
Frauschi: Smart idea, I like this!

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+    }
 
-    uint8_t tmpKey[AES_256_KEY_SIZE];
-    uint32_t tmpKeyLen = sizeof(tmpKey);
-    Cmac    cmac[1];
+    /* Stream the public key out via the already-validated DMA buffer. The copy
+     * cannot fail, so once the key is committed the client is guaranteed to
+     * receive its keyId. */
+    if (ret == 0) {
+        memcpy(clientPubAddr, key->pub, pubLen32);
```
Frauschi, line 7159: Ah well, looks like the public key is in fact also exported for non-ephemeral keys. That's great! Hence, only the comments need to be adapted to this (see my other comment regarding this).

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+    if (ret == WH_ERROR_OK) {
+        ret = _StatefulBridgeFromSlot(
+            &bridge, ctx, keyId, cacheBuf, cacheMeta,
+            WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE);
+    }
+    if (ret == WH_ERROR_OK) {
+        (void)wc_LmsKey_SetWriteCb(key, _LmsBridgeWriteCb);
+        (void)wc_LmsKey_SetReadCb(key, _LmsBridgeReadCb);
+        (void)wc_LmsKey_SetContext(key, &bridge);
+        ret = wc_LmsKey_Reload(key);
```
Frauschi, line 7249: (Observation, no action required necessarily) So we are actually reading the current private key state twice into the key object. `wh_Crypto_LmsDeserializeKey()` copies the `priv_raw` data into the key object, and `wc_LmsKey_Reload()` again reads it via the read callback. The first copy in the deserialization could actually be removed if desired.

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+    if (ret == WH_ERROR_OK) {
+        ret = wh_Server_DmaProcessClientAddress(
+            ctx, (uintptr_t)req.sig.addr, &sigAddr, req.sig.sz,
+            WH_DMA_OPER_CLIENT_WRITE_PRE, (whServerDmaFlags){0});
+        if (ret != WH_ERROR_OK) {
+            res.dmaAddrStatus.badAddr = req.sig;
+        }
+    }
+    if (ret == WH_ERROR_OK) {
+        /* wolfCrypt's flow (verified against wc_lms.c:1439-1474 post-patch):
```
Frauschi: The "post-patch" could now safely be removed I think

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+        (whMessageCrypto_PqcStatefulSigSigsLeftDmaResponse*)cryptoDataOut);
+    *outSize = sizeof(res);
+    return ret;
+#endif /* WOLFSSL_LMS_VERIFY_ONLY */
+}
+#endif /* WOLFSSL_HAVE_LMS */
+
+#ifdef WOLFSSL_HAVE_XMSS
+/* wolfCrypt's wc_XmssKey_MakeKey calls write_private_key with the freshly
+ * generated sk and then ForceZero's key->sk (see wc_xmss.c). To get a usable
```
Frauschi: Interesting, this difference between LMS and XMSS in the wolfCrypt implementation was not obvious to me! I'm not exactly sure about the rationale here and if the asymmetry is on purpose, but the XMSS way here to capture the private key from the callback is definitely the more robust way. Maybe we should switch the LMS key gen to the same pattern? At least we have to keep that in mind in the case that upstream wolfCrypt adds the ForceZero to LMS as well!

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+
+    ret = wh_Server_KeystoreFreshenKey(ctx, keyId, &cacheBuf, &cacheMeta);
+    if (ret == WH_ERROR_OK) {
+        ret = wc_XmssKey_Init(key, NULL, devId);
+        if (ret == 0) {
+            keyInited = 1;
+        }
+    }
+    if (ret == WH_ERROR_OK) {
+        ret = wh_Crypto_XmssDeserializeKey(cacheBuf, (uint16_t)cacheMeta->len,
```
Frauschi, line 7738: (Observation) In contrast to LMS, the XMSS private key is not read twice (cannot work either as the buffer is not yet allocated). So this is well defined.

src/wh_server_crypto.c
```diff
@@ -879,6 +885,255 @@ int wh_Server_MlKemKeyCacheExport(whServerContext* ctx, whKeyId keyId,
...
+    }
+
+    memcpy(b->slotBuf + privOff, priv, privSz);
+    _StatefulBridgeWritePrivLen(b->slotBuf, (uint16_t)privSz);
+    b->meta->len = (whNvmSize)newLen;
+
+    /* Atomic dual-partition commit. Wolfcrypt aborts the sign if this
+     * returns anything other than _SAVED_TO_NV_MEMORY, so the signature
+     * never escapes for an un-persisted index. */
+    rc = wh_Nvm_AddObjectWithReclaim(b->server->nvm, b->meta, b->meta->len,
```
Frauschi, line 955: Both the LMS and XMSS write callbacks unconditionally write the key data to NVM, also for ephemeral RAM-only keys. As the default wolfCrypt CryptoCb implementation currently generates only ephemeral keys (via `MakeExportKey`), this leads to unreferenced and stalled objects in NVM du to `wh_Server_KeystoreEvictKey()` not deleting the NVM data.

src/wh_client_cryptocb.c
```diff
@@ -793,6 +828,268 @@ static int _handlePqcDecaps(whClientContext* ctx, wc_CryptoInfo* info,
...
+        return WC_HW_E;
+    }
+#endif
+
+    switch (type) {
+#if defined(WOLFSSL_HAVE_LMS) && !defined(WOLFSSL_LMS_VERIFY_ONLY)
+        case WC_PQC_STATEFUL_SIG_TYPE_LMS:
+#ifdef WOLFHSM_CFG_DMA
+            if (useDma) {
+                ret = wh_Client_LmsMakeExportKeyDma(
```
Frauschi, line 887: Should we really create ephemeral keys here by default? I'd highly discourage that and would prefer properly NVM-backed keys.

src/wh_server_crypto.c
```diff
@@ -879,6 +885,255 @@ int wh_Server_MlKemKeyCacheExport(whServerContext* ctx, whKeyId keyId,
...
+#ifdef WOLFSSL_HAVE_XMSS
+int wh_Server_XmssKeyCacheImport(whServerContext* ctx, XmssKey* key,
+                                 const char* paramStr, whKeyId keyId,
+                                 whNvmFlags flags, uint16_t label_len,
+                                 uint8_t* label)
+{
+    int            ret = WH_ERROR_OK;
+    uint8_t*       cacheBuf;
+    whNvmMetadata* cacheMeta;
+    uint16_t       slotCapacity = WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE;
```
Frauschi, line 1101: (AI warning). The default value of `WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE` (1200) is too small for XMSS private keys, so users are forced to override this. This fact should be documented somewhere (or the default is increased when XMSS is enabled).

src/wh_crypto.c
```diff
@@ -488,6 +494,285 @@ int wh_Crypto_MlKemDeserializeKey(const uint8_t* buffer, uint16_t size,
 }
 #endif /* WOLFSSL_HAVE_MLKEM */
 
+#if defined(WOLFSSL_HAVE_LMS) || defined(WOLFSSL_HAVE_XMSS)
+/* Stateful hash-based signature key serialization helpers (LMS / XMSS).
+ *
+ * Slot blob layout:
+ *   uint32_t magic;
```
Frauschi: As noted elsewhere already, I'd suggest to create a struct for this header data.

src/wh_crypto.c
```diff
@@ -488,6 +494,285 @@ int wh_Crypto_MlKemDeserializeKey(const uint8_t* buffer, uint16_t size,
...
+
+#define WH_CRYPTO_STATEFUL_SIG_BLOB_MAGIC_LMS  0x4C4D5301u /* 'LMS\1' */
+#define WH_CRYPTO_STATEFUL_SIG_BLOB_MAGIC_XMSS 0x584D5301u /* 'XMS\1' */
+
+static int _StatefulSigEncodeHeader(uint8_t* buffer, uint32_t magic,
+                                    uint16_t pubLen, uint16_t privLen,
+                                    uint16_t paramLen)
+{
+    uint16_t reserved = 0;
+    memcpy(buffer + 0, &magic, sizeof(magic));
```
Frauschi: With a struct type for the header, we could get rid of all these magic offset numbers here

src/wh_crypto.c
```diff
@@ -488,6 +494,285 @@ int wh_Crypto_MlKemDeserializeKey(const uint8_t* buffer, uint16_t size,
...
+        return WH_ERROR_BADARGS;
+    }
+    memcpy(&magic, buffer + 0, sizeof(magic));
+    if (magic != expectMagic) {
+        return WH_ERROR_BADARGS;
+    }
+    memcpy(pubLen,   buffer + 4, sizeof(*pubLen));
+    memcpy(privLen,  buffer + 6, sizeof(*privLen));
+    memcpy(paramLen, buffer + 8, sizeof(*paramLen));
+    if ((uint32_t)WH_CRYPTO_STATEFUL_SIG_HEADER_SZ + *paramLen + *pubLen +
```
Frauschi: Should we also verify the zerod `reserved` value here as a second check in addition to the magic value in the beginning?

src/wh_crypto.c
```diff
@@ -488,6 +494,285 @@ int wh_Crypto_MlKemDeserializeKey(const uint8_t* buffer, uint16_t size,
...
+        (key->params == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    ret = wc_LmsKey_GetPubLen(key, &pubLen32);
+    if (ret != 0) {
+        return WH_ERROR_BADARGS;
+    }
+    pubLen  = (uint16_t)pubLen32;
+    privLen = (uint16_t)HSS_PRIVATE_KEY_LEN(key->params->hash_len);
```
Frauschi, line 603: The LMS (de)serialization logic does not handle `WOLFSSL_WC_LMS_SERIALIZE_STATE` at the moment. This should definitely be fixed (although it will require multiple #ifdef handlings for this, as there are no helper methods in wolfCrypt for that...).
As `WOLFSSL_WC_LMS_SERIALIZE_STATE` will blow up the private key size, most (if not all) real wolfHSM platforms will probably never turn it on. So another option would be to just reject the value and create a proper compile error when a user tries to enable it.

src/wh_crypto.c
```diff
@@ -488,6 +494,285 @@ int wh_Crypto_MlKemDeserializeKey(const uint8_t* buffer, uint16_t size,
...
+    ret = wc_LmsKey_GetPubLen(key, &expectPubLen);
+    if ((ret != 0) || (expectPubLen != pubLen)) {
+        return WH_ERROR_BADARGS;
+    }
+    if (privLen != (uint16_t)HSS_PRIVATE_KEY_LEN(key->params->hash_len)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    p = buffer + WH_CRYPTO_STATEFUL_SIG_HEADER_SZ + paramLen;
+    memcpy(key->pub, p, pubLen);
```
Frauschi, line 686: (Note) I'd love to see `wc_LmsKey_ImportPubRaw()` used here, but that is not possible unfortunately (as it sets the internal key state to `VERIFYONLY`, and that would cause failures in `wc_LmsKey_Reload()` then...).
Using the `ImportPubRaw()` method here would add additional defense-in-depth checks to verify that the stored key parameters (level, ...) actually match the public key.

src/wh_crypto.c
```diff
@@ -488,6 +494,285 @@ int wh_Crypto_MlKemDeserializeKey(const uint8_t* buffer, uint16_t size,
...
+    p += pubLen;
+    memcpy(key->priv_raw, p, privLen);
```
Frauschi: As noted in wh_server_crypto.c already, this copy is actually unnecessary and could be removed (XMSS does not have it at all, too).

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
+    if (ret != WH_ERROR_OK) {
+        return ret;
+    }
+
+    keyId = wh_KeyId_TranslateFromClient(WH_KEYTYPE_CRYPTO,
+                                         ctx->comm->client_id, req.keyId);
+    if (WH_KEYID_ISERASED(keyId)) {
+        return WH_ERROR_BADARGS;
+    }
+
```
Frauschi, line 7345: I think we also have to lock the NVM here as well (like in the Sign function), as `wh_Server_KeystoreFreshenKey()` may conflict with a concurrent NVM operation (not sure about this tbh). Same goes for XMSS as well.

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
```
Frauschi, line 7451: NVM lock necessary?

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
```
Frauschi, line 7835: NVM lock necessary?

src/wh_server_crypto.c
```diff
@@ -6568,97 +6823,1087 @@ static int _HandlePqcKemAlgorithmDma(whServerContext* ctx, uint16_t magic,
...
```
Frauschi, line 7941: NVM lock necessary?

Frauschi (changes requested):

src/wh_server_crypto.c
padelsbach: Thanks for highlighting this. This is actually not great design imho.

The current design is to no-op the write callback from wolfcrypt, returns `WC_LMS_RC_SAVED_TO_NV_MEMORY` (oops), then wolfHSM does its own write after-the-fact on its own. This is partially because it doesn't want to pass around the big data blob and partially because it just wants to own what is written and do other checks. Worse, it kinda violates the spirit (or event the contract?) of the callback and tries to "outsmart" the sequencing of the callbacks -- which wolfcrypt should own. Doh

However, XMSS does this in a slightly better way, doing _some_ work in the actual callback, but still does not actually perform the write-through to NVM -- it's done after calling wolfcrypt. Doh again

Random observation: XMSS does a forceZero on the privkey, but LMS does not, which partially explains how we got here.

padelsbach: Fwiw: the sign write callback does actually call `wh_Nvm_AddObjectWithReclaim`

padelsbach: So yes, wolfHSM lies to wolfcrypt about the write-through state during keygen, but at least will error out before surfacing the keyID to the HSM client. Correct in practice, but hacky and fragile. I'd argue it doesn't honor the spirit of the callback.

padelsbach: So I'd propose that we fix the write callback for both lms/xmss:

- wolfcrypt tells us the exact blob to store `(const void* data, size_t size)`, no data massaging needed (but cannot be fully opaque to wolfHSM iiuc)
- in wolfhsm, actually write through in the callback
  - move the other associated checks (validate DMA buffer, allocate key) to the callback 
- wolfcrypt runs forceZero on the priv_key and/or sk after keygen to prevent these post-call shenigans

Or just live with it as is since it works?

What do you think @Frauschi ?

src/wh_server_crypto.c
padelsbach: 100% agree. See other comment(s)

wolfhsm/wh_client_crypto.h
padelsbach: Updated to indicate it is a boolean, as well as the function definition to match wolfcrypt

src/wh_client_crypto.c
padelsbach: Fixed

wolfhsm/wh_client_crypto.h
padelsbach: I think this is resolved

test/config/user_settings.h
padelsbach: removed

src/wh_server_crypto.c
padelsbach: removed

src/wh_server_crypto.c
padelsbach: updated, I believe this is resolved now

src/wh_server_crypto.c
padelsbach: Ephemeral lms/xmss keys are disabled now

src/wh_client_cryptocb.c
padelsbach: yep, fixed

padelsbach:
Bunch of updates pushed to address the big items 1-5 in your general comment above. @Frauschi, would you mind reviewing again to see if we are aligned? I'll continue to clean up the smaller items in the various comments.

src/wh_server_crypto.c
Frauschi: I see your point. So for both algorithms, the state write-back via the callback is already safe and sound (the new state is written to NVM directly inside the callback). The only open issue is the key gen.

So the biggest “problem” I see is the different “lifetime” of the data:
- The parameters of the key (three values for LMS, string for XMSS) are valid throughout the key generation process on the server side.
- The new private key state is only available **during** the write callback (enforced in XMSS by zeroing the data in the key object upon returning from `MakeKey`; only a “contract” in LMS, at least for now). 
- The public key is only available **after** the `MakeKey` call.

As a result, we either (1) have to preserve the private key state to after the `MakeKey` call (current approach), or (2) we write the private key state immediately to NVM in the callback and then “extend” the data with the public key after the `MakeKey` call returns. 

From a security perspective, both approaches are fine imho, as the overall function only returns success to the caller once the full key blob is stored in NVM. So the only question is code cleanness and maintainability. With that, I think I'd go with approach (2), but that would require adapting the current “serialization” logic, as a unified serialization in one go is not possible anymore. I think the high-level flow in `_HandleXmssKeyGenDma()` / `_HandleLmsKeyGenDma()` would be something like:
1. Between the key init procedure (parameter loading, callback setting etc.) and the actual `MakeKey` call, get a new id from the keystore (`wh_Server_KeystoreGetUniqueId`), get a new suitable slot (`wh_Server_KeystoreGetCacheSlotChecked`), and write the parameters to the buffer (only RAM).
2. Inside the write-callback within the `MakeKey` call, write the private key state to the buffer (passed via context, this could now even be the same callback as the one for signing) and commit this to NVM via `wh_Nvm_AddObjectWithReclaim`.
3. After the `MakeKey` call returns, read the public key from the key object, also store it in the key slot buffer and commit to NVM again, now covering the whole key blob.

The big disadvantage of this approach is the second NVM commitment. But this should be way cleaner than the current approach without any “hacks” or temporary RAM buffers.

src/wh_server_crypto.c
Frauschi: see response above, I think the best way forward is to completely get rid of this pattern.

Frauschi:
@padelsbach I reviewed your changes and resolved the comments that are already handled. I think most of the bigger issues are already handled, with only the big key gen state handling and the structure for the in memory layout still open.

src/wh_server_crypto.c
padelsbach: I believe I found a solid solution that doesn't involve wolfssl changes. I realized the server can pre-validate some of the fields before even calling `MakeKey`. Then it was just a matter of exposing some data to the write callback via the context. Lmk what you think!

wolfhsm/wh_message_crypto.h
padelsbach: A union could raise an endianness issue. Holding off for now

src/wh_client_crypto.c
padelsbach: done

src/wh_server_crypto.c
padelsbach: Struct added for lengths and offsets

src/wh_server_crypto.c
padelsbach: Updated

src/wh_server_crypto.c
padelsbach: done

src/wh_server_crypto.c
padelsbach: Yep, noted

src/wh_server_crypto.c
padelsbach: Removed

src/wh_server_crypto.c
padelsbach: Good point, this is a flow where the two algs deviate

src/wh_server_crypto.c
padelsbach: updated

src/wh_crypto.c
padelsbach: done

src/wh_crypto.c
padelsbach: done

src/wh_crypto.c
padelsbach: Added

src/wh_server_crypto.c
padelsbach: Good call. Sign has the lock, but the others did not. Updated.

The other server algs (AES/ECC/RSA) could use these too -- but out of scope for this PR

src/wh_server_crypto.c
padelsbach: added

src/wh_server_crypto.c
padelsbach: added

src/wh_server_crypto.c
padelsbach: added

src/wh_crypto.c
padelsbach: added an `#error` for this

src/wh_crypto.c
padelsbach: I think it is required for the SigsLeft operation. I added a comment

padelsbach:
@Frauschi, another big update. Please re-review if you have a chance

src/wh_crypto.c
Frauschi: You are right. In fact, yet another asymmetry between XMSS and LMS, as XMSS's `SigsLeft()` reads the private key via the callback.

This should also be something we address in a follow-up work on wolfCrypt to align the two algorithm implementations (like the `ForceZero` stuff of the LMS private key).

src/wh_server_crypto.c
Frauschi: Hm, I'm still not convinced of the current solution. It is much cleaner than before for sure, but it still "mis-uses" the wolfCrypt API for both algorithms and depends on internal wolfCrypt implementation details and private object members:
- We assume that the public key is fully populated when the write callback is executed
- We assume that the private key state data passed to the buffer is exactly the same as the `key->sk` (XMSS) / `key->priv_raw` (LMS) buffer.
- We access these "private" arguments directly from the key object, bypassing the public getters. 

If any of the above changes inside these implementations, our current approach would break hard.

The biggest problem is that the wolfCrypt API for LMS and XMSS is not designed cleanly to store the private key state data together with the public key in one blob. Nonetheless, I'm very keen on designing a flow now using the public API that won't break in the future. And I think that is possible in a clean and secure way.

I would get rid of the `KeyCacheImport()` and the associated `SerializeKey()` methods for both algorithms. These are only used from within the `KeyGen()` methods anyway. Instead, I would propose the following flow inside the `KeyGen()` methods:
1. Do the existing preparation work (key init, set parameters, do DMA address processing, lock NVM, get new keyId).
2. **New**: get a new key cache slot via wh_Server_KeystoreGetCacheSlotChecked(). The associate buffer for this lives in RAM (the buffer is called `cacheBuf` in the following).
3. **New**: create the `whCryptoStatefulSigHeader` structure with the relevant data and store it together with the parameters into `cacheBuf`.
4. **New**: prepare an updated `whServerStatefulSigKeygenCtx` object that contains a pointer to `cacheBuf` and indicates where the private key state data should be stored in it. 
5. Set the callbacks and then call `MakeKey()`. Inside the (modified) `KeygenWriteCb`, the private key state data is written to the right location within `cacheBuf` based on the context metadata.
6. **New**: after the `MakeKey()` call, read the public key via the getter API. Store it in the right spot within `cacheBuf`. The full blob is now populated in RAM.
7. **New**: finally, commit the key blob from RAM to NVM via wh_Server_KeystoreCommitKey().

The API contract of wolfCrypt does not _force_ us to write the private key state to NVM during the write-callback. As long as we only pass SUCCESS and the new keyId to the caller (the client) after the NVM commitment has been successful, this stays secure and robust against any power failures.

Sorry for dwelling so much on this, but I think getting this right without generating us any future foot-guns with wolfCrypt changes is important. 😅

Let me know what you think about this @padelsbach.

**Update**: I just also took a look at other LMS / XMSS implementations and how these are handling this issue. Many of them actually provide an alternative route via the API to get the private key state directly into a RAM buffer without calling a callback during key gen. So the user can store a full blob in its own way without fiddling with a callback. That could be a future improvement for wolfCrypt as well, but for now we have to handle the callback interface...

Frauschi (changes requested):
Mostly LGTM now, thanks for fixing all the issues!

Two things are still open:
1. The key gen handling with the write-callback, see my new comment in the thread.
2. Some places in the code need more gating for `VERIFY_ONLY`, mostly in wh_crypto.c.

src/wh_server_crypto.c
padelsbach: Very valid points, thanks for the thoughtful review. We are bumping into the LMS/XMSS asymmetry in a few places now.

A key realization for me is that LMS and XMSS have different mechanisms surrounding the key state due to the actual crypto. My overly simplified model is:
- XMSS key state is small and is cached in NVM
- LMS state is large, and only a portion of it is in NVM, and the extended RAM-cached state improves the incremental updates per `Sign()` (actually `WOLFSSL_WC_LMS_SERIALIZE_STATE` would be the mode where we store the full key state in NVM iiuc)

Thus, the API contract actually differs:
- XMSS: generate key, put it in NVM, done with RAM backed key, *ForceZero() it*
- LMS: generate key, put _only_ the `priv_raw` in NVM and _preserve_ the `priv` and `priv_data` in RAM (caller owned), and give to caller for future Sign/Verify, no ForceZero

TLDR: for LMS, the caller owns the lifecycle of the key state in RAM. For XMSS, it can be restored from NVM on-demand.

So I was incorrect about API misuse and would now argue that the writeCb for LMS is actually not needed at all. This was the original design in this PR, oops

padelsbach:
@Frauschi, another update addressing those last two items.

Frauschi (commented):
I love the XMSS key gen flow with the callback and the later NVM commitment now! But why haven't you moved LMS to the same flow? I still think that this would be the safest way for LMS as well, even without its internal `ForceZero` of the private key data. I'd like to discuss this with you 😅

All other things except for the LMS key gen callback flow now LGTM! Great work @padelsbach!

padelsbach:
> why haven't you moved LMS to the same flow?

@Frauschi, good observation. It's two different API contracts on the wolfCrypt side. For XMSS, the priv key is read on-demand and is discarded after the `MakeKey` call. Meanwhile LMS populates the client `key` struct which *intentionally* lives on (for performance reasons I'm guessing). I did document this a bit in the 2 `WriteCb` functions. 

But I can make them more similar if you feel strongly.

Frauschi (dismissed):
LGTM! Thanks for incorporating all my requests and comments! 

Not sure if you want to squash all these commits.

padelsbach:
To whomever merges: please merge w/ squash.

Frauschi (approved):

PR #354, opened 2026-05-06
Add callback for message poll loops

padelsbach:
This hook allows callers of the client library to `sleep()` or do other work while waiting for either the outbound message to enqueue or while waiting for an incoming response. 

The use case is initially to avoid a busy wait in the test refactor in #331, but can be useful as a general purpose addition.

No test cases yet -- will add after #331 is merged.


padelsbach:
Not strictly required. Async crypto will alleviate the need for this as well. Closing

PR #355, opened 2026-05-07
Add refactored crypto tests

padelsbach:


bigbrett (changes requested):
@padelsbach realizing we need a way to iterate over devIds for the DMA-capable algos. Any ideas for how to do this cleanly? Look at how we currently do it in crypto tests with the global devIds array - that solution is def janky but there are some logical groupings of crypto tests that should be run identically just with multiple devIds. 

Do we group these on a per-algo level? That way the high level test function signatures remain the same, but then the cmac test (for example) would internally run its core logic for every devId in a loop?

bigbrett (approved):
@padelsbach looks good. 

Just for tracking, this is largely incomplete as we still need coverage of RSA async, CMAC async + DMA-async, Curve25519 SharedSecretCacheKey, ECC SharedSecretCacheKey, ML-DSA 65/87 and  ML-KEM in its entirety.

PR #357, opened 2026-05-09
add macro guard on SHA512 hashType

JacobBarthelmeh:
This was found when developing and integrating wolfHSM into a project. The autoconf ./configure used to compile wolfSSL that wolfHSM was being linked to did not define WOLFSSL_SHA512_HASHTYPE. The changes in this PR allow for linking against wolfSSL compiled without WOLFSSL_SHA512_HASHTYPE and add a CI test case.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -45,6 +45,11 @@ jobs:
     - name: Build and test DMA ASAN
       run: cd test && make clean && make -j DMA=1 ASAN=1 WOLFSSL_DIR=../wolfssl && make run
 
+    # Build and test with DMA, SHA-512 enabled in wolfSSL, but with
+    # WOLFSSL_SHA512_HASHTYPE disabled.
+    - name: Build and test DMA without WOLFSSL_SHA512_HASHTYPE
+      run: cd test && make clean && make -j DMA=1 NO_SHA512_HASHTYPE=1 WOLFSSL_DIR=../wolfssl && make run
+
     # Build and test ASAN build, with wolfCrypt tests enabled.
     - name: Build and test ASAN TESTWOLFCRYPT
       run: cd test && make clean && make -j ASAN=1 TESTWOLFCRYPT=1 WOLFSSL_DIR=../wolfssl && make run
--- a/src/wh_server_crypto.c
+++ b/src/wh_server_crypto.c
@@ -5215,7 +5215,9 @@ static int _HandleSha512Dma(whServerContext* ctx, uint16_t magic, int devId,
     sha512->loLen    = req.resumeState.loLen;
     sha512->hiLen    = req.resumeState.hiLen;
     sha512->buffLen  = 0;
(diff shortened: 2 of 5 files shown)


padelsbach (approved):

padelsbach:
@JacobBarthelmeh, please rebase to pass CI

bigbrett:
@JacobBarthelmeh @padelsbach  unless I'm misunderstanding the intent here, this one is a little silly/overkill to expose as a test makefile toggle. I really dont want to add every single switch (especially a trivial one like this) as a configurable option, lest our CI test suite starts blowing up. No end user is going to care about toggling this. If wolfHSM requires hashtype on, then it should require it on across the board. It is perfectly fine for certain wolfCrypt flags to be mandatory. If it is optional or only tied to a specific wolfHSM feature (like SHA512 DMA in this case) then the source code should be able to handle it on and off based on the proper feature guards or a check in `wh_settings.h` telling the user to turn it on (so we dont need redundant gates everywhere in the code). Things like this are fine to test once. Adding regression/edge cases for every single permutation of a build will result in our PR CI times blowing up and wont really add any value. 

Otherwise EVERY single wolfCrypt config option that is relevant to a subset of features will need something like this and our code and tests will turn into spaghetti.

The POSIX tests really should be testing with everything on. The only things toggleable should me MAJOR wolfHSM features that are truly optional in the sense that a user may want to disable an entire subsystem (SHE, DMA, etc). No reason we really would ever want to make this high-level toggleable for an end user. A compile error saying "hey if you have SHA512 DMA you need the hashtype turned on in wolfCrypt" is plenty.

JacobBarthelmeh:
> @JacobBarthelmeh @padelsbach unless I'm misunderstanding the intent here, this one is a little silly/overkill to expose as a test makefile toggle. I really dont want to add every single switch (especially a trivial one like this) as a configurable option, lest our CI test suite starts blowing up. No end user is going to care about toggling this. If wolfHSM requires hashtype on, then it should require it on across the board. It is perfectly fine for certain wolfCrypt flags to be mandatory. If it is optional or only tied to a specific wolfHSM feature (like SHA512 DMA in this case) then the source code should be able to handle it on and off based on the proper feature guards or a check in `wh_settings.h` telling the user to turn it on (so we dont need redundant gates everywhere in the code). Things like this are fine to test once. Adding regression/edge cases for every single permutation of a build will result in our PR CI times blowing up and wont really add any value.
> 
> Otherwise EVERY single wolfCrypt config option that is relevant to a subset of features will need something like this and our code and tests will turn into spaghetti.
> 
> The POSIX tests really should be testing with everything on. The only things toggleable should me MAJOR wolfHSM features that are truly optional in the sense that a user may want to disable an entire subsystem (SHE, DMA, etc). No reason we really would ever want to make this high-level toggleable for an end user. A compile error saying "hey if you have SHA512 DMA you need the hashtype turned on in wolfCrypt" is plenty.

I can add a PR to revert the CI test for it. I still think we should allow the option to build without SHA512 hash type defined though.

PR #359, opened 2026-05-11
Fix keywrap response size checks

padelsbach:
No new tests (will implement a new test harness for testing server inputs).

Fixes F-3104


test-refactor/misc/wh_test_keywrap_respsize.c
```diff
@@ -0,0 +1,232 @@
+/*
```
padelsbach: New test case might be overkill for these simple fixes -- tried to keep them as simple as possible...

bigbrett: hmm I also dont like the idea of having fake transport spoofing style tests just to ensure we are checking args correctly. Does smell like overkill, at least for now.

Lets phrase the problem in a different manner - assume we want to eventually get to 100% code coverage...given the new test framework, how would we structure it such that the higher level behavioral integration tests can be somewhat isolated from the API/unit/spammy style tests that just ensure "assert" behavior on args. 

Does it make sense to have a totally separate group of "API/arg" tests, also organized by module?

I know this is overkill to discuss this for such a small PR but I'd rather iron these details out up front as we are integrating. 

Thoughts?

padelsbach: I think this falls into a whole category of tests where the server sends back something invalid/corrupted/wrong. The check being fixed here is defense against that, and we're in the process of adding and fixing a bunch of those checks.

A quick client-only test bench could simulate bad inputs from the server. That would help coverage a lot. I've added this as a quick line in the readme.

I'll remove the new (bloated) test file for now.

bigbrett (approved):

PR #361, opened 2026-05-11
Add size check in wh_Client_Curve25519SharedSecret

padelsbach:
Fixes F-3799


bigbrett (changes requested):
Hmm could we actually change this to mirror how we do it in [ECDH](https://github.com/wolfSSL/wolfHSM/blob/80ee2d59ecbfd46bd4b7b6ab8ed742ca5dac3e0f/src/wh_client_crypto.c#L1908) and simply gating the memcpy on out_size != NULL? As written, when a caller passes an output buffer but no size pointer, the function still writes the server's data into that buffer without checking it fits. Would simplify the logic

bigbrett (approved):

PR #364, opened 2026-05-13
Align data in wolfboot test structs

padelsbach:
Fixes this test failure on macos, adds `macos-latest` to unit test runs in CI.

```
ERROR [whTest_ServerImgMgrServerCfgWolfBootRsa4096:1275]: wolfBoot verify method failed: -2000
ERROR [whTest_ServerImgMgr:1527]: wolfBoot RSA4096 image manager tests failed: -2000
Assertion failed: ((0 == whTest_ServerImgMgr(WH_NVM_TEST_BACKEND_FLASH))), function whTest_Unit, file wh_test.c, line 123.
make: *** [run] Abort trap: 6
```


test/gen/wh_test_wolfboot_img_data.h
```diff
@@ -32,7 +32,7 @@
 #include <stddef.h>
 
 /* wolfBoot signed header (RSA4096+SHA256, 1024 bytes) */
-static const uint8_t wolfboot_test_header[] = {
+static const uint8_t wolfboot_test_header[] __attribute__((aligned(4))) = {
```
bigbrett: hmmm unfortunately this is clang/gcc specific and some target toolchains wont support it. I do want us to go to C11 but for now we are C99 :/ 

I would say we should probably add `WH_ALIGN(_x)` to `wh_settings.h`  that can have per-architecture definitions. For now you can just toolchain detect clang and GCC. Believe we do this elsewhere. But in the future for other toolchains we can add support for those as they come online. Then this header will need to include settings.h which i'm sure will break everything else :(

bigbrett (changes requested):

test/gen/wh_test_wolfboot_img_data.h
padelsbach: Updated to use `XALIGNED(4)` from wolfcrypt. Let me know if you'd rather create a separate macro in wolfhsm.

bigbrett: thats fine for now. Will wait until the other fix clears CI then u can rebase this

test-refactor/posix/wh_test_posix_main.c
```diff
@@ -55,7 +55,8 @@
  * meaningful (or portable) on bare-metal client targets. */
 #if defined(WOLFHSM_CFG_THREADSAFE) \
     && defined(WOLFHSM_CFG_GLOBAL_KEYS) \
-    && !defined(WOLFHSM_CFG_NO_CRYPTO)
+    && !defined(WOLFHSM_CFG_NO_CRYPTO) \
+    && !defined(__APPLE__) /* pthread_barrier_t unavailable on macOS */
```
bigbrett, line 59: lol you really cant win on macOS here can you

bigbrett (approved):

PR #365, opened 2026-05-14
F-3805: length check in wh_Client_CertReadTrustedResponse

padelsbach:


test-refactor/wh_test_list.c
```diff
@@ -52,6 +53,7 @@ const whTestCase whTestsServer[] = {
 const size_t whTestsServerCount = sizeof(whTestsServer) / sizeof(whTestsServer[0]);
 
 const whTestCase whTestsClient[] = {
+    { "whTest_CertReadTrustedSmallBuffer", whTest_CertReadTrustedSmallBuffer },
```
bigbrett: instead of being "flat" where each individual test is registered here, how do we feel about this just holding `whTest_ClientCerts()` which is the main driver for all client-side cert tests, then the new test you added could be a static function (as well as all the other cert tests that will be added) but they are all grouped together under one "functional" suite of tests.

For example, we already essentially do this with the crypto algo tests - just have one `whTest_CryptoAes()` and not `whTest_CryptoAes128CheckBadArgsForKey()` with nine million other permutations like that

bigbrett (changes requested):
Yay for using new test harness

test-refactor/wh_test_list.c
padelsbach: Good idea, makes for smaller diffs as tests are added. Coming right up...

padelsbach: done

bigbrett (approved):

PR #367, opened 2026-05-14
F-3803: validate response sizes

padelsbach:


src/wh_client_crypto.c
```diff
@@ -3409,10 +3409,13 @@ int wh_Client_Ed25519Sign(whClientContext* ctx, ed25519_key* key,
                 ret = _getCryptoResponse(dataPtr, WC_PK_TYPE_ED25519_SIGN,
                                          (uint8_t**)&res);
                 if (ret >= 0) {
-                    uint32_t res_total =
+                    const uint32_t res_len_min =
```
bigbrett: we have a tidier idiom for this and the associated length check used elsewhere:

```  if (ret >= 0) {                                                                                                                                             
      const size_t hdr_sz =                                                                                                                                   
      if (res_len < hdr_sz || res->sigSz > (res_len - hdr_sz)) {                
          ret = WH_ERROR_ABORTED;
      }                                                                         
  }    
```

could we just keep the "header" terminology for the subheader and one line if comparison?

bigbrett (changes requested):
looks fine, just a nit too unify style

bigbrett (approved):
@padelsbach lgtm - seems same issues are missing in ML DSA sign and all the DMA variants. Could we add that to the list in a subsequent PR plz?

PR #372, opened 2026-05-18
cache-only shared secrets

bigbrett:
This PR adds cache-mode shared-secret support so ECDH and X25519 outputs can remain server-resident instead of always being returned to the client.

Changes:

-  Extends ECDH and Curve25519 request/response messages with cache flags, output key IDs, and labels. 
-  Adds client/server handling for cached shared-secret outputs using raw key-cache import.
-  Adds crypto tests covering cached ECDH/X25519 secrets, caller-supplied IDs, NONEXPORTABLE behavior, and guard cases.


src/wh_server_crypto.c
```diff
@@ -966,6 +971,35 @@ static int _HandleEccSharedSecret(whServerContext* ctx, uint16_t magic,
...
+            ret = wh_Server_KeystoreGetUniqueId(ctx, &out_key_id);
+        }
+        if (ret == WH_ERROR_OK) {
+            ret =
+                wh_Server_KeyCacheImportRaw(ctx, res_out, res_len, out_key_id,
+                                            flags, WH_NVM_LABEL_LEN, req.label);
+        }
+        /* Scrub the secret from the response buffer regardless of import
+         * success/failure. */
+        memset(res_out, 0, res_len);
```
rizlik, line 1077: would you consider getting a cache slot with `wh_Server_KeystoreGetCacheSlotChecked` before the `wc_ecc_shared_secret` operation to avoid a copy and a memset?

src/wh_server_crypto.c
```diff
@@ -966,6 +971,35 @@ static int _HandleEccSharedSecret(whServerContext* ctx, uint16_t magic,
...
         wc_ecc_free(pub_key);
     }
+
+    /* If caching, move the secret out of the response buffer into a cache
+     * slot and return only its keyId. */
+    if ((ret == WH_ERROR_OK) && cache) {
+        out_key_id = wh_KeyId_TranslateFromClient(
+            WH_KEYTYPE_CRYPTO, ctx->comm->client_id, req.keyId);
+        if (WH_KEYID_ISERASED(out_key_id)) {
+            ret = wh_Server_KeystoreGetUniqueId(ctx, &out_key_id);
```
rizlik: this is probably racey under THREAD_SAFE

rizlik (changes requested):

src/wh_server_crypto.c
bigbrett: Fixed

src/wh_server_crypto.c
bigbrett: OK I thought about this but it means that the crypto would have to occur inside the critical section, otherwise there could be a race on getting a cache slot for global keys. For now I think its best to follow the pattern of the other handlers that work similarly. Would rather pay for a small copy inside the critical section than pay for a public key crypto operation

rizlik: You can use `wh_Server_KeystoreGetUniqueId` and `wh_Server_KeystoreGetCacheSlot` under the critical section, the public key operation can stay outside.
I'm ok with keeping the copy as well.

rizlik (approved):

PR #377, opened 2026-05-20
ML-DSA rename

Frauschi:
Move to the new `wc_MlDsaKey` API and get rid of all "Dilithium" naming.


padelsbach:
benchmark/bench_modules/wh_bench_mod_mldsa.c still uses "dilithium". Can we update that file too?

Frauschi:
> benchmark/bench_modules/wh_bench_mod_mldsa.c still uses "dilithium". Can we update that file too?

Oh, thanks for the hint! Fixed now.

padelsbach (approved):

PR #379, opened 2026-05-21
f-2010 : Add bound check for aes-ctr

miyazakh:
# Fix: AES-CTR `left` field out-of-bounds read in `_HandleAesCtr` and `_HandleAesCtrDma`

## Background

wolfHSM's AES-CTR streaming API preserves keystream state across calls by round-tripping
`aes->left` and `aes->tmp` between the client and the HSM server.  The `left` field
indicates how many cached keystream bytes remain in `aes->tmp[AES_BLOCK_SIZE]`.

`wc_AesCtrEncrypt` internally indexes into `aes->tmp` as:

```c
aes->tmp[AES_BLOCK_SIZE - aes->left]
```

If `aes->left > AES_BLOCK_SIZE`, the subtraction wraps (unsigned underflow), producing a
very large index and reading beyond the end of `aes->tmp` into adjacent struct memory
(potentially including key material).  The XOR'd bytes from that out-of-bounds read are
written to the output buffer and returned to the caller.

## Problem

Both `_HandleAesCtr` and `_HandleAesCtrDma` restored the client-supplied `left` value
without bounds validation:

```c
// Before fix — in both handlers
if (left > AES_BLOCK_SIZE) {
    ret = WH_ERROR_BADARGS;   // sets ret …
}
aes->left = left;             // … but assignment executes unconditionally
if (enc != 0) {               // guard checks enc, not ret
    wc_AesCtrEncrypt(...);    // called even when left is out of range
}
```

The bounds check set `ret = WH_ERROR_BADARGS`, but `aes->left = left` and the subsequent
`wc_AesCtrEncrypt` call were **not** gated on `ret`.  Because the encrypt call used
`if (enc != 0)` (not `if (ret == WH_ERROR_OK)`), a compromised client core could send
`left > 16` and trigger the out-of-bounds read on every request, crossing the HSM trust
boundary.

## Fix

Wrap the context restoration and the encrypt call inside an `else` branch so they are
skipped entirely when `left > AES_BLOCK_SIZE`.

**`_HandleAesCtr` (`src/wh_server_crypto.c`)**

```c
if (left > AES_BLOCK_SIZE) {
    ret = WH_ERROR_BADARGS;
}
else {
    /* Restore streaming CTR context from the previous call. */
    aes->left = left;
    memcpy(aes->tmp, tmp, sizeof(aes->tmp));
    if (enc != 0) {
        ret = wc_AesCtrEncrypt(aes, out, in, len);
    }
    else {
        ret = wc_AesCtrEncrypt(aes, out, in, len);
    }
}
```

**`_HandleAesCtrDma` (`src/wh_server_crypto.c`)** — identical structure; the
post-encrypt metadata copies (`left`, `out_tmp`, `out_iv`) are also moved inside the
`else` branch so they are never reached with an invalid `left`.

## Tests Added

**`test/wh_test_crypto.c`**

| Function | Test description |
|---|---|
| `whTest_CryptoAesAsync` | Sets `aes->left = AES_BLOCK_SIZE + 1`, calls `wh_Client_AesCtr`, expects `WH_ERROR_BADARGS` |
| `whTest_CryptoAesDmaAsync` | Same check via `wh_Client_AesCtrDma` (active when built with `DMA=1`) |

Both tests verify that the server rejects the request before calling `wc_AesCtrEncrypt`,
confirming no out-of-bounds read can occur.

## Files Changed

| File | Change |
|---|---|
| `src/wh_server_crypto.c` | Add `else` guard in `_HandleAesCtr` and `_HandleAesCtrDma` |
| `test/wh_test_crypto.c` | Add bounds-check rejection tests for non-DMA and DMA paths |

### DIFF
--- a/test-refactor/wh_test_list.c
+++ b/test-refactor/wh_test_list.c
@@ -43,6 +43,7 @@ WH_TEST_DECL(whTest_CertVerify);
 WH_TEST_DECL(whTest_NvmOptional);
 WH_TEST_DECL(whTest_ClientCerts);
 WH_TEST_DECL(whTest_Crypto_Aes);
+WH_TEST_DECL(whTest_CryptoAesCtrLeftOob);
 WH_TEST_DECL(whTest_CryptoAesKeyUsagePolicies);
 WH_TEST_DECL(whTest_Crypto_Cmac);
 WH_TEST_DECL(whTest_Crypto_Curve25519);
@@ -74,6 +75,7 @@ const size_t whTestsServerCount = sizeof(whTestsServer) / sizeof(whTestsServer[0
 const whTestCase whTestsClient[] = {
     { "whTest_ClientCerts", whTest_ClientCerts },
     { "whTest_Crypto_Aes", whTest_Crypto_Aes },
+    { "whTest_CryptoAesCtrLeftOob", whTest_CryptoAesCtrLeftOob },
     { "whTest_CryptoAesKeyUsagePolicies", whTest_CryptoAesKeyUsagePolicies },
     { "whTest_Crypto_Cmac", whTest_Crypto_Cmac },
     { "whTest_Crypto_Curve25519", whTest_Crypto_Curve25519 },
--- a/test-refactor/wh_test_list.c
+++ b/test-refactor/wh_test_list.c
@@ -43,7 +43,6 @@ WH_TEST_DECL(whTest_CertVerify);
 WH_TEST_DECL(whTest_NvmOptional);
 WH_TEST_DECL(whTest_ClientCerts);
 WH_TEST_DECL(whTest_Crypto_Aes);
-WH_TEST_DECL(whTest_CryptoAesCtrLeftOob);
 WH_TEST_DECL(whTest_CryptoAesKeyUsagePolicies);
 WH_TEST_DECL(whTest_Crypto_Cmac);
 WH_TEST_DECL(whTest_Crypto_Curve25519);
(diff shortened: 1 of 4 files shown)


bigbrett:
@miyazakh could you also please add the same test to the new refactored test framework in the `test-refactor` directory? We want all tests going forward to be added in both places to streamline the migration to the new system. It is a good chance to get exposed to the new test harness.

dgarske (changes requested):
Please fix merge conflicts

dgarske (dismissed):
Over to @bigbrett

test-refactor/wh_test_list.c
```diff
@@ -73,6 +74,7 @@ const size_t whTestsServerCount = sizeof(whTestsServer) / sizeof(whTestsServer[0
 const whTestCase whTestsClient[] = {
     { "whTest_ClientCerts", whTest_ClientCerts },
     { "whTest_Crypto_Aes", whTest_Crypto_Aes },
+    { "whTest_CryptoAesCtrLeftOob", whTest_CryptoAesCtrLeftOob },
```
bigbrett: We don't need a new top-level test for this. Instead, this should be a static function called as part of``whTest_Crypto_Aes()`.

bigbrett (changes requested):
One quick test re-org, otherwise looks great!

test-refactor/wh_test_list.c
miyazakh: Changed the test as `static`

bigbrett (approved):

PR #380, opened 2026-05-21
f-2708: Add test cases for KeyUsage

miyazakh:
# Test: Add key usage policy enforcement tests for AES CTR/ECB/GCM and all DMA handlers

## Background

`wh_Server_KeystoreEnforceKeyUsage` is called in every AES cipher handler on the server
side to prevent a cached key from being used in an operation its policy does not permit
(e.g., an ENCRYPT-only key being used to decrypt).  However, the existing positive tests
in `whTestCrypto_Aes` always cache keys with both `WH_NVM_FLAGS_USAGE_ENCRYPT` and
`WH_NVM_FLAGS_USAGE_DECRYPT` set, so the enforcement logic is never actually exercised.

## Problem

Three gaps were identified:

1. **No negative tests for CTR, ECB, or GCM** — only `_HandleAesCbc` had rejection tests
   (lines 5101–5214 of `wh_test_crypto.c`).  Deleting the enforcement call from any of
   the other three handlers caused zero test failures.

2. **No tests at all for the DMA handlers** — `_HandleAesCtrDma`, `_HandleAesEcbDma`,
   `_HandleAesCbcDma`, and `_HandleAesGcmDma` each contain enforcement code but had no
   test coverage of any kind.

## Fix

Added negative tests to `whTest_CryptoKeyUsagePolicies` in `test/wh_test_crypto.c`.
Each test caches a key with restricted flags, then attempts the forbidden operation and
asserts that the server returns `WH_ERROR_USAGE`.

### Non-DMA handlers

| Mode | Test | Key flags cached |
|---|---|---|
| AES-CTR | encrypt (`enc=1`) without ENCRYPT flag | `WH_NVM_FLAGS_NONE` |
| AES-CTR | decrypt (`enc=0`) without DECRYPT flag | `WH_NVM_FLAGS_USAGE_ENCRYPT` |
| AES-ECB | encrypt without ENCRYPT flag | `WH_NVM_FLAGS_NONE` |
| AES-ECB | decrypt without DECRYPT flag | `WH_NVM_FLAGS_USAGE_ENCRYPT` |
| AES-GCM | encrypt without ENCRYPT flag | `WH_NVM_FLAGS_NONE` |
| AES-GCM | decrypt without DECRYPT flag | `WH_NVM_FLAGS_USAGE_ENCRYPT` |

### DMA handlers (active when built with `DMA=1`)

| Mode | Test |
|---|---|
| AES-CTR DMA | encrypt / decrypt rejection |
| AES-ECB DMA | encrypt / decrypt rejection |
| AES-CBC DMA | encrypt / decrypt rejection |
| AES-GCM DMA | encrypt / decrypt rejection |

### Notes

- AES-CTR uses `wh_Client_AesCtr` directly (with `enc=0` / `enc=1`) because
  `wc_AesCtrEncrypt` is symmetric and always sends `enc=1` through the wolfcrypt
  crypto callback.
- AES-ECB and AES-GCM use the standard wolfcrypt functions (`wc_AesEcbEncrypt`,
  `wc_AesGcmDecrypt`, etc.), which set the direction flag correctly through the callback.
- For GCM DMA decrypt, `dec_tag` is passed as a non-NULL zero buffer to satisfy the
  client-side `BADARGS` guard; the server rejects the request at the key usage check
  before any cryptographic operation or DMA address processing occurs.

## Files Changed

| File | Change |
|---|---|
| `test/wh_test_crypto.c` | Add 14 rejection tests (6 non-DMA + 8 DMA) to `whTest_CryptoKeyUsagePolicies` |

### DIFF
--- a/test/wh_test_crypto.c
+++ b/test/wh_test_crypto.c
@@ -13297,6 +13297,544 @@ int whTest_CryptoKeyUsagePolicies(whClientContext* client, WC_RNG* rng)
     if (ret != 0)
         return ret;
 #endif /* HAVE_AES_CBC */
+
+#ifdef WOLFSSL_AES_COUNTER
+    /* AES-CTR: encrypt without ENCRYPT flag */
+    WH_TEST_PRINT("  Testing AES CTR encrypt without ENCRYPT flag...\n");
+    {
+        Aes     aes[1];
+        uint8_t iv[AES_BLOCK_SIZE] = {0};
+        uint8_t ctrCipher[16]      = {0};
+
+        keyId = WH_KEYID_ERASED;
+        ret   = wh_Client_KeyCache(client, WH_NVM_FLAGS_NONE,
+                                   (uint8_t*)"ctr-no-enc", strlen("ctr-no-enc"),
+                                   key, keyLen, &keyId);
+        if (ret == 0) {
+            ret = wc_AesInit(aes, NULL, WH_DEV_ID);
+            if (ret == 0) {
+                ret = wh_Client_AesSetKeyId(aes, keyId);
+                if (ret == 0) {
+                    (void)wc_AesSetIV(aes, iv);
+                    ret = wh_Client_AesCtr(client, aes, 1, plaintext,
+                                           sizeof(plaintext), ctrCipher);
+                    if (ret == WH_ERROR_USAGE) {
+                        WH_TEST_PRINT("    PASS: Correctly denied encryption\n");
+                        ret = 0;
+                    }
+                    else {
+                        WH_ERROR_PRINT(
+                            "    FAIL: Expected WH_ERROR_USAGE, got %d\n", ret);
+                        ret = WH_ERROR_ABORTED;
+                    }
+                }
+                wc_AesFree(aes);
+            }
+            wh_Client_KeyEvict(client, keyId);
+        }
+    }
+    if (ret != 0)
+        return ret;
+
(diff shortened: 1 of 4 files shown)


test/wh_test_crypto.c
```diff
@@ -10503,6 +10503,556 @@ int whTest_CryptoKeyUsagePolicies(whClientContext* client, WC_RNG* rng)
...
+                        ret = WH_ERROR_ABORTED;
+                    }
+                }
+                wc_AesFree(aes);
+            }
+            wh_Client_KeyEvict(client, keyId);
+        }
+    }
+    if (ret != 0)
+        return ret;
```
bigbrett: no braceless if statements please. There are lots in this PR

bigbrett (changes requested):

bigbrett:
@miyazakh could you also please add the same test to the new refactored test framework in the `test-refactor` directory? We want all tests going forward to be added in both places to streamline the migration to the new system. It is a good chance to get exposed to the new test harness.

test/wh_test_crypto.c
miyazakh: Add braces to braceless if statements in key usage tests

dgarske (changes requested):
Please fix merge conflicts

dgarske (approved):
Over to @bigbrett

bigbrett (approved):

PR #383, opened 2026-05-28
logout user in the case that the underlying connection gets disconnected

JacobBarthelmeh:
Follow up to item 4 from https://github.com/wolfSSL/wolfHSM/pull/270

### DIFF
--- a/src/wh_auth.c
+++ b/src/wh_auth.c
@@ -184,7 +184,8 @@ int wh_Auth_Logout(whAuthContext* context, whUserId user_id)
 }
 
 
-/* returns result of locking, clears context user regardless of lock success. */
+/* Clears the local session under the auth lock. Fails without clearing if
+ * the lock can not be acquired, to avoid racing an in-flight login. */
 int wh_Auth_Reset(whAuthContext* context)
 {
     int rc;
@@ -193,12 +194,11 @@ int wh_Auth_Reset(whAuthContext* context)
         return WH_ERROR_BADARGS;
     }
 
-    /* Best-effort lock, but clear the session regardless */
     rc = WH_AUTH_LOCK(context);
-    memset(&context->user, 0, sizeof(whAuthUser));
     if (rc == WH_ERROR_OK) {
+        memset(&context->user, 0, sizeof(whAuthUser));
         (void)WH_AUTH_UNLOCK(context);
-    }
+    } /* LOCK() */
     return rc;
 }
 
--- a/src/wh_auth.c
+++ b/src/wh_auth.c
@@ -184,6 +184,25 @@ int wh_Auth_Logout(whAuthContext* context, whUserId user_id)
 }
 
 
+/* returns result of locking, clears context user regardless of lock success. */
+int wh_Auth_Reset(whAuthContext* context)
+{
+    int rc;
+
+    if (context == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Best-effort lock, but clear the session regardless */
+    rc = WH_AUTH_LOCK(context);
+    memset(&context->user, 0, sizeof(whAuthUser));
+    if (rc == WH_ERROR_OK) {
+        (void)WH_AUTH_UNLOCK(context);
+    }
+    return rc;
+}
+
+
 /* Check on request authorization and action permissions for current user
  * logged in */
 int wh_Auth_CheckRequestAuthorization(whAuthContext* context, uint16_t group,
--- a/src/wh_server.c
+++ b/src/wh_server.c
@@ -93,10 +93,12 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
         server->auth->user.user_id != WH_USER_ID_INVALID) {
         whUserId stale_id  = server->auth->user.user_id;
         int      logout_rc = wh_Auth_Logout(server->auth, stale_id);
-        if (logout_rc != WH_ERROR_OK ||
-            server->auth->user.user_id != WH_USER_ID_INVALID) {
-            memset(&server->auth->user, 0, sizeof(server->auth->user));
+        if (logout_rc != WH_ERROR_OK) {
+            WH_LOG(&server->log, WH_LOG_LEVEL_SECEVENT,
+                   "Stale auth session force-cleared during server init after "
+                   "logout failure");
         }
+        (void)wh_Auth_Reset(server->auth);
     }
 #endif /* WOLFHSM_CFG_ENABLE_AUTHENTICATION */
 #ifdef WOLFHSM_CFG_HWKEYSTORE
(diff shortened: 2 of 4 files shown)


src/wh_server.c
```diff
@@ -87,6 +87,17 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
     server->nvm = config->nvm;
 #ifdef WOLFHSM_CFG_ENABLE_AUTHENTICATION
     server->auth = config->auth;
+    /* auth context is externally owned; clear any stale session left over from
+     * a prior connection (Logout first so backend callback runs). */
+    if (server->auth != NULL &&
+        server->auth->user.user_id != WH_USER_ID_INVALID) {
+        whUserId stale_id  = server->auth->user.user_id;
+        int      logout_rc = wh_Auth_Logout(server->auth, stale_id);
```
bigbrett: would we ever want to propagate this failure back to the caller?

src/wh_server.c
```diff
@@ -177,6 +188,27 @@ int wh_Server_SetConnected(whServerContext *server, whCommConnected connected)
...
 
+#ifdef WOLFHSM_CFG_ENABLE_AUTHENTICATION
+    /* Log out any active user on disconnect, including abrupt drops where
+     * COMM_CLOSE never arrives. */
+    if (connected == WH_COMM_DISCONNECTED &&
+        server->connected != WH_COMM_DISCONNECTED &&
+        server->auth != NULL &&
+        server->auth->user.user_id != WH_USER_ID_INVALID) {
+        whUserId user_id   = server->auth->user.user_id;
+        int      logout_rc = wh_Auth_Logout(server->auth, user_id);
```
bigbrett: would we ever want to propagate this failure back to the caller? Perhaps not, just trying to think it through.

test/wh_test_auth.c
```diff
@@ -1415,6 +1428,41 @@ int whTest_AuthTCP(whClientConfig* clientCfg)
 }
 
 
+#if !defined(WOLFHSM_CFG_TEST_CLIENT_ONLY_TCP) && \
+    defined(WOLFHSM_CFG_ENABLE_SERVER)
+static int _whTest_Auth_AbruptDisconnect(whClientContext* client_ctx)
```
bigbrett: nit: no need for `whTest` prefix if its a static function

bigbrett (dismissed):
LGTM. A few questions and  a nit but nothing that would block merge

src/wh_server.c
```diff
@@ -177,6 +188,27 @@ int wh_Server_SetConnected(whServerContext *server, whCommConnected connected)
...
+
+        /* wh_Auth_Logout only clears the session when the backend callback
+         * returns WH_ERROR_OK. Force-clear if it failed or left the user set,
+         * so a stale identity can never carry over to the next connection
+         * (mirrors the defensive clear in wh_Server_Init). */
+        if (logout_rc != WH_ERROR_OK ||
+            server->auth->user.user_id != WH_USER_ID_INVALID) {
+            memset(&server->auth->user, 0, sizeof(server->auth->user));
```
billphipps: Consider unconditionally zeroing this.  Also above too.  Losing privileges should be no-fail, just like destructors.  Gaining privileges is a must-pass.

billphipps (dismissed):
Please consider unconditionally zeroing userid when logging out. Please resolve additional comments.

src/wh_server.c
JacobBarthelmeh: The propagating back of the return values is a good point. When writing this I had opted for the behavior of these logout's being used with bad connection states and the thought of not holding up a fresh restart if the authentication state was stale. Kind of a last resort and reset to remove the stale user ID. It is missing log output for the fail cases though.

src/wh_server.c
JacobBarthelmeh: Added logging for logout failure and unconditional user reset.

src/wh_server.c
JacobBarthelmeh: Added wh_Auth_Reset for an internal to wolfHSM user reset function and called it unconditionally. It also tries to make use of the lock when available before calling memset.

test/wh_test_auth.c
JacobBarthelmeh: updated

JacobBarthelmeh:
@bigbrett and @billphipps , I believe I've addressed the current feedback. Thank you for reviewing

src/wh_auth.c
```diff
@@ -184,6 +184,25 @@ int wh_Auth_Logout(whAuthContext* context, whUserId user_id)
...
+int wh_Auth_Reset(whAuthContext* context)
+{
+    int rc;
+
+    if (context == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Best-effort lock, but clear the session regardless */
+    rc = WH_AUTH_LOCK(context);
```
bigbrett, line 197: @JacobBarthelmeh Seems odd to still clear the session if you can't acquire the lock - this means someone else could be in process of reading. Perhaps the backend would get put into a weird state if synchronous access isn't respected? I'd rather just hard fail if you can't acquire the lock, and let the caller deal with it since it probably means their entire system is broken? I guess you could make the argument that the locking function really shouldn't ever fail, but kind of a bucket of worms. Thoughts?

JacobBarthelmeh: I'm okay with changing it to be a hard fail when the lock fails. I think that is better behavior than what I'd added for continuing to memset anyway.

bigbrett: @JacobBarthelmeh sounds good, pls assign me when ready for re-review

src/wh_server.c
JacobBarthelmeh: Also updated the if statement so that wh_Auth_Reset is always called, not just when the user.user_id is set. https://github.com/wolfSSL/wolfHSM/pull/383/changes#diff-a350c6f23e21d6aec09b776df6f81fe9c22ad4aab574292aefb99d0579acf76fR92-R93

src/wh_server.c
```diff
@@ -87,6 +87,26 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
...
+            rc = wh_Auth_Logout(server->auth, stale_id);
+            if (rc != WH_ERROR_OK) {
+                WH_LOG(&server->log, WH_LOG_LEVEL_SECEVENT,
+                   "Stale auth session force-cleared during server init after "
+                   "logout failure");
+            }
+        }
+        rc = wh_Auth_Reset(server->auth);
+        if (rc != WH_ERROR_OK) {
+            WH_LOG(&server->log, WH_LOG_LEVEL_SECEVENT,
```
bigbrett, line 124: @JacobBarthelmeh this look like the log will indeed be dropped since it is after the reset. Could you address this pls? Then we can get it in

JacobBarthelmeh:
@bigbrett addressed the logging placement feedback, reviewed a re-kick off of AI reviews and rebased onto master. I think it's ready for re-review, thanks for your patience on this PR.

bigbrett (approved):

PR #384, opened 2026-05-28
add SHA3 support

bigbrett:
Adds support for SHA3 to wolfHSM. Supports blocking wolfCrypt and async native API surface.


wolfhsm/wh_message_crypto.h
```diff
@@ -864,6 +872,116 @@ int wh_MessageCrypto_TranslateSha2Response(
...
+ * variant). */
+typedef struct {
+    whMessageCrypto_Sha3State resumeState;
+    uint8_t hash[64]; /* WC_SHA3_512_DIGEST_SIZE - max across variants */
+} whMessageCrypto_Sha3Response;
+
+/* Per-variant max-inline update sizes (block sizes differ across variants
+ * so each gets its own macro, rounded down to a whole-block multiple). */
+#define WH_MESSAGE_CRYPTO_SHA3_224_MAX_INLINE_UPDATE_SZ         \
+    (((WOLFHSM_CFG_COMM_DATA_LEN -                              \
```
bigbrett, line 930: fixed

src/wh_server_crypto.c
```diff
@@ -5279,6 +5437,110 @@ static int _HandleSha512Dma(whServerContext* ctx, uint16_t magic, int devId,
...
+        ret = ops.updateFn(sha3, inlineData, req.inSz);
+    }
+
+    if (ret == 0 && req.input.sz > 0) {
+        ret = wh_Server_DmaProcessClientAddress(
+            ctx, req.input.addr, &inAddr, req.input.sz,
+            WH_DMA_OPER_CLIENT_READ_PRE, (whServerDmaFlags){0});
+        if (ret == WH_ERROR_OK) {
+            preOk = 1;
+            ret   = ops.updateFn(sha3, inAddr, req.input.sz);
```
bigbrett, line 6323: maps to underlying wolfSSL args, no change necessary

wolfhsm/wh_message_crypto.h
```diff
@@ -875,6 +883,131 @@ int wh_MessageCrypto_TranslateSha2Response(
...
+ * Non-final updates: inSz must be a multiple of the variant's block
+ * size (144/136/104/72). The client buffers any partial-block tail
+ * locally in sha3->t[] and only sends it on Final with isLastBlock=1.
+ */
+
+/* SHA3 resume state - 200-byte Keccak state + buffer index, shared across
+ * SHA3-224/256/384/512. Layout chosen so the uint64_t array sits at the
+ * end (8-byte aligned). */
+typedef struct {
+    uint32_t i; /* reserved, always 0 — the partial-block buffer lives on
```
Frauschi: Nit: why having this `i` variable that is always zero and the 4 byte padding in the first place? Why not just removing that entirely? Then the structure would still be 8-byte aligned.

src/wh_client_crypto.c
```diff
@@ -8416,6 +8416,750 @@ int wh_Client_Sha512Dma(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
...
+    }
+
+    capacity = _Sha3UpdatePerCallCapacity(sha, v);
+    if (inLen > capacity) {
+        return WH_ERROR_BADARGS;
+    }
+    if (inLen == 0) {
+        return WH_ERROR_OK;
+    }
+
```
Frauschi, line 8517: `if (wh_CommClient_IsRequestPending(ctx->comm) == 1)` check is missing from all non-DMA request methods.

Frauschi (changes requested):
Only minor issues, otherwise LGTM.

Just one question: is SHAKE128/SHAKE256 missing on purpose?

bigbrett:
@Frauschi didn't even think about SHAKE tbh. I can add in a subsequent PR. This was largely driven by a customer request for SHA3 who never mentioned Shake

Frauschi:
@bigbrett Ah ok, I just asked because SHAKE is the one required for the PQC algos we talked about a few weeks ago.

src/wh_client_crypto.c
bigbrett: We don't really need these in the non-DMA handlers, as they were introduced to prevent mutation of state from the client-side DMA callbacks. Since non-DMA doesn't have any mutateable hw state then its fine to just run until it fails

wolfhsm/wh_message_crypto.h
bigbrett: oops that wasnt padding, it was meant to be a buffer index in a previous impl that got scrubbed but wasn't removed properly. Fixing.

Frauschi (dismissed):
LGTM

src/wh_client_crypto.c
```diff
@@ -8416,6 +8416,738 @@ int wh_Client_Sha512Dma(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
...
+            sha->t[sha->i++] = in[i++];
+        }
+        if (sha->i == v->blockSize) {
+            memcpy(inlineData + wirePos, sha->t, v->blockSize);
+            wirePos += v->blockSize;
+            sha->i = 0;
+        }
+    }
+
+    /* Pack as many whole blocks from input as fit inline. */
```
AlexLanzano: Nit: Fix grammar?

src/wh_client_crypto.c
```diff
@@ -8416,6 +8416,738 @@ int wh_Client_Sha512Dma(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
...
+    inlineData = (uint8_t*)(req + 1);
+
+    savedI = sha->i;
+    memcpy(savedT, sha->t, sha->i);
+
+    /* Top up the local partial buffer. If it completes a full block, copy
+     * the assembled block as the first inline block. */
+    if (sha->i > 0) {
+        while (i < inLen && sha->i < v->blockSize) {
+            sha->t[sha->i++] = in[i++];
```
AlexLanzano, line 8534: Maybe we do an additional bounds check prior to this to see that:
`v->blockSize < MIN(sizeof(sha->t), sizeof(in))`

src/wh_client_crypto.c
```diff
@@ -8416,6 +8416,738 @@ int wh_Client_Sha512Dma(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
...
+
+    ret = _getCryptoResponse(dataPtr, v->hashType, (uint8_t**)&res);
+    if (ret >= 0) {
+        if (dataSz <
+            sizeof(whMessageCrypto_GenericResponseHeader) + sizeof(*res)) {
+            return WH_ERROR_ABORTED;
+        }
+        for (k = 0; k < 25; k++) {
+            sha->s[k] = res->resumeState.s[k];
+        }
```
AlexLanzano: Nit: use memcpy instead? Same for other instances

src/wh_client_crypto.c
```diff
@@ -8416,6 +8416,738 @@ int wh_Client_Sha512Dma(whClientContext* ctx, wc_Sha512* sha, const uint8_t* in,
...
+WH_SHA3_VARIANT_API(256)
+#endif
+#ifndef WOLFSSL_NOSHA3_384
+WH_SHA3_VARIANT_API(384)
+#endif
+#ifndef WOLFSSL_NOSHA3_512
+WH_SHA3_VARIANT_API(512)
+#endif
+
+#undef WH_SHA3_VARIANT_API
```
AlexLanzano: Not the biggest fan of this. Would rather just lay these functions out explicitly to keep them grepable and easy to jump to with LSPs or other debugging tools.

src/wh_client_crypto.c
bigbrett: The block size isn't something that can vary at runtime, its fixed at compile time based on the algo so it can't really ever overflow. Was gonna add a static assert but the condition got weird and crazy verbose so I think its okay as is?

src/wh_client_crypto.c
bigbrett: fixed

src/wh_client_crypto.c
bigbrett: fixed

src/wh_client_crypto.c
bigbrett: fixed

src/wh_client_crypto.c
AlexLanzano: Yeah, it just makes me uneasy that we don't bounds check on the actual size of `t`. But if it's guaranteed that `blockSize` can't be larger than `t` or `in` then I'm all good.

AlexLanzano (approved):

PR #385, opened 2026-05-28
ci gcov fix

bigbrett:
Upgrades gcov to fix CI error after runner upgrade


padelsbach (approved):
lgtm, thanks for the fix!

PR #386, opened 2026-05-28
add restriction on users ability to set credentials

JacobBarthelmeh:
Follow up to item 3 from https://github.com/wolfSSL/wolfHSM/pull/270


rizlik (approved):
Nice work.
As a small improvement in the AuthBase backend that can be postponed to a followup PR: consider to allow the admin to overwrite other user credentials without knowledge of the credentials itself

bigbrett (approved):

PR #388, opened 2026-05-29
Add data len check in wh_CommClient_RecvResponse

padelsbach:
`wh_CommClient_RecvResponse` does a `memcpy` into a caller provided buf without checking the length -- the function does not contain a length param, however the calling functions do. 

This change adds the `data_size` param, so this change does break the API. Alternative was to use existing `uint16_t* out_size` as an `inout` param, but that felt too sneaky and easy to overlook.

All 130ish callers updated. `src-jp` references the params (and is updated), but the English `src` does not.

### DIFF
--- a/src/wh_comm.c
+++ b/src/wh_comm.c
@@ -156,9 +156,10 @@ int wh_CommClient_SendRequest(whCommClient* context, uint16_t magic,
  * *out_size set to the required size. On success *out_size holds the actual
  * payload size.
  */
-int wh_CommClient_RecvResponse(whCommClient* context,
-        uint16_t* out_magic, uint16_t* out_kind, uint16_t* out_seq,
-        uint16_t* out_size, uint16_t data_size, void* data)
+int wh_CommClient_RecvResponse(whCommClient* context, uint16_t* out_magic,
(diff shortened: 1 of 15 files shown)


src/wh_comm.c
```diff
@@ -221,15 +224,20 @@ int wh_CommClient_RecvResponse(whCommClient* context,
                 return WH_ERROR_NOTREADY;
             }
 
-            if (    (data != NULL) &&
-                    (data_size != 0) &&
-                    (data != context->data)) {
-                memcpy(data, context->data, data_size);
+            if ((data != NULL) &&
```
padelsbach: This is by design

AlexLanzano:
@padelsbach Please fix the merge conflicts when you can. Thanks

AlexLanzano (dismissed):

AlexLanzano:
@bigbrett Passing this off to you since this is going to technically break the public API but I don't believe that many (if any) customer's directly call the function in question

bigbrett (changes requested):
@padelsbach looks great, but this will need to be rebased and then you will see CI fail to show you where new callsites have to be modified (all in the SHE layer)

bigbrett (approved):

PR #389, opened 2026-05-29
Add data len check in wh_CommServer_RecvRequest

padelsbach:
Server side companion fix to #388, with similar new length check.

### DIFF
--- a/src/wh_comm.c
+++ b/src/wh_comm.c
@@ -350,14 +350,14 @@ int wh_CommServer_Init(whCommServer* context, const whCommServerConfig* config,
 
 int wh_CommServer_RecvRequest(whCommServer* context,
         uint16_t* out_magic, uint16_t* out_kind, uint16_t* out_seq,
-        uint16_t* out_size, void* data)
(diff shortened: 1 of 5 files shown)


src/wh_comm.c
```diff
@@ -373,21 +373,26 @@ int wh_CommServer_RecvRequest(whCommServer* context,
...
         }
         if (rc == 0) {
-            data_size = size - sizeof(*context->hdr);
+            req_size = size - sizeof(*context->hdr);
             magic = context->hdr->magic;
             kind = wh_Translate16(magic, context->hdr->kind);
             seq = wh_Translate16(magic, context->hdr->seq);
 
             /* Copy the data from the internal buffer if necessary */
             if (    (data != NULL) &&
```
padelsbach: fixed

AlexLanzano (approved):

bigbrett (approved):

bigbrett (approved):

bigbrett:
@padelsbach can we get a negative test to ensure that the error check for overflow holds? Currently we only have test coverage of known good scenarios that will never actually trigger an error.

PR #391, opened 2026-05-29
Add checking in SHE response handlers

padelsbach:
Check for group, action and len in SHE response handlers. Found with experimental fuzzing.

### DIFF
--- a/src/wh_client_she.c
+++ b/src/wh_client_she.c
@@ -96,12 +96,8 @@ int wh_Client_SheSetUidResponse(whClientContext* c)
 
     resp = (whMessageShe_SetUidResponse*)wh_CommClient_GetDataPtr(c->comm);
     ret = wh_Client_RecvResponse(c, &group, &action, &dataSz, (uint8_t*)resp);
-    if (ret == WH_ERROR_OK) {
-        if (group != WH_MESSAGE_GROUP_SHE ||
-            action != WH_SHE_SET_UID ||
-            dataSz < sizeof(*resp)) {
-            ret = WH_ERROR_ABORTED;
-        }
+    if (ret == WH_ERROR_OK && dataSz < sizeof(*resp)) {
+        ret = WH_ERROR_ABORTED;
     }
     if (ret == WH_ERROR_OK) {
         ret = resp->rc;
@@ -157,12 +153,8 @@ int wh_Client_SheSecureBoot(whClientContext* c, uint8_t* bootloader,
             initResp = (whMessageShe_SecureBootInitResponse*)respBuf;
         } while (ret == WH_ERROR_NOTREADY);
     }
-    if (ret == 0) {
-        if (group != WH_MESSAGE_GROUP_SHE ||
-            action != WH_SHE_SECURE_BOOT_INIT ||
-            dataSz < sizeof(*initResp)) {
-            ret = WH_ERROR_ABORTED;
-        }
+    if (ret == 0 && dataSz < sizeof(*initResp)) {
+        ret = WH_ERROR_ABORTED;
     }
 
     /* send update sub command until we've sent the entire bootloader */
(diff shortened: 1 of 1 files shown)


src/wh_client_she.c
```diff
@@ -96,6 +96,13 @@ int wh_Client_SheSetUidResponse(whClientContext* c)
 
     resp = (whMessageShe_SetUidResponse*)wh_CommClient_GetDataPtr(c->comm);
     ret = wh_Client_RecvResponse(c, &group, &action, &dataSz, (uint8_t*)resp);
+    if (ret == WH_ERROR_OK) {
```
padelsbach: Could macro'ize this, but not sure if that's more readable

src/wh_client_she.c
```diff
@@ -640,11 +712,19 @@ int wh_Client_SheEncEcbResponse(whClientContext* c, uint8_t* out, uint32_t sz)
 
     ret = wh_Client_RecvResponse(c, &group, &action, &dataSz, (uint8_t*)resp);
     if (ret == 0) {
-        if (resp->rc != WH_SHE_ERC_NO_ERROR) {
-            ret = resp->rc;
+        if (group != WH_MESSAGE_GROUP_SHE ||
+            action != WH_SHE_ENC_ECB ||
+            dataSz < sizeof(*resp) ||
+            dataSz < sizeof(*resp) + resp->sz) {
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_client_she.c
```diff
@@ -96,6 +96,13 @@ int wh_Client_SheSetUidResponse(whClientContext* c)
 
     resp = (whMessageShe_SetUidResponse*)wh_CommClient_GetDataPtr(c->comm);
     ret = wh_Client_RecvResponse(c, &group, &action, &dataSz, (uint8_t*)resp);
+    if (ret == WH_ERROR_OK) {
+        if (group != WH_MESSAGE_GROUP_SHE ||
+            action != WH_SHE_SET_UID ||
```
bigbrett: all these group and action checks are redundant, as they are handled in the comm layer. Otherwise every single client response check would need these.

Also, do you think the size check could be pushed down to the comm layer somehow in a uniform way, perhaps via same mechanism as #389 and #388? Currently these would need to be done inline everywhere.

Note we dont care THAT much about this fuzzing style stuff - currently the threat model is such that transports are trusted. If someone can modify data in your transport, all bets are off. So temped to just not do the infinite bikeshedding around input sanitation that fenrir keeps pointing out....

bigbrett (changes requested):

src/wh_client_she.c
padelsbach: Fair. Updated the PR to add only the `dataSz < sizeof(*resp)` which seems legit useful

padelsbach:
Passing over to @AlexLanzano in Brett's absence

AlexLanzano (approved):

AlexLanzano (changes requested):
Hmm, Actually on a second look these checks still wouldn't prevent an overflow since the data is copied prior to these check. This needs to be handled at the comm layer probably by adding a dataLen arg to the wh_Client_RecvResponse and propagating that to the comm layer.

padelsbach:
> Hmm, Actually on a second look these checks still wouldn't prevent an overflow since the data is copied prior to these check. This needs to be handled at the comm layer probably by adding a dataLen arg to the wh_Client_RecvResponse and propagating that to the comm layer.

I think you're referring to this copy down in `wh_CommClient_RecvResponse`:

```
            if (    (data != NULL) &&
                    (data_size != 0) &&
                    (data != context->data)) {
                memcpy(data, context->data, data_size);
            }
```

The vast majority of the handlers provide the `data` buffer matching `context->data`, so the `data != context->data` check prevents this memcpy.

In any case, the checks added here are `recv < expected` so they wouldn't catch an overflow anyway. They're mainly to prevent a *too small* buffer from being over indexed.

But...

I see your point that this could be done in a more uniform way, perhaps by passing `sizeof(*resp)` into `wh_Client_RecvResponse`. I'm into it if we don't care about breaking the public API.

AlexLanzano:
> I see your point that this could be done in a more uniform way, perhaps by passing sizeof(*resp) into wh_Client_RecvResponse. I'm into it if we don't care about breaking the public API.

Yeah, I think in this case it's fine to break the API. I don't believe that many (if any) users are calling the comm client send/recv functions directly in their applications

padelsbach:
I've implemented the common check in #453 as a separate PR so this one can be kept for reference. Closing

PR #398, opened 2026-06-03
Add refactored auth tests

padelsbach:
Notes:

- When auth is enabled, the posix test now logs in as admin before _any_ tests run so the server allows all messages (see preexisting `if (server->auth != NULL)` check in 
`wh_Server_HandleRequestMessage`)
- Auth tests mostly copied verbatim, except:
  - no separate setup/teardown of server in each test, and instead:
  - at start of each auth test, logout to establish baseline
  - at end of each auth test, log out of any test users, and re-login as admin
- Fix in legacy code: free `testuser4` which was silently leaked


AlexLanzano (dismissed):
Looks solid! One thing to note, the legacy test does provide an additional test that I dont see here that verified that the specified auth cb function is actually called. Not sure if this should be added here too or an overkill?

padelsbach:
> One thing to note, the legacy test does provide an additional test that I dont see here that verified that the specified auth cb function is actually called.

Good catch. I added a small check to mirror the legacy one. And rebased to resolve conflicts.

AlexLanzano (approved):

PR #401, opened 2026-06-03
Set committed=1 on cached keys read back from NVM

padelsbach:
When reading keys from NVM into the cache, they are marked as `committed=0` and thus cannot be evicted, and could eventually cause key readback to fail. 

This change sets `committed=1` if the key is read from NVM. New test catches the bug.


billphipps:
Can you see if this fix corrects ZD21666?  User provided a set of use cases that certainly exercise this eviction problem, but I was not convinced this fix alone was sufficient.

AlexLanzano (approved):

PR #403, opened 2026-06-05
Fix client *Dma address translation (KeyExport / NvmAdd / NvmRead) and KeyCacheDma use-after-free

Frauschi:
### Background

 A few client *Dma APIs wrote the raw client pointer straight into the wire message instead of routing it through the DMA address-translation callback like the rest of the family.

This is invisible on flat-memory ports (the POSIX SHM reference, where client and server share an address space), so it survived CI. It breaks on a real split-address-space port — found during DMA bring-up on one of the new hardware platforms for wolfHSM, where the server runs on a separate security CPU with no view of host RAM, so it bounds-checks the untranslated pointer and rejects the request (`WH_ERROR_BADARGS`).

### What this fixes

- Missing translation in `KeyExportDma`, `KeyExportPublicDma`, `NvmAddObjectDma`, `NvmReadDma`. Each now runs `*_PRE `translation before sending and the matching `*_POST` on the response, mirroring `RngGenerateDma`.
- Use-after-free in `KeyCacheDmaRequest`, which ran its `POST` inside the request — releasing/recycling the scratch before the server read the buffer (benign on POSIX, a real UAF on any port whose callback allocates). It now defers the `POST` to `KeyCacheDmaResponse`.

### Notable implementation details

- The per-call-site async-DMA structs are consolidated into a single `whClientDmaAsyncBuf` (carrying the `POST` direction) plus a two-buffer `nvmAdd` variant, with a shared `wh_Client_DmaAsyncPost()` helper that returns the `POST` status (a failed unmap/copy-back is surfaced, not swallowed).
- Each request clears its mapping slot(s) before the `PRE`, so a path that leaves a slot unpopulated (e.g. metadata-only `NvmAddObjectDma`) can't make the response `POST` stale union memory; `data_hostaddr` is 0 when there is no data buffer, so a raw client pointer is never forwarded.
- Requests fail fast with `WH_ERROR_REQUEST_PENDING` on a busy transport (no mapping acquired then leaked).
- `wh_{Client,Server}_DmaRegisterAllowList()` now accept `NULL` to unregister, symmetric with `DmaRegisterCb(NULL)`.

### Tests

The reason these bugs survived CI is that no POSIX test ever registered a non-identity DMA callback — a missing translation looked identical to a correct call. This PR adds a shared bounce-pool harness (`test/wh_test_dma.c`) that models a split-address-space port: the client callback bounces each buffer through a dedicated pool and hands the server a pool address; the server callback rejects any untranslated address. Freed slots are poisoned (so a premature `POST` corrupts the data) and a `POST` matching no live slot is recorded as a stray/double free.

- `wh_test_clientserver.c` drives the keystore/NVM `*Dma` APIs end-to-end — including metadata-only and injected `PRE`-failure cases — in the single-thread pump harness, where ordering makes the use-after-free deterministic.
- `wh_test_crypto.c` registers the same callback on the threaded harness, so the entire crypto/cert `*Dma` suite (AES/SHA/CMAC/RNG/Ed25519/ML-DSA/ML-KEM/cert, plus KeyExportPublicDma) runs through translation automatically.
- Also fixes a pre-existing bug the new coverage surfaced: `_testDma` left a stack-local allow list registered after returning (a dangling pointer ASAN flags once a later test uses server DMA).

Runs under `make DMA=1` (already set by every CI job).

### Validation

- `make DMA=1 ASAN=1`: passes, 0 ASAN findings.
- Default non-DMA build: passes.
- Each fix was confirmed to be caught by the new tests (regressions re-introduced → tests fail).

### DIFF
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -1430,11 +1430,10 @@ int wh_Client_KeyCacheDmaRequest(whClientContext* c, uint32_t flags,
                                  const void* keyAddr, uint16_t keySz,
                                  uint16_t keyId)
 {
-    int                                ret             = WH_ERROR_OK;
-    whMessageKeystore_CacheDmaRequest* req             = NULL;
-    uintptr_t                          keyAddrPtr      = 0;
-    uint16_t                           capSz           = 0;
-    int                                keyAddrAcquired = 0;
+    int                                ret        = WH_ERROR_OK;
+    whMessageKeystore_CacheDmaRequest* req        = NULL;
+    uintptr_t                          keyAddrPtr = 0;
+    uint16_t                           capSz      = 0;
 
     if (c == NULL || (labelSz > 0 && label == NULL)) {
         return WH_ERROR_BADARGS;
@@ -1449,11 +1448,6 @@ int wh_Client_KeyCacheDmaRequest(whClientContext* c, uint32_t flags,
         return WH_ERROR_BADARGS;
     }
     memset(req, 0, sizeof(*req));
-    req->id       = keyId;
-    req->flags    = flags;
-    req->labelSz  = 0;
-    req->key.sz   = keySz;
-    req->key.addr = 0;
 
     /* Clear the slot up front so a skipped PRE leaves nothing for POST. */
     c->dma.asyncCtx.buf.sz = 0;
@@ -1465,28 +1459,29 @@ int wh_Client_KeyCacheDmaRequest(whClientContext* c, uint32_t flags,
         c, (uintptr_t)keyAddr, (void**)&keyAddrPtr, keySz,
         WH_DMA_OPER_CLIENT_READ_PRE, (whDmaFlags){0});
     if (ret == WH_ERROR_OK) {
-        keyAddrAcquired                 = 1;
-        req->key.addr                   = (uint64_t)keyAddrPtr;
         c->dma.asyncCtx.buf.xformedAddr = keyAddrPtr;
         c->dma.asyncCtx.buf.clientAddr  = (uintptr_t)keyAddr;
         c->dma.asyncCtx.buf.sz          = keySz;
         c->dma.asyncCtx.buf.postOper    = WH_DMA_OPER_CLIENT_READ_POST;
-    }
 
-    /* Copy label if provided, truncate if necessary */
-    if (labelSz > 0 && label != NULL) {
-        capSz = (labelSz > WH_NVM_LABEL_LEN) ? WH_NVM_LABEL_LEN : labelSz;
-        req->labelSz = capSz;
-        memcpy(req->label, label, capSz);
-    }
+        /* Build and send the request now that the buffer is mapped. */
+        req->id       = keyId;
+        req->flags    = flags;
+        req->key.addr = (uint64_t)keyAddrPtr;
+        req->key.sz   = keySz;
+        if (labelSz > 0 && label != NULL) {
+            capSz = (labelSz > WH_NVM_LABEL_LEN) ? WH_NVM_LABEL_LEN : labelSz;
+            req->labelSz = capSz;
+            memcpy(req->label, label, capSz);
+        }
 
-    if (ret == WH_ERROR_OK) {
         ret = wh_Client_SendRequest(c, WH_MESSAGE_GROUP_KEY, WH_KEY_CACHE_DMA,
                                     sizeof(*req), (uint8_t*)req);
     }
 
-    if (ret != WH_ERROR_OK && keyAddrAcquired) {
-        /* SendRequest failed: the Response will not run, so POST now. */
+    /* On any failure release the mapping; POST no-ops on the unset slot, so a
+     * failed PRE needs no separate guard. */
+    if (ret != WH_ERROR_OK) {
         (void)wh_Client_DmaAsyncPost(c, &c->dma.asyncCtx.buf);
     }
     return ret;
@@ -1533,14 +1528,11 @@ int wh_Client_KeyCacheDmaResponse(whClientContext* c, uint16_t* keyId)
         }
     }
 
-    /* POST cleanup: release the mapping once the server has read it. Surface a
-     * POST failure if the operation otherwise succeeded. */
-    {
-        int postRc = wh_Client_DmaAsyncPost(c, &c->dma.asyncCtx.buf);
-        if (ret == WH_ERROR_OK) {
-            ret = postRc;
-        }
-    }
+    /* POST cleanup: release the input mapping the server has finished reading.
+     * The key is already cached server-side, so failing to release the
+     * client-side scratch is a cleanup issue, not an operation failure; don't
+     * override a successful result with it. */
+    (void)wh_Client_DmaAsyncPost(c, &c->dma.asyncCtx.buf);
     return ret;
 }
 
@@ -1562,10 +1554,9 @@ int wh_Client_KeyCacheDma(whClientContext* c, uint32_t flags, uint8_t* label,
 int wh_Client_KeyExportDmaRequest(whClientContext* c, uint16_t keyId,
                                   const void* keyAddr, uint16_t keySz)
 {
-    whMessageKeystore_ExportDmaRequest* req             = NULL;
-    uintptr_t                           keyAddrPtr      = 0;
-    int                                 ret             = WH_ERROR_OK;
-    int                                 keyAddrAcquired = 0;
+    whMessageKeystore_ExportDmaRequest* req        = NULL;
+    uintptr_t                           keyAddrPtr = 0;
+    int                                 ret        = WH_ERROR_OK;
 
     if (c == NULL || keyId == WH_KEYID_ERASED) {
         return WH_ERROR_BADARGS;
@@ -1581,10 +1572,6 @@ int wh_Client_KeyExportDmaRequest(whClientContext* c, uint16_t keyId,
         return WH_ERROR_BADARGS;
     }
 
-    req->id       = keyId;
-    req->key.addr = 0;
-    req->key.sz   = keySz;
-
     /* Clear the slot up front so a skipped PRE leaves nothing for POST. */
     c->dma.asyncCtx.buf.sz = 0;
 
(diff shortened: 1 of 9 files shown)


src/wh_client.c
```diff
@@ -1430,30 +1430,48 @@ int wh_Client_KeyCacheDmaRequest(whClientContext* c, uint32_t flags,
...
-    req->flags   = flags;
-    req->labelSz = 0;
-
-    /* Set up DMA buffer info */
+    req->id       = keyId;
+    req->flags    = flags;
+    req->labelSz  = 0;
     req->key.sz   = keySz;
-    ret           = wh_Client_DmaProcessClientAddress(
+    req->key.addr = 0;
```
rizlik: we can probably move this *after* `wh_Client_DmaProcessClientAddress`

src/wh_client.c
rizlik: The label copying should not happen on `ret != WH_ERROR_OK`

src/wh_client.c
```diff
@@ -1467,9 +1485,10 @@ int wh_Client_KeyCacheDmaRequest(whClientContext* c, uint32_t flags,
                                     sizeof(*req), (uint8_t*)req);
     }
 
-    (void)wh_Client_DmaProcessClientAddress(
-        c, (uintptr_t)keyAddr, (void**)&keyAddrPtr, keySz,
-        WH_DMA_OPER_CLIENT_READ_POST, (whDmaFlags){0});
+    if (ret != WH_ERROR_OK && keyAddrAcquired) {
```
rizlik: sometime the PR run always `wh_Client_DmaAsyncPost` and rely on the `sz == 0` to skip the operation, while sometime it rely on boolean like `keyAddrAcquired`. Consider dropping the boolean or using just a single approach.

src/wh_client.c
```diff
@@ -1531,23 +1562,56 @@ int wh_Client_KeyCacheDma(whClientContext* c, uint32_t flags, uint8_t* label,
...
     req =
         (whMessageKeystore_ExportDmaRequest*)wh_CommClient_GetDataPtr(c->comm);
     if (req == NULL) {
         return WH_ERROR_BADARGS;
     }
+
     req->id       = keyId;
-    req->key.addr = (uint64_t)((uintptr_t)keyAddr);
+    req->key.addr = 0;
     req->key.sz   = keySz;
```
rizlik: this can be moved after DmaProcessClientAddress

src/wh_client.c
```diff
@@ -1531,23 +1562,56 @@ int wh_Client_KeyCacheDma(whClientContext* c, uint32_t flags, uint8_t* label,
...
+
+    if (ret == WH_ERROR_OK) {
+        ret = wh_Client_SendRequest(c, WH_MESSAGE_GROUP_KEY, WH_KEY_EXPORT_DMA,
+                                    sizeof(*req), (uint8_t*)req);
+    }
+
+    if (ret != WH_ERROR_OK && keyAddrAcquired) {
+        /* SendRequest failed: the Response will not run, so POST now. */
+        (void)wh_Client_DmaAsyncPost(c, &c->dma.asyncCtx.buf);
+    }
```
rizlik: same comment of above

src/wh_client.c
```diff
@@ -1595,6 +1662,15 @@ int wh_Client_KeyExportDmaResponse(whClientContext* c, uint8_t* label,
...
         }
     }
+
+    /* POST cleanup: copy results back and release the mapping; surface a POST
+     * failure if the operation otherwise succeeded. */
+    {
+        int postRc = wh_Client_DmaAsyncPost(c, &c->dma.asyncCtx.buf);
+        if (rc == WH_ERROR_OK) {
+            rc = postRc;
+        }
```
rizlik, line 1650: I'm not really convinced of surfacing the POST errors. 
Hiding an error is bad, but reporting a successful operation as failed has also its own implications (will the user try to retry the operation? how the user can know the state of cached key from the error?).

src/wh_client.c
```diff
@@ -1616,26 +1692,57 @@ int wh_Client_KeyExportPublicDmaRequest(whClientContext* c, whKeyId keyId,
...
             c->comm);
     if (req == NULL) {
         return WH_ERROR_BADARGS;
     }
+
     req->id       = keyId;
     req->algo     = algo;
-    req->key.addr = (uint64_t)((uintptr_t)keyAddr);
+    req->key.addr = 0;
     req->key.sz   = keySz;
```
rizlik: move after process client

src/wh_client_nvm.c
```diff
@@ -766,19 +827,53 @@ int wh_Client_NvmReadDmaRequest(whClientContext* c, whNvmId id,
...
-                                 &msg);
+    msg.data_hostaddr = (uint64_t)dataAddrPtr;
+
+    if (ret == WH_ERROR_OK) {
+        ret = wh_Client_SendRequest(c, WH_MESSAGE_GROUP_NVM,
+                                    WH_MESSAGE_NVM_ACTION_READDMA, sizeof(msg),
+                                    &msg);
+    }
+
+    if (ret != WH_ERROR_OK && dataAddrAcquired) {
```
rizlik: we can drop the boolean if we rely on the `sz == 0` check.

src/wh_client_nvm.c
```diff
@@ -766,19 +827,53 @@ int wh_Client_NvmReadDmaRequest(whClientContext* c, whNvmId id,
...
 
     msg.id            = id;
     msg.offset        = offset;
     msg.data_len      = data_len;
-    msg.data_hostaddr = (uint64_t)(uintptr_t)data;
-    return wh_Client_SendRequest(c, WH_MESSAGE_GROUP_NVM,
-                                 WH_MESSAGE_NVM_ACTION_READDMA, sizeof(msg),
-                                 &msg);
+    msg.data_hostaddr = (uint64_t)dataAddrPtr;
+
```
rizlik: guard inside a `ret == WH_ERROR_OK` if block or move in the next one.

rizlik (changes requested):

src/wh_client.c
Frauschi: done

src/wh_client.c
Frauschi: fixed

src/wh_client.c
Frauschi: fixed

src/wh_client.c
Frauschi: done

src/wh_client.c
Frauschi: done

src/wh_client_nvm.c
Frauschi: fixed

src/wh_client_nvm.c
Frauschi: fixed

src/wh_client.c
Frauschi: On operations where the server reads from the buffer, a POST error is now "hidden", as this is only cleanup and the actual operation on the server succeeded.
On operations where the client receives actual data from the server, a failing POST still fails the entire operation, as otherwise the client cannot be sure the data is valid.

src/wh_client.c
Frauschi: addressed.

rizlik (dismissed):
Thanks, lgtm!

src/wh_client.c
```diff
@@ -1430,46 +1430,60 @@ int wh_Client_KeyCacheDmaRequest(whClientContext* c, uint32_t flags,
...
-        capSz = (labelSz > WH_NVM_LABEL_LEN) ? WH_NVM_LABEL_LEN : labelSz;
-        req->labelSz = capSz;
-        memcpy(req->label, label, capSz);
-    }
-
     if (ret == WH_ERROR_OK) {
+        c->dma.asyncCtx.buf.xformedAddr = keyAddrPtr;
+        c->dma.asyncCtx.buf.clientAddr  = (uintptr_t)keyAddr;
+        c->dma.asyncCtx.buf.sz          = keySz;
+        c->dma.asyncCtx.buf.postOper    = WH_DMA_OPER_CLIENT_READ_POST;
```
bigbrett: This seems really clunky - should we not have a generic pre- wrapper like you added for the post operation that primes the asyncCtx under the hood based on the input args?

src/wh_client.c
```diff
@@ -1492,6 +1506,9 @@ int wh_Client_KeyCacheDmaResponse(whClientContext* c, uint16_t* keyId)
     }
 
     ret = wh_Client_RecvResponse(c, &group, &action, &size, (uint8_t*)resp);
+    if (ret == WH_ERROR_NOTREADY) {
```
bigbrett, line 1503: it would be helpful to have a comment about why NOTREADY gets an early return but other errors do not (releasing the DMA mapping)

src/wh_client.c
bigbrett: @Frauschi this would be a good code comment to have, as it is very much not obvious why some call sites for the post are void casted, and others are optionally error-propagated. Reading the code I had the same concerns as marco.

src/wh_client_dma.c
```diff
@@ -94,4 +96,23 @@ int wh_Client_DmaProcessClientAddress(whClientContext* client,
...
+        return WH_ERROR_OK;
+    }
+
+    addr = buf->xformedAddr;
+    rc   = wh_Client_DmaProcessClientAddress(client, buf->clientAddr,
+                                             (void**)&addr, (size_t)buf->sz,
+                                             buf->postOper, (whDmaFlags){0});
+    /* Clear the slot even on failure so a later Response cannot re-run the
+     * POST; the failure is returned to the caller. */
+    buf->sz = 0;
```
bigbrett: part of me thinks we should just clear the whole struct since it is no longer in use? Makes it seem more explicit.

src/wh_client_nvm.c
```diff
@@ -695,19 +695,67 @@ int wh_Client_NvmAddObjectDmaRequest(whClientContext* c,
...
+    /* Fail fast if busy: don't acquire a mapping a rejected send would leak. */
+    if (wh_CommClient_IsRequestPending(c->comm) == 1) {
+        return WH_ERROR_REQUEST_PENDING;
+    }
+
+    /* Clear both slots up front: a metadata-only object leaves the data slot
+     * unset, and the Response must not POST a stale (shared-union) size. */
+    c->dma.asyncCtx.nvmAdd.meta.sz = 0;
+    c->dma.asyncCtx.nvmAdd.data.sz = 0;
+
```
bigbrett, line 715: same comment as above re- clearing the whole struct. I wonder if this should have a helper just to make it explicit (1wh_Client_DmaAsyncCtxReset()`) or "clear" or  something like that?

bigbrett (dismissed):
@Frauschi I'm approving in light of my comments earlier. If you want it in ASAP you can go ahead and click merge since both Marco and I approved, though I'd like the comments addressed at some point. If you would rather do it in a follow-up PR, go ahead and merge. If you want to follow up with another PR on this branch, then you are welcome to do that too. Up to you, both fine with me.

src/wh_client.c
Frauschi: Fixed by adding the new `wh_Client_DmaAsyncPre()` helper analogous to the post helper.

src/wh_client.c
Frauschi: Added an explanatory comment (on all locations)

src/wh_client_dma.c
Frauschi: fixed

src/wh_client_nvm.c
Frauschi: Cleared now completely within the post helper.

Frauschi:
@bigbrett I addressed your comments right away. The new pre and post helper are currently only applied to the *DMA methods touched in this PR. I'll do a follow-up PR that migrates all the other *Dma methods (mostly the many crypto operations) to these as well.

bigbrett (approved):

PR #406, opened 2026-06-11
Devid per client

bigbrett:
## Overview:
This PR gives each client its own device ID, so you can run several clients in one process. You set a client's ID in its config (leave it 0 to get the default), and pass WH_CLIENT_DEVID(c) or the raw devId itself to wolfCrypt. The old shared IDs still work if you only have one client. DMA is now a simple per-client on/off switch instead of a separate device ID.

## Changes:

- Each client gets its own device ID, set in its config (0 means use the default). Read it back with WH_CLIENT_DEVID(c).
- The old shared IDs (WH_DEV_ID, WH_DEV_ID_DMA) still work, but only with one client per process. You can now override their values at compile time.
- Turn DMA on or off per client with wh_Client_SetDmaMode(). When it's on, calls that can't use DMA fall back to the normal path automatically.
- Init and cleanup are safer with multiple clients: bad client IDs are rejected up front, a failed init undoes only its own work, and one client shutting down no longer affects the others.
- Updated tests, benchmarks, examples, and docs to match the new devId scheme. The legacy global devIds are exercised via the wolfCrypt test suite.

### DIFF
--- a/src/wh_client.c
+++ b/src/wh_client.c
@@ -59,23 +59,48 @@
 #include "wolfhsm/wh_message_counter.h"
 #include "wolfhsm/wh_client.h"
 
-#ifndef WOLFHSM_CFG_NO_CRYPTO
-const int WH_DEV_IDS_ARRAY[WH_NUM_DEVIDS] = {
-        WH_DEV_ID,
-#ifdef WOLFHSM_CFG_DMA
-        WH_DEV_ID_DMA,
-#endif /* WOLFHSM_CFG_DMA */
-};
-#endif /* WOLFHSM_CFG_NO_CRYPTO */
-
 int wh_Client_Init(whClientContext* c, const whClientConfig* config)
 {
     int rc = 0;
-    if((c == NULL) || (config == NULL)) {
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+    /* Which cryptoCb registrations this Init made, so the failure path can
+     * undo just those and leave other clients' entries alone. */
+    int clientCbRegistered = 0;
+    int globalCbRegistered = 0;
+#ifdef WOLFHSM_CFG_DMA
+    int dmaCbRegistered = 0;
+#endif /* WOLFHSM_CFG_DMA */
+#endif /* !WOLFHSM_CFG_NO_CRYPTO */
+
+    /* Client id must be 1..WH_CLIENT_ID_MAX (the server checks it at connect).
+     */
+    if ((c == NULL) || (config == NULL) || (config->comm == NULL) ||
+        (config->comm->client_id == 0) ||
+        (config->comm->client_id > WH_CLIENT_ID_MAX)) {
         return WH_ERROR_BADARGS;
     }
 
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+    /* devId 0 means "use the default WH_DEV_ID"; any other value must be
+     * positive. */
+    if (config->devId < 0) {
+        return WH_ERROR_BADARGS;
+    }
+#ifdef WOLFHSM_CFG_DMA
+    /* WH_DEV_ID_DMA is reserved for the DMA-only callback, so reject it. */
+    if (config->devId == WH_DEV_ID_DMA) {
+        return WH_ERROR_BADARGS;
+    }
+#endif /* WOLFHSM_CFG_DMA */
+#endif /* !WOLFHSM_CFG_NO_CRYPTO */
+
     memset(c, 0, sizeof(*c));
+
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+    /* Store the devId for this context. A nonzero devId also means "init
+     * succeeded"; the failure path below sets it back to 0 so Cleanup knows. */
+    c->devId = (config->devId == 0) ? WH_DEV_ID : config->devId;
+#endif /* !WOLFHSM_CFG_NO_CRYPTO */
 
     rc = wh_CommClient_Init(c->comm, config->comm);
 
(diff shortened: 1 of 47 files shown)


src/wh_client.c
```diff
@@ -126,6 +212,21 @@ int wh_Client_Cleanup(whClientContext* c)
...
+     * unwinding a failed init). The global WH_DEV_ID / WH_DEV_ID_DMA entries
+     * are unregistered unconditionally */
+    if (c->devId != 0) {
+        (void)wc_CryptoCb_UnRegisterDevice(c->devId);
+        (void)wc_CryptoCb_UnRegisterDevice(WH_DEV_ID);
+#ifdef WOLFHSM_CFG_DMA
+        (void)wc_CryptoCb_UnRegisterDevice(WH_DEV_ID_DMA);
+#endif /* WOLFHSM_CFG_DMA */
+    }
     (void)wolfCrypt_Cleanup();
```
bigbrett: fixed

src/wh_client.c
```diff
@@ -85,35 +110,90 @@ int wh_Client_Init(whClientContext* c, const whClientConfig* config)
...
+         * always go to the most recently initialized client, so passing it to
+         * wolfCrypt only works when there is one client; with more, each must
+         * use its own devId. Unregister first in case wolfCrypt won't
+         * re-register the same devId. */
         if (rc == 0) {
-            rc = wc_CryptoCb_RegisterDevice(WH_DEV_ID,
-                    wh_Client_CryptoCb, c);
+            wc_CryptoCb_UnRegisterDevice(WH_DEV_ID);
+            rc = wc_CryptoCb_RegisterDevice(WH_DEV_ID, wh_Client_CryptoCb, c);
             if (rc != 0) {
```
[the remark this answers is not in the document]
bigbrett: this is to mimic the legacy behavior. Broken by design but allows backwards compat.

src/wh_client.c
```diff
@@ -85,35 +110,90 @@ int wh_Client_Init(whClientContext* c, const whClientConfig* config)
...
-                if (rc != 0) {
-                    rc = WH_ERROR_ABORTED;
-                }
+        /* Point the global WH_DEV_ID_DMA at this context (DMA path only).
+         * Single-client only, like WH_DEV_ID above. */
+        if (rc == 0) {
+            wc_CryptoCb_UnRegisterDevice(WH_DEV_ID_DMA);
+            rc = wc_CryptoCb_RegisterDevice(WH_DEV_ID_DMA,
+                                            wh_Client_CryptoCbDma, c);
+            if (rc != 0) {
```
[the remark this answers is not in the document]
bigbrett: same answer as above

src/wh_client.c
```diff
@@ -59,24 +59,49 @@
...
+#endif /* !WOLFHSM_CFG_NO_CRYPTO */
 
-int wh_Client_Init(whClientContext* c, const whClientConfig* config)
-{
-    int rc = 0;
-    if((c == NULL) || (config == NULL)) {
+    /* Client id must be 1..WH_CLIENT_ID_MAX (the server checks it at connect).
+     */
+    if ((c == NULL) || (config == NULL) || (config->comm == NULL) ||
+        (config->comm->client_id == 0) ||
```
billphipps, line 78: Is disallowing client_id==0 required?  I thought comm->client_id would imply an unauthenticated/global access only.

src/wh_client.c
```diff
@@ -85,35 +110,90 @@ int wh_Client_Init(whClientContext* c, const whClientConfig* config)
...
 
+        /* Point the global WH_DEV_ID at this context. Calls on WH_DEV_ID
+         * always go to the most recently initialized client, so passing it to
+         * wolfCrypt only works when there is one client; with more, each must
+         * use its own devId. Unregister first in case wolfCrypt won't
+         * re-register the same devId. */
         if (rc == 0) {
-            rc = wc_CryptoCb_RegisterDevice(WH_DEV_ID,
-                    wh_Client_CryptoCb, c);
+            wc_CryptoCb_UnRegisterDevice(WH_DEV_ID);
```
billphipps, line 140:
```suggestion
            (void)wc_CryptoCb_UnRegisterDevice(WH_DEV_ID);
```
Directly imply that we are ignoring any error return here.

src/wh_client.c
```diff
@@ -436,6 +534,32 @@ int wh_Client_GetCryptoAffinity(whClientContext* c, uint32_t* out_affinity)
     return WH_ERROR_OK;
 }
 
+int wh_Client_SetDmaMode(whClientContext* c, int useDma)
+{
+    if (c == NULL) {
+        return WH_ERROR_BADARGS;
+    }
+#ifdef WOLFHSM_CFG_DMA
+    c->dma.preferDma = (uint32_t)(useDma ? 1 : 0);
```
billphipps, line 543: Nit.  arg to the function is "use" DMA, while the state variable is "prefer".  Should there be another state that mimics the "require DMA" mode when invoked using the DMA DEV ID?  Prolly needless, but just thinking out loud.

test-refactor/misc/wh_test_client_devid.c
```diff
@@ -0,0 +1,313 @@
...
+                 .transport_context = (void*)tmcc2,
+                 .transport_config  = (void*)tmcf2,
+                 .client_id         = WH_TEST_DEFAULT_CLIENT_ID + 1,
+    }};
+    whClientContext             client2[1]  = {0};
+    whClientConfig              c_conf2[1]  = {{
+                      .comm = cc_conf2,
+    }};
+
+    /* Client ids outside 1..WH_CLIENT_ID_MAX are rejected before any
```
billphipps, line 175:
```suggestion
    /* Comm Client ids outside 1..WH_CLIENT_ID_MAX are rejected before any
```

billphipps (approved):
Minor nits and questions.  Resolve at your descretion.  Looks great!

src/wh_client.c
AlexLanzano, line 1: Not sure if I like the "if devId == 0 then use the default WH_DEV_ID"

Opens us up to the case where the user does the following:
```
Client1Cfg = {.devId = 0}
Client2Cfg = {.devId = 1}

wh_Client_Init(Client1Cfg....) // Registers WH_DEV_ID
wh_Client_Init(Client2Cfg....) // Unregisters WH_DEV_ID, Registers devId 1
```
I know it's documented but I can see how a user could easily overlook this.

We definitely should do the following check once my wolfSSL [PR](https://github.com/wolfSSL/wolfssl/pull/10604) lands:
```
if (config->devId != 0 && wc_CryptoCb_IsDeviceRegistered(WH_DEV_ID) == 1) {
    error out
}
```

AlexLanzano (approved):

src/wh_client.c
bigbrett: yes, it is right now, as that value is reserved for server side use only to distinguish the global user (which currently is only used for global key routing). However, that is NOT something currently set by the client, as this is instead communicated via flags.

src/wh_client.c
bigbrett: yeah the backwards compat choices are biting us in the ass already. Not sure there is much we can do about this until your fix lands?

devId == 0 defaulting was the "free" backwards compat so existing clients who zero initialize the config/context don't have to change anything. That said, your scenario is legit, albeit innocuous since it involves mixing the legacy scenario with the new way which should be technically disallowed. That said, I don't think we can do much about it until your fix lands in wolfSSL though, since we would need a registration check.

I think we can just update this once your fix lands.

src/wh_client.c
bigbrett: will address in a follow up

src/wh_client.c
bigbrett: can think about it later and tweak as a follow up

PR #407, opened 2026-06-12
fixes for CC-RH compatibility

miyazakh:
Renesas CC-RH compiler doesn't allow the following line

`byte* p = (uint8_t*)(in + sizeof(foo));`


bigbrett:
seems like this was fixed in a different PR, closing

PR #414, opened 2026-06-23
feat: add sbom Makefile target

MarkAtwood:
> [!IMPORTANT]
> **Blocked on wolfSSL/wolfssl#10343** — `scripts/gen-sbom` has not merged to wolfssl master yet. Do not merge this PR before #10343 lands. Until then the target fails with a clear error pointing at the missing script.

## Summary

Adds `make sbom`, `make install-sbom` and `make uninstall-sbom` targets that produce CycloneDX and SPDX SBOM files for wolfHSM, at parity with the canonical `scripts/sbom.am` fragment and wolfTPM's CMake target.

- Parses version from `ChangeLog.md` (e.g., `# wolfHSM Release v1.4.0`)
- Enumerates sources from `src/*.c` (sorted for reproducible output)
- Artifact hashing is done by `gen-sbom --srcs` (OmniBOR gitoid over the source list). `--srcs` was chosen over `--lib` deliberately: wolfHSM ships as source into embedded builds, so there is no canonical installed library to hash
- Records the effective build configuration by preprocessing `wolfhsm/wh_settings.h` with the same defines/include path as the test build (`WOLFHSM_CFG_DIR` selects the config; point it at your port's `wolfhsm_cfg.h`/`user_settings.h`)
- Lists wolfssl as a dependency component, version read from `WOLFSSL_DIR`; `SBOM_DEP_WOLFSSL=no` for `WOLFHSM_CFG_NO_CRYPTO` builds
- Validates the SPDX output with `pyspdxtools` and converts it to tag-value; `SBOM_VALIDATE=no` skips both
- Honors `SOURCE_DATE_EPOCH` (defaults to last git commit time) — repeated runs are byte-identical
- `install-sbom` places files under `$(PREFIX)/share/doc/wolfhsm` (opt-in; plain `make`/`make install` never touches SBOM files)
- No build system changes — pure Makefile target

## Usage

```sh
make sbom WOLFSSL_DIR=/path/to/wolfssl/source
make install-sbom WOLFSSL_DIR=... [PREFIX=/usr/local] [DESTDIR=...]
```

Outputs: `wolfhsm-<version>.cdx.json`, `wolfhsm-<version>.spdx.json`, `wolfhsm-<version>.spdx` (tag-value)

## Requirements

- `WOLFSSL_DIR` must point to a wolfssl source tree containing `scripts/gen-sbom` (wolfSSL/wolfssl#10343, pending a future wolfssl release)
- `python3`, plus `pip install spdx-tools` unless `SBOM_VALIDATE=no`

## Test plan
- [x] `make sbom WOLFSSL_DIR=...` — CDX/SPDX/tag-value produced, correct version and source count
- [x] Config macros recorded (33 `WOLFHSM_CFG_*`/wolfSSL options with the default test config)
- [x] wolfssl dependency component present with correct version; absent with `SBOM_DEP_WOLFSSL=no`
- [x] Reproducibility: two consecutive runs byte-identical
- [x] `SBOM_VALIDATE=no` skips validation/tag-value; missing `pyspdxtools` fails with a clear error otherwise
- [x] `install-sbom`/`uninstall-sbom` round-trip under `DESTDIR`
- [x] `make -n clean` does not shell out (GNU make 3.81 parse regression check)


README.md
```diff
@@ -25,3 +25,28 @@ please refer to the following resources.
...
+    --supplier "wolfSSL Inc." \
+    --options-h $WOLFSSL_DIR/include/wolfssl/options.h \
+    --srcs $WOLFHSM_DIR/src/*.c
+```
+
+`WOLFSSL_DIR` must point to a wolfssl source tree containing `scripts/gen-sbom`
+(branch `feat/sbom-embedded`, or `master` once wolfSSL/wolfssl#10343 merges).
+`WOLFHSM_DIR` is the root of the wolfHSM source tree.
+
+Requires `python3` and `pyspdxtools` (`pip install spdx-tools`).
```
bigbrett: this doc snippet looks outdated based on the changes in this PR, shouldn't we just recommend the end user run the new makefile target?

bigbrett (changes requested):
Running `make sbom WOLFSSL_DIR=../wolfssl` against the latest wolfSSL master I get the following error:

```
ERROR: ../wolfssl/scripts/gen-sbom not found.                                                                                                                                                 
       Use a wolfSSL tree that includes SBOM support.                                                                                                                                         
make: *** [Makefile:73: sbom] Error 1                       
```

Did this not merge into wolfSSL master yet? If so it shouldn't merge into wolfHSM.

dgarske (changes requested):
https://github.com/wolfSSL/wolfHSM/pull/414#pullrequestreview-4649786841

https://github.com/wolfSSL/wolfHSM/pull/414#pullrequestreview-4549387514

Makefile
```diff
@@ -56,3 +56,48 @@ clean:
 	make -C benchmark clean
 	make -C tools clean
 	make -C examples clean
+
+# ---- SBOM generation ----
+CC            ?= cc
+WOLFSSL_DIR   ?= ../../wolfssl
```
jackctj117: **🟠 [Medium] Default WOLFSSL_DIR is off by one level for the top-level Makefile** · _convention_

The new default `WOLFSSL_DIR ?= ../../wolfssl` is copied from the sub-Makefiles in `test/`, `benchmark/`, etc., which live one directory below the repo root — for them, `../../wolfssl` correctly resolves to a sibling checkout. But this assignment is in the *top-level* Makefile (repo root), so `../../wolfssl` resolves to the grandparent directory, i.e. `\<grandparent>/wolfssl`, not the documented sibling `\<parent>/wolfssl`. On a standard sibling checkout, `make sbom` with no override will fail the `[ ! -d "$(WOLFSSL_DIR)" ]` guard. The correct default here is `../wolfssl`. This won't cause silent misbehavior (the directory-exists guard produces a clear error and the PR docs tell users to pass `WOLFSSL_DIR=` explicitly), but the default is misleading and inconsistent with the repo's documented sibling layout (see CLAUDE.md 'Dependency layout').

**Fix:** Change the top-level default to `../wolfssl` so a standard sibling checkout works out-of-the-box, or drop the default entirely and rely solely on the guard message since the docs already require an explicit `WOLFSSL_DIR=`.

Makefile
```diff
@@ -56,3 +56,48 @@ clean:
...
+	    exit 1; \
+	fi
+	@if [ ! -f "$(WOLFSSL_DIR)/scripts/gen-sbom" ]; then \
+	    echo "ERROR: $(WOLFSSL_DIR)/scripts/gen-sbom not found." >&2; \
+	    echo "       Use a wolfSSL tree that includes SBOM support." >&2; \
+	    exit 1; \
+	fi
+	@echo "wolfHSM version: $(VERSION)"
+	@echo "Sources:         $(words $(SRCS)) .c files in src/"
+	@_defines=$$(mktemp /tmp/wolfhsm-defines.XXXXXX) && \
```
jackctj117: **🟠 [Medium] gen-sbom --options-h receives only compiler builtin macros, not wolfHSM/wolfSSL config** · _Logic_

The new `sbom` target generates the defines file passed to `--options-h` by preprocessing an EMPTY translation unit: `$(CC) -dM -E -I. -I$(WOLFSSL_DIR) -x c /dev/null`. Because `/dev/null` includes no headers, `-dM -E` emits only the compiler's predefined macros (e.g. `__GNUC__`, `__STDC__`), NOT any `WOLFHSM_CFG_*` or wolfSSL `HAVE_*`/`NO_*` configuration macros. The `-I.` / `-I$(WOLFSSL_DIR)` include paths have no effect since nothing is `#include`d. As a result the SBOM's build-configuration record is effectively empty, which undermines the stated CRA-compliance purpose of accurately capturing the build configuration, and diverges from the README, which instead points `--options-h` at a real `options.h`. Severity views differ across scans: the review mode rated this Medium (Low confidence, treating capture of compiler builtins as possibly intentional), while the review-security mode rated it Low (High confidence); the stricter Medium is retained.

**Fix:** Confirm what gen-sbom expects in --options-h. If it should reflect wolfHSM/wolfSSL configured macros, preprocess a stub that includes the relevant config header (e.g. `wolfhsm/wh_settings.h`) with the appropriate user_settings/config defines, or point `--options-h` at a concrete options header as the README shows; if only compiler builtins are needed, drop the no-op `-I.`/`-I$(WOLFSSL_DIR)` flags to avoid implying otherwise.

Makefile
```diff
@@ -56,3 +56,48 @@ clean:
 	make -C benchmark clean
 	make -C tools clean
 	make -C examples clean
+
+# ---- SBOM generation ----
+CC            ?= cc
+WOLFSSL_DIR   ?= ../../wolfssl
+VERSION       := $(shell sed -n 's/^# wolfHSM Release v\([0-9][0-9.]*\).*/\1/p' ChangeLog.md | head -1)
```
jackctj117: **🔵 [Low] VERSION shell-out runs on every make invocation, not just sbom** · _Logic_

`VERSION := $(shell sed -n ... ChangeLog.md | head -1)` uses immediate (`:=`) expansion at the top level of the root Makefile, so the `sed | head` subshell runs on *every* invocation (`make all`, `make test`, `make clean`, etc.), even though `VERSION` is only used by the `sbom` target. This is harmless but wasteful, and if `ChangeLog.md` were ever absent the shell would still run (silently producing an empty value) for unrelated targets. Note the SBOM_CDX/SBOM_SPDX names reference VERSION, so a lazy `=` assignment keeps them working.

**Fix:** Use a recursive `=` assignment (evaluated only when referenced) or a target-specific variable so the shell only runs when the `sbom` target is invoked.

Makefile
jackctj117: **🔵 [Low] Temp file uses hardcoded /tmp prefix, ignoring $TMPDIR** · _Resource Leak_

The defines file is created with `mktemp /tmp/wolfhsm-defines.XXXXXX`, hardcoding `/tmp`. `mktemp` is used safely here (atomic creation with a random `XXXXXX` suffix and mode 0600, plus a `trap ... EXIT` cleanup scoped to the single recipe shell), so there is no symlink/predictability/TOCTOU vulnerability. However, hardcoding `/tmp` ignores `$TMPDIR` and can fail on environments where `/tmp` is unavailable or non-writable (some CI sandboxes, restricted containers, macOS).

**Fix:** Use `mktemp` without a hardcoded directory (respects `$TMPDIR`), e.g. `mktemp "$${TMPDIR:-/tmp}/wolfhsm-defines.XXXXXX"` or plain `mktemp`, for better portability across macOS/CI sandboxes.

README.md
```diff
@@ -25,3 +25,28 @@ please refer to the following resources.
...
+
+## SBOM / EU CRA Compliance
+
+wolfHSM generates a Software Bill of Materials (SBOM) in CycloneDX 1.6 and
+SPDX 2.3 formats to support compliance with the EU Cyber Resilience Act (CRA).
+
+wolfHSM uses a custom build system; invoke `gen-sbom` from the wolfssl source
+tree directly:
+
+```sh
```
jackctj117: **🔵 [Low] README manual invocation diverges from the Makefile sbom target** · _Logic_

The README documents a manual `gen-sbom` invocation that differs from what `make sbom` actually runs: the README passes `--options-h $WOLFSSL_DIR/include/wolfssl/options.h` (a checked-in wolfSSL header) whereas the Makefile generates a defines file via `$(CC) -dM -E`; the README omits `--license-file LICENSING`, `--cdx-out`, and `--spdx-out` that the Makefile passes; and the README derives the version with `head -1 ... | grep -oE '[0-9]+\.[0-9]+\.[0-9]+'` while the Makefile uses a `sed` pattern. Two documented-but-divergent invocation paths risk drifting out of sync and produce SBOMs with different content depending on which route a user follows.

**Fix:** Lead the README section with `make sbom WOLFSSL_DIR=...` (matching the new target, including `--license-file`) and keep the raw gen-sbom command as an explicit 'manual invocation' fallback, so both routes produce equivalent SBOMs and the documented flags stay consistent with the Makefile.

jackctj117 (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`**Overall recommendation:** `COMMENT`
**Findings:** 5 total — 5 posted, 0 skipped
**5 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Medium] [review] **Default WOLFSSL_DIR is off by one level for the top-level Makefile** — `Makefile:62`
- [Medium] [review+review-security] **gen-sbom --options-h receives only compiler builtin macros, not wolfHSM/wolfSSL config** — `Makefile:87-99`
- [Low] [review+review-security] **VERSION shell-out runs on every make invocation, not just sbom** — `Makefile:63`
- [Low] [review+review-security] **Temp file uses hardcoded /tmp prefix, ignoring $TMPDIR** — `Makefile:87`
- [Low] [review+review-security] **README manual invocation diverges from the Makefile sbom target** — `README.md:37-44`

*Review generated by [Skoll](https://github.com/wolfSSL/fenrir/tree/main/skoll)*

Makefile
jackctj117: Blocker — Makefile fails to parse on GNU Make 3.81 (macOS default make)
Line 63's VERSION := $(shell sed ... '\([0-9][0-9.]*\)' ...) trips a known make-3.81 function-scanner bug on the \(...\) parens. Because it's a file-scope :=, the error aborts every target, not just sbom:
$ make -n clean
Makefile:63: *** unterminated call to function `shell': missing `)'.  Stop.
Repro: make 3.81 fails; gmake 4.4.1 is fine. The base Makefile parses under 3.81, so this is a regression. Fix by dropping the capture-group parens (e.g. grep -oE 'v[0-9.]+' + strip) or computing VERSION inside the sbom recipe (which also resolves Copilot's every-invocation note).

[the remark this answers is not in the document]
MarkAtwood: Fixed — changed `VERSION :=` to a recursive `VERSION =` (and SBOM_CDX/SBOM_SPDX likewise) so the sed only runs when the sbom target is invoked. Confirmed `make -n clean` no longer shells out.

Makefile
```diff
@@ -56,3 +56,48 @@ clean:
...
+SRCS          := $(wildcard src/*.c)
```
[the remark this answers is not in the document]
MarkAtwood: Fixed — `SRCS := $(sort $(wildcard src/*.c))` for reproducible ordering. Left the empty-list guard out: src/ always has .c files here and gen-sbom already errors on an empty --srcs.

Makefile
```diff
@@ -56,3 +56,48 @@ clean:
...
+	fi && \
+	_py=$$(command -v python3 2>/dev/null || command -v python 2>/dev/null) && \
+	[ -n "$$_py" ] || { echo "ERROR: python3 not found." >&2; exit 1; } && \
+	"$$_py" $(WOLFSSL_DIR)/scripts/gen-sbom \
+	    --name wolfhsm \
+	    --version $(VERSION) \
+	    --supplier "wolfSSL Inc." \
+	    --license-file LICENSING \
+	    --options-h "$$_defines" \
+	    --srcs $(SRCS) \
```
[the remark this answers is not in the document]
MarkAtwood: Fixed — `SRCS := $(sort $(wildcard src/*.c))` for reproducible ordering.

Makefile
```diff
@@ -56,3 +56,48 @@ clean:
...
+	@echo "wolfHSM version: $(VERSION)"
+	@echo "Sources:         $(words $(SRCS)) .c files in src/"
+	@_defines=$$(mktemp /tmp/wolfhsm-defines.XXXXXX) && \
+	trap 'rm -f "$$_defines"' EXIT && \
+	if ! $(CC) -dM -E -I. -I$(WOLFSSL_DIR) -x c /dev/null >"$$_defines" 2>/dev/null; then \
+	    echo "ERROR: $(CC) -dM -E failed." >&2; exit 1; \
+	fi && \
+	_py=$$(command -v python3 2>/dev/null || command -v python 2>/dev/null) && \
+	[ -n "$$_py" ] || { echo "ERROR: python3 not found." >&2; exit 1; } && \
+	"$$_py" $(WOLFSSL_DIR)/scripts/gen-sbom \
```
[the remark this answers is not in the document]
MarkAtwood: Fixed — requires python3 explicitly now and errors if it's missing; dropped the `python` fallback.

Makefile
```diff
@@ -56,3 +56,48 @@ clean:
...
+	fi
+	@if [ ! -f "$(WOLFSSL_DIR)/scripts/gen-sbom" ]; then \
+	    echo "ERROR: $(WOLFSSL_DIR)/scripts/gen-sbom not found." >&2; \
+	    echo "       Use a wolfSSL tree that includes SBOM support." >&2; \
+	    exit 1; \
+	fi
+	@echo "wolfHSM version: $(VERSION)"
+	@echo "Sources:         $(words $(SRCS)) .c files in src/"
+	@_defines=$$(mktemp /tmp/wolfhsm-defines.XXXXXX) && \
+	trap 'rm -f "$$_defines"' EXIT && \
```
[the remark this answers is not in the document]
MarkAtwood: Fixed — switched to `trap '...' 0`, the portable form.

README.md
MarkAtwood: Fixed — the section now leads with `make sbom WOLFSSL_DIR=../wolfssl` and keeps the raw gen-sbom command as a documented fallback, with flags matching the target.

Makefile
MarkAtwood: Fixed — the root Makefile's sibling checkout is `../wolfssl`, so the default is now `../wolfssl`.

Makefile
MarkAtwood: Fixed — `mktemp "$${TMPDIR:-/tmp}/wolfhsm-defines.XXXXXX"` so it respects $TMPDIR.

README.md
MarkAtwood: Fixed — README manual invocation now matches the Makefile target (same flags and version derivation), with the raw command kept as a documented fallback.

Makefile
MarkAtwood: Fixed — recursive `VERSION =` now; the sed only runs when the sbom target is invoked.

MarkAtwood: Fixed in 4c0af52. The actual trigger is the `#` inside the file-scope $(shell): make <= 4.2 strips everything after it as a comment at read time, which is what leaves the unterminated function call — that's also why it aborted every target. The version extraction no longer uses `#` or escaped parens (`sed -n 's/^. wolfHSM Release v//p' | cut -d' ' -f1`), so nothing make-version-sensitive is left at file scope. Verified parse (`make -n clean`) and a full `make sbom` end-to-end on GNU Make 4.3; I don't have a 3.81 here, so a macOS confirm would be welcome.

Makefile
MarkAtwood: Good catch — confirmed. Preprocessing /dev/null with no -include emits only compiler builtins, which gen-sbom's noise filter then drops, so the SBOM's config section is effectively empty (the generated CDX had zero config macros). Deciding the right fix — point --options-h at a real options header, add `-include <config header>`, or use the --user-settings path — is a small design question I'd rather settle deliberately than paper over. Tracking it as a follow-up; not folding it into this PR unless you want it here.

MarkAtwood:
@bigbrett Correct — gen-sbom hasn't merged to wolfssl master. It's in wolfssl PR wolfSSL/wolfssl#10343, which is still open, so no wolfssl release or master checkout has scripts/gen-sbom yet. The target can't create that file; what it can do is fail clearly. I updated the error to name the dependency and point you at setting WOLFSSL_DIR to a tree that has the script, and the README says the same. So the wolfHSM side is ready to use the moment #10343 lands; it just no-ops with a clear message until then. I did not vendor a copy of gen-sbom into wolfHSM. If you'd rather hold this PR until #10343 merges, flag it and I'll mark it blocked.

Makefile
MarkAtwood: Update: implemented in 4cef60f rather than deferred. The recipe now preprocesses wolfhsm/wh_settings.h with the same defines/include path as the test build (-DWOLFHSM_CFG -DWOLFSSL_USER_SETTINGS, WOLFHSM_CFG_DIR on the path, default test/config), so the SBOM records the effective config — 31 WOLFHSM_CFG_* plus the wolfSSL options from user_settings.h, verified in the generated CDX. Integrators point WOLFHSM_CFG_DIR at their port's config dir. Known trade-off, documented inline on the target: wh_settings.h's system includes drag ~330 libc constants into the dump, and they are deliberately not filtered here because a prefix allowlist would drop real unprefixed options (GCM_TABLE_4BIT, FP_MAX_BITS, ...). The durable fix is a provenance-aware noise filter in gen-sbom itself (wolfSSL#10343); the Makefile comment spells out two implementation options and this recipe needs no change once that lands.

MarkAtwood:
Force-pushed a cleanup: this branch was cut from a stale shallow base, which made ~7,800 lines of already-merged main content (armv8m-tz, hwkeystore, test files) appear as PR additions. Rebased onto current main; the PR now contains only the SBOM change: +133 in Makefile and README.

Verified on the rebased head: `make sbom WOLFSSL_DIR=<wolfssl> WOLFHSM_CFG_DIR=test/config` generates both artifacts (wolfhsm-1.4.0.cdx.json, wolfhsm-1.4.0.spdx.json) against wolfSSL PR #10343's gen-sbom.

@bigbrett @dgarske ready for re-review; your earlier comments on the unrelated files are moot.

MarkAtwood:
Resolved the July 10 review threads: every fix discussed in them is carried in the rebuilt branch (d03b224 contains the cumulative Makefile/README state, verified end-to-end with make sbom). The conversation history references pre-rebuild commit ids; the content is all here.

sameehj:
@MarkAtwood 

direction is right, but the target should reuse gen-sbom the way the rest of the family does

Thanks Mark. The overall approach — a dedicated make sbom target that calls wolfSSL's gen-sbom — is the correct call for wolfHSM, since wolfHSM is a hand-written Makefile and can't include scripts/sbom.am like the autotools repos (wolfMQTT, wolfProvider, wolfCLU, wolfEngine, wolfssh, wolfscep). We already have a precedent for a non-automake reimplementation: wolfTPM's CMake sbom target, which was explicitly brought to parity with the canonical scripts/sbom.am. That's the bar this PR should hit.

Right now #414 diverges from the canonical generator in a few ways, which is what the review threads are circling:

It rolls its own source hash (SHA-256 fingerprint of src/*.c) instead of using gen-sbom's own input. gen-sbom already supports --lib (hash the built library from a staged install) and --srcs/--srcs-file (OmniBOR gitoid for embedded/source builds). Using a private hash means the wolfHSM SBOM checksum isn't comparable to the other products and doesn't describe the shipped artifact. Please pick --lib or --srcs deliberately and note which in the PR.

--options-h only receives compiler builtins, not wolfHSM/wolfSSL config macros (jackctj117's finding). The point of --options-h is to record the configured feature macros, so this needs the real CC -dM -E dump with wolfHSM's actual CPPFLAGS/CFLAGS + config header.

Missing parity items the fragment/CMake target have: install-sbom / uninstall-sbom, SOURCE_DATE_EPOCH for reproducible output, and the SPDX validation / tag-value step.

Sequencing: as bigbrett and dgarske noted, gen-sbom is still on wolfSSL PR #10343 and hasn't merged to master. The clear-error no-op behavior is fine, but this PR shouldn't merge before #10343 lands — let's mark it blocked-on that.

Net: keep the bespoke-Makefile approach, but rework it to parity with wolfTPM's target (drive everything through gen-sbom, capture real macros, add the install/reproducibility bits), and gate the merge on #10343. Happy to pair on the diff.

MarkAtwood:
@sameehj parity items are in as of 073f4eb:

- **SPDX validation + tag-value**: `pyspdxtools` validates the SPDX output and emits `wolfhsm-<ver>.spdx`. `SBOM_VALIDATE=no` skips both (same gate name #10343 is growing for the Gentoo case).
- **SOURCE_DATE_EPOCH**: honored, defaults to last git commit time. Verified two consecutive runs produce byte-identical output.
- **install-sbom / uninstall-sbom**: opt-in, `$(PREFIX)/share/doc/wolfhsm` (`PREFIX=/usr/local` default, `DESTDIR` honored). No `uninstall-hook` chaining because the root Makefile has no install/uninstall targets to hook.
- **Dependency component**: the SBOM now lists wolfssl with the version read from `WOLFSSL_DIR` (`--dep-wolfssl`); `SBOM_DEP_WOLFSSL=no` for `WOLFHSM_CFG_NO_CRYPTO` builds.

On the artifact hash: the July 14 rebuild already dropped the private fingerprint — hashing goes through `gen-sbom --srcs` (sorted `src/*.c`, OmniBOR gitoid). That over `--lib` is deliberate: wolfHSM ships as source into embedded builds, so there is no canonical installed library to hash. Now stated in the PR description.

Config macros: also in since the rebuild — the recipe preprocesses `wolfhsm/wh_settings.h` with the same defines/include path as the test build, so the SBOM records effective `WOLFHSM_CFG_*` + wolfSSL options (33 config macros with the default test config), not compiler builtins.

Sequencing: agreed. Marked blocked on wolfSSL/wolfssl#10343 at the top of the description — @bigbrett that also answers your question: gen-sbom is still on that PR, this one must not merge before it.

README.md
```diff
@@ -25,3 +25,69 @@ please refer to the following resources.
 - [wolfHSM Manual](https://www.wolfssl.com/documentation/manuals/wolfhsm/index.html)
 - [wolfHSM API Reference](https://www.wolfssl.com/documentation/manuals/wolfhsm/appendix01.html)
 - [wolfHSM Examples](https://github.com/wolfSSL/wolfHSM/tree/main/examples)
+
+## SBOM / EU CRA Compliance
```
bigbrett: Could we pls move this to a new markdown document like SBOM.md or EU-CRA.md?, I'd like to keep the top level README as lean as possible. You can add an sbom generation bullet to the bulleted list above pointing to that doc if you want

bigbrett (changes requested):

sameehj:
## Update: migrate to vendored wolfGlass driver

Replaces the hand-rolled `gen-sbom` Makefile recipe with vendored `tools/sbom` from wolfGlass `9bdf5b7`.

### Critical fix vs earlier draft
Config capture uses `SBOM_OPTIONS_H` (preprocess `wh_settings.h` with `-I`/`-include`), **not** `SBOM_CFLAGS`. The driver's `--cflags` path drops `-I`/`-include` and previously produced a 2-property empty SBOM.

### Also in this update
- `WOLFHSM_CFG_DIR` is required (no silent `test/config` default)
- Core-only CI canary with `--min-properties 50` and `--require-dep-version wolfssl`
- Docs moved to `docs/SBOM.md`; follow-ups listed in `docs/SBOM-FOLLOWUPS.md`
- `.gitignore` un-ignores `tools/sbom/build/` (bare `Build/` matched it on macOS)

Local smoke: 1793 properties, `wolfssl` dep version `5.9.2`.

MarkAtwood:
gen-sbom is on wolfssl master now (#10343), so `make sbom WOLFSSL_DIR=../wolfssl WOLFHSM_CFG_DIR=<config>` works against master. The error you hit is gone.

Both failing checks are fixed.

scan-build: the WOLFHSM_CFG_DIR guard was a plain `$(error)` at file scope. Make evaluates that when it reads the Makefile, so it killed every target, not just the sbom ones. `make scan` died at Makefile:80. It now only fires when an sbom goal is actually requested.

sbom_canary: gcc and clang gave different SBOMs. `-dM` dumps the compiler's own builtins. gen-sbom drops `__`-prefixed names but keeps system macros whose values point at them, so ATOMIC_BOOL_LOCK_FREE expands to `__GCC_ATOMIC_BOOL_LOCK_FREE` on gcc and `__CLANG_ATOMIC_BOOL_LOCK_FREE` on clang. Function-like macro bodies differ too (`idx ##VAR` vs `idx##VAR`). The capture now keeps only object-like project config macros: 230 instead of 2138, identical under both compilers. All the WOLFHSM_CFG_*, WOLFSSL_*, HAVE_* and NO_* macros are still recorded. What went away is glibc stuff like BUFSIZ.

The filter is a prefix allowlist, which will rot. I tried it without one (all-caps, object-like, skip `__`) and it still differed by 34 lines, so shape alone isn't enough to tell config from system macros. Follow-up #5 in docs/SBOM-FOLLOWUPS.md is to read the macro list from the settings headers instead, probably in the shared driver.

README: cut to one bullet in the resources list, as you asked. docs/SBOM.md already covers the WOLFHSM_CFG_DIR requirement.

048c405 is green.

PR #418, opened 2026-06-23
Fix benchmark guard macros

miyazakh:
# Fix benchmark guard macros

## Summary

- Fix incorrect preprocessor guard for ML-DSA 87 benchmark functions in `wh_bench_mod_mldsa.c`
- Add missing `WOLFHSM_CFG_ENABLE_SERVER` guard around `wh_Bench_ServerCfgLoop()` in `wh_bench.c`

## Changes

| File | Description |
|------|-------------|
| `benchmark/bench_modules/wh_bench_mod_mldsa.c` | Replace incorrect guard `WOLFSSL_MLDSA_NO_SIGN` with `WOLFSSL_NO_ML_DSA_87` for ML-DSA 87 benchmark section |
| `benchmark/wh_bench.c` | Wrap `wh_Bench_ServerCfgLoop()` with `#if defined(WOLFHSM_CFG_ENABLE_SERVER)` guard |

### Fix typo in ML-DSA 87 guard (`wh_bench_mod_mldsa.c`)

The wrong macro `WOLFSSL_MLDSA_NO_SIGN` was used to guard the ML-DSA 87 benchmark section. The correct macro is `WOLFSSL_NO_ML_DSA_87`, consistent with the guards used for ML-DSA 44 and ML-DSA 65 in the same file.

- **Before:** `#if !defined(WOLFSSL_MLDSA_NO_SIGN)`
- **After:** `#if !defined(WOLFSSL_NO_ML_DSA_87)`

Without this fix, defining `WOLFSSL_NO_ML_DSA_87` to disable ML-DSA 87 support would not exclude the corresponding benchmark functions from compilation.

### Guard server benchmark loop (`wh_bench.c`)

`wh_Bench_ServerCfgLoop()` uses server APIs (`wh_Server_Init`, `wh_Server_HandleRequestMessage`, etc.) and must not be compiled in configurations where server support is disabled. A stub implementation is provided under `#else` so that callers such as `_whBenchServerTask()` link cleanly without server support.

- **Before:** No guard
- **After:** Full implementation wrapped with `#if defined(WOLFHSM_CFG_ENABLE_SERVER)`; stub (returns `WH_ERROR_OK`) provided under `#else`

## Impact

- No logic changes; preprocessor guard fixes only.
- Builds with `WOLFSSL_NO_ML_DSA_87` will now correctly exclude ML-DSA 87 benchmark functions.
- Builds without `WOLFHSM_CFG_ENABLE_SERVER` will now link cleanly via the stub implementation.

### DIFF
--- a/benchmark/bench_modules/wh_bench_mod_mldsa.c
+++ b/benchmark/bench_modules/wh_bench_mod_mldsa.c
@@ -1079,61 +1079,112 @@ int wh_Bench_Mod_MlDsa44KeyGenDma(whClientContext*  client,
 int wh_Bench_Mod_MlDsa65Sign(whClientContext* client, whBenchOpContext* ctx,
                              int id, void* params)
 {
+#if !defined(WOLFSSL_MLDSA_NO_SIGN) && \
+    !defined(WOLFHSM_CFG_TEST_CLIENT_LARGE_DATA_DMA_ONLY)
     (void)client;
     (void)ctx;
     (void)id;
     (void)params;
     return WH_ERROR_NOTIMPL;
+#else
+    (void)client;
+    (void)ctx;
+    (void)id;
+    (void)params;
+    return WH_ERROR_NOTIMPL;
+#endif
 }
 
 int wh_Bench_Mod_MlDsa65SignDma(whClientContext* client, whBenchOpContext* ctx,
                                 int id, void* params)
 {
+#if defined(WOLFHSM_CFG_DMA) && !defined(WOLFSSL_MLDSA_NO_SIGN)
     (void)client;
     (void)ctx;
     (void)id;
     (void)params;
     return WH_ERROR_NOTIMPL;
+#else
+    (void)client;
+    (void)ctx;
+    (void)id;
+    (void)params;
+    return WH_ERROR_NOTIMPL;
+#endif
 }
 
 int wh_Bench_Mod_MlDsa65Verify(whClientContext* client, whBenchOpContext* ctx,
                                int id, void* params)
 {
+#if !defined(WOLFSSL_MLDSA_NO_VERIFY) && \
+    !defined(WOLFHSM_CFG_TEST_CLIENT_LARGE_DATA_DMA_ONLY)
     (void)client;
     (void)ctx;
     (void)id;
     (void)params;
     return WH_ERROR_NOTIMPL;
+#else
+    (void)client;
+    (void)ctx;
+    (void)id;
+    (void)params;
+    return WH_ERROR_NOTIMPL;
+#endif
 }
 
 int wh_Bench_Mod_MlDsa65VerifyDma(whClientContext*  client,
                                   whBenchOpContext* ctx, int id, void* params)
 {
+#if defined(WOLFHSM_CFG_DMA) && !defined(WOLFSSL_MLDSA_NO_VERIFY)
     (void)client;
     (void)ctx;
     (void)id;
     (void)params;
     return WH_ERROR_NOTIMPL;
+#else
+    (void)client;
+    (void)ctx;
+    (void)id;
+    (void)params;
+    return WH_ERROR_NOTIMPL;
+#endif
 }
 
 int wh_Bench_Mod_MlDsa65KeyGen(whClientContext* client, whBenchOpContext* ctx,
                                int id, void* params)
 {
+#if !defined(WOLFSSL_MLDSA_NO_MAKE_KEY) && \
+    !defined(WOLFHSM_CFG_TEST_CLIENT_LARGE_DATA_DMA_ONLY)
     (void)client;
     (void)ctx;
     (void)id;
     (void)params;
     return WH_ERROR_NOTIMPL;
+#else
+    (void)client;
+    (void)ctx;
+    (void)id;
+    (void)params;
+    return WH_ERROR_NOTIMPL;
+#endif
 }
 
 int wh_Bench_Mod_MlDsa65KeyGenDma(whClientContext*  client,
                                   whBenchOpContext* ctx, int id, void* params)
 {
+#if defined(WOLFHSM_CFG_DMA) && !defined(WOLFSSL_MLDSA_NO_MAKE_KEY)
     (void)client;
     (void)ctx;
     (void)id;
     (void)params;
     return WH_ERROR_NOTIMPL;
+#else
+    (void)client;
+    (void)ctx;
+    (void)id;
+    (void)params;
+    return WH_ERROR_NOTIMPL;
+#endif
 }
 #endif /* !defined(WOLFSSL_NO_ML_DSA_65) */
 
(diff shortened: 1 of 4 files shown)


benchmark/wh_bench.c
```diff
@@ -1,1195 +1,1197 @@
...
+    /* Clean up */
+    if (ret == WH_ERROR_NOTREADY) {
+        /* Ignore not ready status */
+        ret = WH_ERROR_OK;
+    }
+    (void)wh_Server_Cleanup(server);
+
+    return ret;
+}
+#endif /* WOLFHSM_CFG_ENABLE_SERVER */
```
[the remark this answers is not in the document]
miyazakh: No, this concern is not valid.                             
                                         
  After our changes, the declaration in wh_bench.h, the      
  implementation in wh_bench.c, and every call site are all  
  consistently guarded by WOLFHSM_CFG_ENABLE_SERVER. There is
   no code path that can reach wh_Bench_ServerCfgLoop()      
  without the guard defined, so no undefined symbol issue    
  exists.

  If a caller outside the guard were added in the future, the
   missing declaration would produce a compile-time error —
  which is earlier and more explicit than the linker error a
  stub would prevent.

benchmark/wh_bench.h
```diff
@@ -47,8 +39,18 @@ int wh_Bench_ClientCfg(whClientConfig* clientCfg, int transport);
...
+ * Runs the client benchmarks against a server using POSIX threads
+ * transport: The type of transport to use for communication
+ * moduleIndex: The specific benchmark module to run (optional, -1 for all)
+ * Returns 0 on success and a non-zero error code on failure
+ */
+int wh_Bench_ClientServer_Posix(int transport, int moduleIndex);
+
 /* Server-side processing loop for benchmarking */
 int wh_Bench_ServerCfgLoop(whServerConfig* serverCfg);
+#endif /* WOLFHSM_CFG_ENABLE_SERVER */
```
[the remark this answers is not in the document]
miyazakh: The call site in wh_bench_main.c was updated from #if               
  defined(WOLFHSM_CFG_TEST_POSIX) to #if 
  defined(WOLFHSM_CFG_TEST_POSIX) &&                                  
  defined(WOLFHSM_CFG_ENABLE_SERVER), ensuring the call is compiled
  out when server support is disabled.

benchmark/wh_bench.c
```diff
@@ -1,1195 +1,1197 @@
...
+
+#ifndef WOLFHSM_CFG_NO_CRYPTO
+    wc_FreeRng(crypto->rng);
+    wolfCrypt_Cleanup();
+#endif
+
+    return WH_ERROR_OK;
+}
+
+#endif /* WOLFHSM_CFG_ENABLE_SERVER */
```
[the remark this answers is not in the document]
miyazakh: No, this concern is not valid.

benchmark/bench_modules/wh_bench_mod_mldsa.c
```diff
@@ -1138,7 +1138,7 @@ int wh_Bench_Mod_MlDsa65KeyGenDma(whClientContext*  client,
 #endif /* !defined(WOLFSSL_NO_ML_DSA_65) */
 
 /* Benchmark functions for ML-DSA 87 */
-#if !defined(WOLFSSL_MLDSA_NO_SIGN)
+#if !defined(WOLFSSL_NO_ML_DSA_87)
```
bigbrett, line 1192: Do we not also still need `WOLFSSL_MLDSA_NO_SIGN` here? Shouldn't all the sign operations be gated on `WOLFSSL_MLDSA_NO_SIGN` and then individual security levels gated by their respective macros?

benchmark/wh_bench.h
```diff
@@ -47,8 +39,18 @@ int wh_Bench_ClientCfg(whClientConfig* clientCfg, int transport);
  */
 int wh_Bench_ClientCtx(whClientContext* client, int transport);
 
+#if defined(WOLFHSM_CFG_ENABLE_SERVER)
```
bigbrett, line 43: should be both client and server here?

benchmark/wh_bench_main.c
```diff
@@ -81,12 +81,16 @@ int main(int argc, char** argv)
         }
     }
 
-#if defined(WOLFHSM_CFG_TEST_POSIX)
+#if defined(WOLFHSM_CFG_TEST_POSIX) && defined(WOLFHSM_CFG_ENABLE_SERVER)
```
bigbrett, line 84: client too, right?

bigbrett (changes requested):

benchmark/bench_modules/wh_bench_mod_mldsa.c
miyazakh: Add `WOLFSSL_MLDSA_NO_SIGN` and related operation-specific guard macros.

benchmark/wh_bench.h
miyazakh: I think `WOLFHSM_CFG_ENABLE_CLIENT` is not needed here since the benchmark always assumes client support is enabled — consistent with `wh_Bench_ClientCfg` and `wh_Bench_ClientCtx`, which are also not gated on it.

Updated `wh_bench.h` to match the guards used in `wh_bench.c`: wrapped all declarations in `WOLFHSM_CFG_BENCH_ENABLE`, moved `wh_server.h` inside `WOLFHSM_CFG_ENABLE_SERVER`, and added `WOLFHSM_CFG_TEST_POSIX `around `wh_Bench_ClientServer_Posix` to match its definition and call site.

benchmark/wh_bench_main.c
miyazakh: My understanding is that benchmark code always assumes client support is enabled, making `WOLFHSM_CFG_ENABLE_CLIENT `implicit and redundant here.

benchmark/bench_modules/wh_bench_mod_mldsa.c
```diff
@@ -1079,124 +1079,226 @@ int wh_Bench_Mod_MlDsa44KeyGenDma(whClientContext*  client,
 int wh_Bench_Mod_MlDsa65Sign(whClientContext* client, whBenchOpContext* ctx,
                              int id, void* params)
 {
+#if !defined(WOLFSSL_MLDSA_NO_SIGN) && \
+    !defined(WOLFHSM_CFG_TEST_CLIENT_LARGE_DATA_DMA_ONLY)
```
bigbrett, line 1083: I'm confused why in your latest commit you added all of these gates putting identical void casting logic in both cases? We shouldnt add conditional implementations unless it does something.

bigbrett (changes requested):

benchmark/bench_modules/wh_bench_mod_mldsa.c
miyazakh: In your earlier comment, you asked, "Do we not also still need WOLFSSL_MLDSA_NO_SIGN here?", so I guarded the sign operation in the same way as `wh_Bench_Mod_MlDsa44Sign()`, which returns `WH_ERROR_NOTIMPL `when `WOLFSSL_MLDSA_NO_SIGN `is defined.

However, since `_benchMlDsaSign`() does not currently support `ML-DSA 65/87`, the added guard creates a redundant code path. I applied the same pattern to the other functions for consistency.

Going forward, I see two options:

1. Remove the added macro guards.
2. Extend `_benchMlDsaSign(), _benchMlDsaVerify(), and _benchMlDsaKeyGen()` to support ML-DSA 65 and 87.

bigbrett: based on your original diff, the comment was valid since you removed the dependency on sign, but I see no why you did that. I can merge this now but can you please add a follow up PR introducing benchmark support for all security levels? Thanks!

bigbrett (approved):

PR #424, opened 2026-06-25
Add test case that wrapped keys cannot be committed or erased

padelsbach:
Related to F-4707 (false positive)


test-refactor/misc/wh_test_multiclient.c
```diff
@@ -1240,6 +1240,89 @@ static int _testWrappedKey_LocalWrap_GlobalKey_NonOwnerNoWrapKey(
...
 #undef WRAPPED_KEY_SIZE
 }
+
+/*
+ * Test: Unwrapped key cannot persist to NVM
+ * - Wrapped keys are cache-only; the unwrapped key must never reach NVM
+ * - This invariant prevents an unwrapped key from shadowing a protected NVM
+ *   entry, so commit and erase must reject a wrapped key id
+ */
+static int _testWrappedKey_UnwrapCache_NoNvmPersist(
```
bigbrett: Why is this test in the multiclient test group? It only uses one client

bigbrett (changes requested):

test-refactor/misc/wh_test_multiclient.c
padelsbach: I had grouped all the keywrap tests together there. 

I moved the two single client keywrap tests over to the client-server group

bigbrett (approved):

PR #425, opened 2026-06-25
wolfHSM Hardening Fixes

aidangarske:
```
F-6031, F-5475, F-6029, F-6030, F-5473, F-5474
```

### DIFF
--- a/examples/posix/wh_posix_server/wh_posix_server_cfg.c
+++ b/examples/posix/wh_posix_server/wh_posix_server_cfg.c
@@ -493,8 +493,15 @@ static void parseNvmInitFile(const char* filePath)
 
         /* Parse the file path */
         token = strtok(NULL, " ");
-        if (!token || sscanf(token, "%s", filePath) != 1) {
+        if (!token) {
             WOLFHSM_CFG_PRINTF("Error on line %d: Malformed entry - missing file path\n",
+                    lineNumber);
+            fclose(file);
+            exit(EXIT_FAILURE);
+        }
+        if (snprintf(filePath, sizeof(filePath), "%s", token) >=
+            (int)sizeof(filePath)) {
+            WOLFHSM_CFG_PRINTF("Error on line %d: File path too long\n",
                     lineNumber);
             fclose(file);
             exit(EXIT_FAILURE);
--- a/port/posix/posix_transport_shm.c
+++ b/port/posix/posix_transport_shm.c
@@ -208,8 +208,8 @@ static int posixTransportShm_CreateMap(char* name, uint16_t req_size,
 
     /* Attempt to remove any existing object. */
     (void)shm_unlink(name);
-    /* Create shared memory object and set the size */
-    fd = shm_open(name, O_CREAT | O_RDWR, PTSHM_CREATEMODE);
+    /* O_EXCL fails the open if a racing process re-created the name */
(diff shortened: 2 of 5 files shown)


bigbrett (commented):
I'm medium on these force zero changes. In all of our currently supported transports, the transport buffer itself has to be trusted or the whole threat model falls apart, and the client's local comm buffers even more so. Zeroizing these here in light of these assumptions is overkill IMO and would be a bit of a performance hit. I think we need to have a uniform policy about when we require force zeroing of shared memory and apply it across the entire code base at once. Perhaps a build time configuration that would enable "strict" force zeroing where the server just clears the entirety of the comm buffers once received? But again, pretty sure this is overkill. @billphipps @rizlik thoughts?

aidangarske:
> I'm medium on these force zero changes. In all of our currently supported transports, the transport buffer itself has to be trusted or the whole threat model falls apart, and the client's local comm buffers even more so. Zeroizing these here in light of these assumptions is overkill IMO and would be a bit of a performance hit. I think we need to have a uniform policy about when we require force zeroing of shared memory and apply it across the entire code base at once. Perhaps a build time configuration that would enable "strict" force zeroing where the server just clears the entirety of the comm buffers once received? But again, pretty sure this is overkill. @billphipps @rizlik thoughts?

@bigbrett, I am fine with doing it however you want seems like there is a lot of tradeoffs for each. Currently wolfssl does it like this but if we want to make a better design for wolfHSM that sounds good! Ill leave it for now till we decide

Frauschi (dismissed):
Besides the open discussion on the ForceZero policy for CommBuf data, the other changes LGTM.

(Only to give my two cents to the ForceZero discussion: I'm with @bigbrett on this. Have them in, but behind some kind of feature flag.)

rizlik:
> I'm medium on these force zero changes. In all of our currently supported transports, the transport buffer itself has to be trusted or the whole threat model falls apart, and the client's local comm buffers even more so. Zeroizing these here in light of these assumptions is overkill IMO and would be a bit of a performance hit. I think we need to have a uniform policy about when we require force zeroing of shared memory and apply it across the entire code base at once. Perhaps a build time configuration that would enable "strict" force zeroing where the server just clears the entirety of the comm buffers once received? But again, pretty sure this is overkill. @billphipps @rizlik thoughts?

@bigbrett  I do think zeroing is worth even if the comm buffer is considered trusted, being zeroing more a defense in depth against possible client co-resident threats. 

But yes having an performance opt-out guard is also a nice to have but more importantly I second the idea of doing that in a single place, probably when the client receives the response.
The problem I see here is that as this PR does the zeroing there is an implicit assumption that the transport cb send is done with the buffer the moment it returns. 
If this is something we want to lock in the transport API, we should add it in the docs, otherwise, zeroing must happen at response time.

regarding zeroing at the server:  this adds the requirement for a server to be able to write on the request buffer, while until now it can be mapped read-only.

aidangarske:
@bigbrett @rizlik, I added a `WOLFHSM_CFG_DISABLE_CLIENT_COMM_ZEROIZE` macro that is a specific performance opt out macro for the zeroization changes and has client communication-buffer zeroization by default. The cleanup is now centralized and preserve the in place response and DMA handling until parsing completes then the time sensitive fields are cleared . Let me know if this is what you guys where thinking

bigbrett (approved):

PR #428, opened 2026-06-25
CI FIx whTest_NvmDma basline check

aidangarske:
Fixes CI failures


padelsbach:
@aidangarske, I think the fix in #429 is more robust. What do you think?

PR #430, opened 2026-06-26
Add test case for SHE secure boot

padelsbach:
Fixes F-5471

Adds test cases to legacy and refactor for the checks in _ReportInvalidSheState. _ReportInvalidSheState has a tiny refactor and now allows GET_STATUS regardless of state, per SHE spec.

### DIFF
--- a/src/wh_server_she.c
+++ b/src/wh_server_she.c
@@ -1617,16 +1617,32 @@ static int _ReportInvalidSheState(whServerContext* server, uint16_t magic,
 {
     (void)req_packet;
     (void)req_size;
+    int ret = 0;
 
-    /* TODO does SHE specify what this error should be? */
-    /* if we haven't secure booted, only allow secure boot requests */
-    if ((server->she->sbState != WH_SHE_SB_SUCCESS &&
-         (action != WH_SHE_SECURE_BOOT_INIT &&
-          action != WH_SHE_SECURE_BOOT_UPDATE &&
-          action != WH_SHE_SECURE_BOOT_FINISH && action != WH_SHE_GET_STATUS &&
-          action != WH_SHE_SET_UID)) ||
-        (action != WH_SHE_SET_UID && server->she->uidSet == 0)) {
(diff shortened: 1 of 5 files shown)


src/wh_server_she.c
```diff
@@ -1617,16 +1617,32 @@ static int _ReportInvalidSheState(whServerContext* server, uint16_t magic,
...
-         (action != WH_SHE_SECURE_BOOT_INIT &&
-          action != WH_SHE_SECURE_BOOT_UPDATE &&
-          action != WH_SHE_SECURE_BOOT_FINISH && action != WH_SHE_GET_STATUS &&
-          action != WH_SHE_SET_UID)) ||
-        (action != WH_SHE_SET_UID && server->she->uidSet == 0)) {
-        /* Create an error response based on the action */
+    if (action == WH_SHE_GET_STATUS) {
+        /* Status read is always permitted per AUTOSAR spec, even before boot 
+         *or UID setup. */
+    }
```
AlexLanzano, line 1625: Can't we just drop this check entirely?

src/wh_server_she.c
```diff
@@ -1617,16 +1617,32 @@ static int _ReportInvalidSheState(whServerContext* server, uint16_t magic,
 {
     (void)req_packet;
     (void)req_size;
+    int ret = 0;
```
AlexLanzano: This declaration needs to come before the `(void)req_packet;` statement in order to be C90 compliant

AlexLanzano (changes requested):
Looks great! Just two small things. Will merge once addressed

src/wh_server_she.c
padelsbach: Ugh, good catch. Will fix

src/wh_server_she.c
padelsbach: Not without impacting the `else` cases below. I like this style for self-documentation, even if it is a no-op.

AlexLanzano (approved):

PR #450, opened 2026-07-13
Implement SHE GET_ID command with client side request, server side handler, and tests

AlexLanzano:


docs/src/5-Features.md
```diff
@@ -698,6 +699,32 @@ The SHE spec also requires every key to carry a 28-bit monotonic update counter
...
+```c
+whServerSheConfig sheConfig = {
+    .getUidCb = myGetUid,
+    .setUidCb = NULL,        /* fused UID, read-only */
+    .uidCtx   = &myPlatform,
+};
+whServerConfig serverConfig = { /* ... */ .she = she, .sheConfig = &sheConfig };
+```
+
+When a getter is installed the server reads the UID on demand at each point that needs it and keeps no copy in `whServerSheContext`, so the UID never outlives a single request in server RAM. The trade-off is that the getter is called on every gated SHE request, and more than once for commands that use the UID, so it must be cheap, idempotent, and free of side effects. Leaving `sheConfig` unset preserves the original in-context behavior exactly.
```
bigbrett: remove this line

docs/src/5-Features.md
```diff
@@ -698,6 +699,32 @@ The SHE spec also requires every key to carry a 28-bit monotonic update counter
 
 `RAM_KEY` is the one exception to NVM-backed storage. The spec defines it as volatile, so the server caches the loaded key in its [key cache](#key-cache-key-ids-and-nvm-backing-store) but never calls into the NVM layer for it; eviction or reset clears it. All other slots, including `PRNG_SEED`, persist.
 
+### SHE UID Storage
+
+By default the 15-byte ECU UID lives in the caller-owned `whServerSheContext`, which means it is RAM-resident and must be re-provisioned with `CMD_SET_UID` after every reset. Integrators whose UID is fused, held in OTP, or kept in NVM can instead point the server at that store with a pair of callbacks:
```
bigbrett: This reads like the default behavior is that the UID is RAM only, but really there isn't a default it just depends on if you register callbacks or now. I'd rephrase to make it clear that the callback field would be what dictates the behavior.

src/wh_message_she.c
```diff
@@ -429,4 +429,36 @@ int wh_MessageShe_TranslateVerifyMacResponse(
...
+    uint16_t magic, const whMessageShe_GetIdResponse* src,
+    whMessageShe_GetIdResponse* dest)
+{
+    if ((src == NULL) || (dest == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+    WH_T32(magic, dest, src, rc);
+    if (src != dest) {
+        memcpy(dest->uid, src->uid, WH_SHE_UID_SZ);
+        memcpy(dest->mac, src->mac, WH_SHE_KEY_SZ);
```
bigbrett, line 458: would these not need translation?

src/wh_server.c
```diff
@@ -91,6 +91,20 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
     server->devId  = config->devId;
 #ifdef WOLFHSM_CFG_SHE_EXTENSION
     server->she = config->she;
+    /* The SHE context is caller-owned and not covered by the memset above, so
+     * the callbacks are assigned on both paths. */
```
bigbrett:
```suggestion
```

wolfhsm/wh_server_she.h
```diff
@@ -40,6 +40,24 @@
...
+
+/* Persists the WH_SHE_UID_SZ byte UID provisioned by WH_SHE_SET_UID. */
+typedef int (*whServerSheSetUidCb)(struct whServerContext_t* server, void* ctx,
+                                   const uint8_t* uid);
+
+typedef struct {
+    whServerSheGetUidCb getUidCb; /* NULL = use in-context uid[]/uidSet */
+    whServerSheSetUidCb setUidCb; /* NULL = UID is read-only */
+    void*               uidCtx;   /* opaque, passed back to both callbacks */
+} whServerSheConfig;
```
bigbrett, line 59: could we not have the callbacks take a pointer to a `whServerSheContext` instead of `whServerContext` to avoid the impicit `wh_server.h` dependency? In my mind the only server-specific thing that it would need would be a pointer to the NVM instance, which could be passed through the callback-specific context pointer anyway?

src/wh_server_she.c
```diff
@@ -159,6 +166,68 @@ static int _TranslateSheReturnCode(int ret)
     return ret;
 }
 
+/* Reads the UID into outUid. Returns WH_ERROR_NOTFOUND if unprovisioned.
+ * The UID is public, so failure paths clear the buffer for determinism rather
+ * than secrecy and do not need wh_Utils_ForceZero. */
```
bigbrett:
```suggestion
```

src/wh_server_she.c
```diff
@@ -159,6 +166,68 @@ static int _TranslateSheReturnCode(int ret)
...
+    memcpy(outUid, she->uid, WH_SHE_UID_SZ);
+    return 0;
+}
+
+/* Stores a UID. Returns WH_ERROR_NOTIMPL if the UID is read-only. */
+static int _StoreUid(whServerContext* server, const uint8_t* uid)
+{
+    whServerSheContext* she = server->she;
+
+    if (she->getUidCb != NULL) {
```
bigbrett, line 196: why do we need a non-NULL `getUidCb` here?

src/wh_server_she.c
```diff
@@ -159,6 +166,68 @@ static int _TranslateSheReturnCode(int ret)
...
+{
+    uint8_t uid[WH_SHE_UID_SZ];
+    int     ret;
+
+    if (server->she->getUidCb == NULL) {
+        return (server->she->uidSet != 0) ? 1 : 0;
+    }
+
+    ret = _GetUid(server, uid);
+    memset(uid, 0, sizeof(uid));
```
bigbrett, line 219: maybe just make `_GetUid` able to take a NULL output pointer and skip the copy, so we dont need a dummy buffer here?

src/wh_server_she.c
```diff
@@ -597,6 +686,13 @@ static int _LoadKey(whServerContext* server, uint16_t magic, uint16_t req_size,
             }
         }
     }
+    /* fetch the UID once for the M1 comparison and the M4 response */
+    if (ret == 0) {
+        ret = _GetUid(server, uid);
+        if (ret != 0) {
+            ret = WH_SHE_ERC_MEMORY_FAILURE;
+        }
```
bigbrett, line 692: this pattern is repeated a lot, would it make more sense for `_GetUid()` to just return the right error codes so they don't need to be remuxed everywhere?

src/wh_server_she.c
```diff
@@ -1638,6 +1742,86 @@ static int _VerifyMac(whServerContext* server, uint16_t magic,
...
+        memcpy(macIn, req.challenge, WH_SHE_KEY_SZ);
+        memcpy(macIn + WH_SHE_KEY_SZ, uid, WH_SHE_UID_SZ);
+        macIn[WH_SHE_KEY_SZ + WH_SHE_UID_SZ] = sreg;
+
+        /* Load the MASTER_ECU_KEY. Per the SHE spec, if the slot is empty the
+         * MAC is computed with an all-zero key rather than returning an error.
+         * wh_Server_KeystoreReadKey already applies this substitution for the
+         * MASTER_ECU_KEY slot (returns a zero-filled key and rc 0), so the
+         * WH_ERROR_NOTFOUND branch below is only a defensive backstop for
+         * builds/paths where that substitution does not occur. */
```
bigbrett:
```suggestion
```

src/wh_server_she.c
```diff
@@ -1964,6 +2174,14 @@ int wh_Server_HandleSheRequest(whServerContext* server, uint16_t magic,
                 (void)WH_SERVER_NVM_UNLOCK(server);
             } /* WH_SERVER_NVM_LOCK() */
             break;
+        case WH_SHE_GET_ID:
+            ret = WH_SERVER_NVM_LOCK(server);
```
bigbrett, line 2170: why are we taking the lock here?

docs/src/5-Features.md
```diff
@@ -672,10 +672,11 @@ The SHE client API is declared in `wolfhsm/wh_client_she.h` and maps one-to-one
 - **Bulk crypto**: `wh_Client_SheEncEcb` / `wh_Client_SheEncCbc` / `wh_Client_SheDecEcb` / `wh_Client_SheDecCbc` (`CMD_ENC_*` / `CMD_DEC_*`) — AES-ECB and AES-CBC encrypt and decrypt against a selected key slot
 - **MAC**: `wh_Client_SheGenerateMac` / `wh_Client_SheVerifyMac` (`CMD_GENERATE_MAC` / `CMD_VERIFY_MAC`) — CMAC generation and verification against a selected key slot
 - **Status**: `wh_Client_SheGetStatus` (`CMD_GET_STATUS`) — reads the SHE status register (SREG)
+- **Module identity**: `wh_Client_SheGetId` (`CMD_GET_ID`) — returns the ECU UID, the status register, and a CMAC over the caller's challenge, UID, and status register computed under the `MASTER_ECU_KEY`, letting a party that holds that key verify the module's identity. If the `MASTER_ECU_KEY` slot is empty the MAC is computed with an all-zero key, per the spec.
```
bigbrett:
```suggestion
- **Module identity**: `wh_Client_SheGetId` (`CMD_GET_ID`) — returns the ECU UID, the status register, and a CMAC over the caller's challenge, UID, and status register computed under the `MASTER_ECU_KEY`, letting a party that holds that key verify the module's identity. If the `MASTER_ECU_KEY` slot is empty the MAC is computed with an all-zero key.
```

docs/src/5-Features.md
```diff
@@ -736,6 +763,8 @@ SHE secure boot is implemented as a three-phase state machine that the client dr
 
 While the state machine is in any state other than `SUCCESS`, the SHE handler refuses every non-boot command except `CMD_GET_STATUS` and `CMD_SET_UID`, returning `WH_SHE_ERC_SEQUENCE_ERROR`. This is what allows the SHE module to gate cryptographic services on a successful boot measurement: once boot has succeeded, the rest of the SHE command set unlocks; on a boot failure the keys remain inaccessible and only status queries are honored.
 
+The same gate also enforces that a UID has been provisioned. When [UID storage callbacks](#she-uid-storage) are installed, that check queries the integrator's store, and a store that reports a failure causes every command except `CMD_GET_STATUS` to return `WH_SHE_ERC_MEMORY_FAILURE`. Status stays readable in every case, per the spec.
```
bigbrett:
```suggestion
The same stateful gate also enforces that a UID has been provisioned. When [UID storage callbacks](#she-uid-storage) are installed, that check queries the integrator's store, and a store that reports a failure causes every command except `CMD_GET_STATUS` to return `WH_SHE_ERC_MEMORY_FAILURE`. 
```

bigbrett (changes requested):
Run `git-clang-format main` to reformat the diff plz

src/wh_server_she.c
AlexLanzano: To sync the KeyStoreRead of the MASTER_ECU_KEY.

src/wh_message_she.c
AlexLanzano: These are accessed/used as byte arrays on both ends so there wouldn't be any need to translate

src/wh_server_she.c
AlexLanzano: If getUidCb is set then it's safe to assume that the the RAM based uid storage isn't being used and shouldn't be fell back on.

src/wh_server_she.c
AlexLanzano: My only concern with this is that I can see users overlooking that a NULL output pointer is an acceptable arg when implementing the getUidCb.

src/wh_server_she.c
AlexLanzano: I think it's best that the underling caller of the `_GetUid` determines the SHE error code else we will need to pass in some sort of state to the _GetUid function in order to determine the right SHE error code. I.E. take a look at `_UidIsProvisioned` which calls _GetUid and how it's caller determines the right SHE error code

wolfhsm/wh_server_she.h
AlexLanzano: I feel like at that point we should just have the signature be:
`int (*whServerSheGetUidCb)(void* ctx, uint8_t* outUid)`

Thoughts?

bigbrett: yeah thats fair, if we don't have any built-in need for that context field, maybe it can just be voided

src/wh_server_she.c
bigbrett: I meant the _GetUid function, not the callback

AlexLanzano: _GetUid is just a fancy wrapper for the callback so changing _GetUid would just move the dummy uid buffer into _GetUid.

I ended up just allowing the callback to accept a null output pointer and documenting it.

wolfhsm/wh_server_she.h
AlexLanzano: Fixed

src/wh_server_she.c
AlexLanzano: Guessing this is resolved?

bigbrett (approved):

PR #452, opened 2026-07-14
Fix ML-DSA key cache size check for verify-only builds

miyazakh:
## Summary

When `WOLFSSL_DILITHIUM_VERIFY_ONLY` is defined, the server only needs to
cache the public key DER, not the full keypair. The previous static assert
unconditionally required the cache buffer to fit a full keypair DER, causing
unnecessary failures in verify-only configurations.

This change introduces `WH_SERVER_MLDSA_MAX_CACHE_DER_SIZE` to select the
correct maximum size at compile time:

- `WOLFSSL_DILITHIUM_VERIFY_ONLY` defined: `MLDSA_MAX_PUB_KEY_DER_SIZE`
- Otherwise: `MLDSA_MAX_BOTH_KEY_DER_SIZE`

The static assert is updated to use this macro so it reflects the actual
requirement for each build configuration.

## Testing

- Build with and without `WOLFSSL_DILITHIUM_VERIFY_ONLY` and confirm the
  static assert passes in both cases.


src/wh_server_crypto.c
```diff
@@ -768,11 +768,19 @@ int wh_Server_CacheExportCurve25519Key(whServerContext* server, whKeyId keyId,
...
 #ifdef WOLFSSL_HAVE_MLDSA
-/* The big key cache buffer must be able to hold a full ML-DSA keypair DER,
- * otherwise wh_Server_MlDsaKeyCacheImport() can never succeed. */
+/* When verify-only, the server caches only the public key DER. Otherwise it
+ * must be able to hold a full keypair DER (public + private). */
+#ifdef WOLFSSL_DILITHIUM_VERIFY_ONLY
+#define WH_SERVER_MLDSA_MAX_CACHE_DER_SIZE MLDSA_MAX_PUB_KEY_DER_SIZE
+WH_UTILS_STATIC_ASSERT(
+    WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE >= MLDSA_MAX_PUB_KEY_DER_SIZE,
+    "WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE too small for ML-DSA public key DER");
```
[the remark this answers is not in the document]
miyazakh: Fixed

src/wh_server_crypto.c
```diff
@@ -768,11 +768,19 @@ int wh_Server_CacheExportCurve25519Key(whServerContext* server, whKeyId keyId,
...
+#else
+#define WH_SERVER_MLDSA_MAX_CACHE_DER_SIZE MLDSA_MAX_BOTH_KEY_DER_SIZE
 WH_UTILS_STATIC_ASSERT(
     WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE >= MLDSA_MAX_BOTH_KEY_DER_SIZE,
     "WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE too small for ML-DSA keypair DER");
```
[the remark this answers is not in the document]
miyazakh: Fixed

padelsbach (changes requested):
Looks good, but can you please add simple test cases to build-and-test.yml and build-and-test-refactor.yml

padelsbach:
Agree with the Fenrir finding. Please correct uses of `WH_SERVER_MLDSA_MAX_CACHE_DER_SIZE`.

src/wh_server_crypto.c
padelsbach, line 4942: Can you add the `(void)devId;` here and around line 6632?

.github/workflows/build-and-test-refactor.yml
```diff
@@ -53,6 +53,10 @@ jobs:
     - name: Build and test refactor DMA ASAN LMS verify-only XMSS full
       run: cd test-refactor/posix && make clean && make -j DMA=1 ASAN=1 LMS_VERIFY_ONLY=1 WOLFSSL_DIR=../../wolfssl && make run
 
+    # Build and test with ML-DSA in verify-only mode
+    - name: Build and test refactor DMA ASAN MLDSA verify-only
+      run: cd test-refactor/posix && make clean && make -j DMA=1 ASAN=1 MLDSA_VERIFY_ONLY=1 WOLFSSL_DIR=../../wolfssl && make run
```
padelsbach, line 58: To exercise the fix, please set `WOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE ` to less than `MLDSA_MAX_BOTH_KEY_DER_SIZE` (7520). 

iiuc, this buffer reduction is one of the main benefits of verify-only

src/wh_server_crypto.c
miyazakh: Added

.github/workflows/build-and-test-refactor.yml
miyazakh: Makefiles: added `KEYCACHE_BIG_BUFSIZE` variable

Both `test/Makefile` and `test-refactor/posix/Makefile` now accept:

```
KEYCACHE_BIG_BUFSIZE=<N>
```

which passes `-DWOLFHSM_CFG_SERVER_KEYCACHE_BIG_BUFSIZE=<N>` to the compiler.

CI: added reduced-keycache build step

Added a step to both `build-and-test.yml` and `build-and-test-refactor.yml`:

```
MLDSA_VERIFY_ONLY=1 KEYCACHE_BIG_BUFSIZE=4096
```

`4096 < MLDSA_MAX_BOTH_KEY_DER_SIZE (7520)`, so the static assert would fail
without `MLDSA_VERIFY_ONLY`. With verify-only, only the public key DER needs to
fit (`MLDSA_MAX_PUB_KEY_DER_SIZE = 2614 < 4096`), so the assert passes.

padelsbach:
@miyazakh there was a github outage earlier which caused test failures. Can you push again to re-trigger CI?

miyazakh:
@padelsbach 
It seems that all the tests passed after manual re-testing.

padelsbach (dismissed):

padelsbach:
LGTM, will give @bigbrett a chance for a final review.

.github/workflows/build-and-test.yml
```diff
@@ -53,6 +53,17 @@ jobs:
...
+    # Build and test with ML-DSA in verify-only mode
+    - name: Build and test DMA ASAN MLDSA verify-only
+      run: cd test && make clean && make -j DMA=1 ASAN=1 MLDSA_VERIFY_ONLY=1 WOLFSSL_DIR=../wolfssl && make run
+
+    # Build and test with ML-DSA verify-only and a reduced keycache buffer.
+    # 4096 < MLDSA_MAX_BOTH_KEY_DER_SIZE(7520) so this would fail the static
+    # assert without MLDSA_VERIFY_ONLY; it passes because verify-only only
+    # needs MLDSA_MAX_PUB_KEY_DER_SIZE(2614). LMS/XMSS full keys fit in 4096.
+    - name: Build and test DMA ASAN MLDSA verify-only reduced keycache
+      run: cd test && make clean && make -j DMA=1 ASAN=1 MLDSA_VERIFY_ONLY=1 KEYCACHE_BIG_BUFSIZE=4096 WOLFSSL_DIR=../wolfssl && make run
```
bigbrett: I appreciate the spirit of adding test cases for both of these but IMO its pretty overkill to add two separate builds and expose two internal parameters just to essentially verify at compile time.I'd prefer to skip this second one that verifies the static assert hits just for the sake of conserving CI time.IMO something like this is better suited for a nightly (which we don't have yet.

bigbrett (changes requested):
overall looks good, I think one of the test cases should be removed as it is overkill.

.github/workflows/build-and-test.yml
miyazakh: Removed the reduced-keycache CI steps. The KEYCACHE_BIG_BUFSIZE Makefile variable is kept so it can be used for manual verification or a future nightly run.

bigbrett (approved):

PR #456, opened 2026-07-16
Use enum size instead of MAX_BENCH_OPS

padelsbach:
**NOTE:** this change switches wh_bench.c from CRLF to LF. Turn off the whitespace comparison when reviewing the diff for sanity.

This change removes `MAX_BENCH_OPS` which needed to be adjusted whenever the test list was updated. Instead, just use the size of the enum.


Frauschi (approved):
LGTM

bigbrett (approved):

PR #458, opened 2026-07-17
Return a generated public key from cached keygen

Frauschi:
### Summary

Cached key generation previously returned only a `keyId` — to get the public key of a freshly generated cached key, the client had to make a second `ExportPublicKey` round-trip. This PR adds the new LMS/XMSS ergonomic (keygen also hands back the public key in one round-trip, introduced in #352) to the remaining public-key algorithms: **RSA, ECC, Curve25519, Ed25519, ML-DSA, and ML-KEM**.

The server serializes the public-only key into the **existing** keygen response body, so there are no new wire structures and the change is fully backward-compatible. Existing `MakeCacheKey` callers keep working unchanged.

### What changed

**New client API** (`wolfhsm/wh_client_crypto.h`) — opt-in wrappers that cache the key *and* return its public part into the caller's key object, stamping the cached `keyId` on it:

- `wh_Client_RsaMakeCacheKeyAndExportPublic(...)`
- `wh_Client_EccMakeCacheKeyAndExportPublic(...)`
- `wh_Client_Curve25519MakeCacheKeyAndExportPublic(...)`
- `wh_Client_Ed25519MakeCacheKeyAndExportPublic(...)`
- `wh_Client_MlDsaMakeCacheKeyAndExportPublic(...)` + `wh_Client_MlDsaMakeCacheKeyDma(...)`
- `wh_Client_MlKemMakeCacheKeyAndExportPublic(...)` + `wh_Client_MlKemMakeCacheKeyDma(...)`

**Server** (`src/wh_server_crypto.c`) — on the cache path, serializes the public-only key into the keygen response body using the same helpers as the keystore export path (`wc_RsaKeyToPublicDer`, `wc_EccPublicKeyToDer(...,1)`, `wc_Curve25519PublicKeyToDer(...,1)`, `wc_Ed25519PublicKeyToDer(...,1)`, `wc_MlDsaKey_PublicKeyToDer(...,1)`, `wc_MlKemKey_*`). DMA paths reuse the `key` buffer + `keySize`.

**Client** (`src/wh_client_crypto.c`) — the new wrappers parse the public key from the response body and set the returned `keyId`.

### Design notes

- **Wire-compatible / backward-compatible.** Reuses the existing keygen response body (`keyId` + `len` + trailing DER); no new request/response structs, no translate or padding-test changes. Callers of the existing `MakeCacheKey` functions ignore the extra bytes. Existing public API signatures are untouched.

- **Fail-closed export on the server.** The server always serializes the public key into the cache-path response body. A freshly generated key is expected to serialize, so a failure is treated as fatal: the server evicts the just-committed key and returns the error rather than handing back a `keyId` for a key with no public part — no cache slots are orphaned.

- **Strict contract on the client.** The `...AndExportPublic` wrappers return `WH_ERROR_ABORTED` if the server returns an empty body, while still propagating the `keyId` so the caller can evict the cached key.

- **CryptoCb deliberately unchanged.** `wc_MakeRsaKey` and friends stay ephemeral (the full key is returned to the client). Cached keygen keys have no auto-eviction (wolfCrypt does not route `wc_*_free` through the CryptoCb), so routing keygen through the cache would leak HSM cache slots on a make/use/free loop and change durability semantics. The new behavior is reachable only via the native `MakeCacheKeyAndExportPublic` functions.

- ML-DSA public-key serialization is guarded by `WOLFSSL_MLDSA_PUBLIC_KEY`.

### Testing

Tests were added to **both** suites:

- `test/wh_test_crypto.c` (main suite) — all six algorithms, including ML-KEM.
- `test-refactor/client-server/wh_test_crypto_{rsa,ecc,curve25519,ed25519,mldsa}.c` (the refactor suite has no ML-KEM file yet, so ML-KEM is main-suite only).

Each test cross-checks the keygen-returned public key against `wh_Client_<Algo>ExportPublicKey` (byte-equal) and proves it with a real operation — encrypt/HSM-decrypt, HSM-sign/client-verify, or ECDH round-trip. A negative test covers the `EPHEMERAL` + NULL-arg contract.

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -81,7 +81,7 @@
  * and the blocking wrappers wh_Client_EccMakeCacheKey/EccMakeExportKey. */
 static int _EccMakeKeyRequest(whClientContext* ctx, int size, int curveId,
                               whKeyId key_id, whNvmFlags flags,
-                              uint16_t label_len, uint8_t* label);
+                              uint16_t label_len, const uint8_t* label);
 static int _EccMakeKeyResponse(whClientContext* ctx, whKeyId* out_key_id,
                                ecc_key* out_key);
 #endif /* HAVE_ECC */
@@ -1931,7 +1931,7 @@ int wh_Client_EccExportPublicKey(whClientContext* ctx, whKeyId keyId,
  * blocking poll) to consume the reply. */
 static int _EccMakeKeyRequest(whClientContext* ctx, int size, int curveId,
                               whKeyId key_id, whNvmFlags flags,
-                              uint16_t label_len, uint8_t* label)
+                              uint16_t label_len, const uint8_t* label)
 {
     whMessageCrypto_EccKeyGenRequest* req     = NULL;
     uint8_t*                          dataPtr = NULL;
@@ -2021,14 +2021,17 @@ static int _EccMakeKeyResponse(whClientContext* ctx, whKeyId* out_key_id,
             *out_key_id = key_id;
         }
 
-        /* If a DER blob is present (ephemeral/export keygen), deserialize it
-         * into the supplied key context.  When called from the cache wrapper,
-         * out_key is NULL and the server returns an empty body. */
+        /* A DER blob is present for both ephemeral/export keygen and the
+         * cache-and-export path. When key material was requested (out_key !=
+         * NULL) but the server returned an empty body, treat it as an error so
+         * a caller of ...AndExportPublic never receives an unpopulated key. */
         if (out_key != NULL) {
             if (res->len > 0) {
                 uint8_t* key_der  = (uint8_t*)(res + 1);
                 uint16_t der_size = (uint16_t)(res->len);
-                /* Ephemeral key — leave devCtx ERASED */
+                /* Leave devCtx ERASED here: ephemeral keygen keeps it that way,
+                 * and the cache-and-export wrapper overwrites it with the cached
+                 * keyId after this returns. */
                 wh_Client_EccSetKeyId(out_key, WH_KEYID_ERASED);
                 ret =
                     wh_Crypto_EccDeserializeKeyDer(key_der, der_size, out_key);
@@ -2036,8 +2039,7 @@ static int _EccMakeKeyResponse(whClientContext* ctx, whKeyId* out_key_id,
                                          der_size);
             }
             else {
-                /* Cached key — stamp the assigned keyId */
-                wh_Client_EccSetKeyId(out_key, key_id);
+                ret = WH_ERROR_ABORTED;
             }
         }
     }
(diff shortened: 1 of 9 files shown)


src/wh_client_crypto.c
```diff
@@ -3051,6 +3111,44 @@ int wh_Client_Curve25519MakeCacheKey(whClientContext* ctx, uint16_t size,
...
+    whKeyId in_keyId;
+
+    if ((ctx == NULL) || (inout_key_id == NULL) || (pub == NULL)) {
+        return WH_ERROR_BADARGS;
+    }
+
+    /* Ephemeral keygen belongs to the export path, not the cache path. */
+    if (flags & WH_NVM_FLAGS_EPHEMERAL) {
+        return WH_ERROR_BADARGS;
+    }
```
bigbrett, line 3136: Do we actually need this check? Here an everywhere else in this file.

src/wh_server_crypto.c
```diff
@@ -375,9 +375,25 @@ static int _HandleRsaKeyGen(whServerContext* ctx, uint16_t magic, int devId,
...
+                     * failure as fatal: evict the just-committed key and
+                     * propagate the error rather than returning a keyId with no
+                     * public key. */
+                    int pub_ret = wc_RsaKeyToPublicDer(rsa, out, max_size);
+                    if (pub_ret > 0) {
+                        der_size = (uint16_t)pub_ret;
+                    }
+                    else {
+                        (void)wh_Server_KeystoreEvictKey(ctx, key_id);
+                        ret = (pub_ret < 0) ? pub_ret : WH_ERROR_ABORTED;
```
bigbrett: So we actually need this ternary assignment to propagate the error? This pattern is in a lot of places in this file, but do we really need to handle the `pub_ret==0` case separate from just propagating the return value of `pub_ret` at this point?

bigbrett (dismissed):
LGTM, flagged a few random nits I'm totally fine with you skipping if not important. @rizlik over to you.

src/wh_client_crypto.c
Frauschi: I'd keep it. It ensures that the user does not _accidentally_ pass the `WH_NVM_FLAGS_EPHEMERAL` flag, still expecting a cached key. If the flag is passed, the key is not stored on the server at all, so that would break the API-contract.

src/wh_server_crypto.c
Frauschi: I'd keep it, as `pub_ret==0` can be interpreted as `SUCCESS`, although a zero sized DER is not really a valid public key. This will probably never happen in real life, so it's merely defense-in-depth.

rizlik (commented):
My only doubt is that the ML-DSA MakeCacheKey response now carries the public key, which may not fit in the comm buffer, making the API fail on constrained devices. Those devices can use the new DMA make-and-export API, but it's probably worth a comment somewhere so that users know how to fix this when they upgrade.
Beside that, LGTM

Frauschi:
@rizlik As discussed offline, I modified the logic to make the public key export "best effort". The server tries to store the public key in the response buffer. If the buffer is too small, no public key is stored there, but the response is sent anyway. On the client, the response is parsed. For existing `MakeCacheKey()` callers, both cases (public key present and no public key present as buffer-too-small) result in the same existing behavior to not cause any regressions. In the new `MakeCacheKeyAndExportPublic()` API, a missing public key in the response is handled as an error (evict the generated key from the server and return the error to the caller).

rizlik:
LGTM. @bigbrett are you OK with the new best-effor behaviour?

rizlik (approved):

bigbrett (approved):

PR #460, opened 2026-07-20
Nvm ids per client

bigbrett:
Refactors NVM layer to add per-client NVM id translation for all client-facing requests. Client NVM objecs are no longer global identifiers, but now are scoped similar to existing keyIds: every client gets 1-255 unique objects dedicated to them, with 1-255 global objects shared across clients.

It also adds enforcement for an additional server-side rule: any NVM object used for something by the server, either internally or in response to a client API (e.g. certificate API) now MUST be classified by their on NVM object type. This prevents maliciously misusing the cert API to name non-cert objects.

It also introduces some test-only SHE functionality such that the SHE layer remains compatible. 

The bulk of this PR is testing code.


src/wh_server_nvm.c
```diff
@@ -342,7 +453,8 @@ int wh_Server_HandleNvmRequest(whServerContext* server,
             rc = WH_SERVER_NVM_LOCK(server);
             if (rc == WH_ERROR_OK) {
                 rc = _HandleNvmRead(server, data, req.offset, req.data_len,
-                                    &req.data_len, req.id);
+                                    &req.data_len,
+                                    _NvmTranslateFromClient(server, req.id));
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_server_nvm.c
```diff
@@ -301,11 +404,19 @@ int wh_Server_HandleNvmRequest(whServerContext* server,
                     (whMessageNvm_DestroyObjectsRequest*)req_packet, &req);
 
             if (req.list_count <= WH_MESSAGE_NVM_MAX_DESTROY_OBJECTS_COUNT) {
+                whNvmId
+                    translated_ids[WH_MESSAGE_NVM_MAX_DESTROY_OBJECTS_COUNT];
+                whNvmId i;
+                for (i = 0; i < req.list_count; i++) {
+                    translated_ids[i] =
+                        _NvmTranslateFromClient(server, req.list[i]);
+                }
```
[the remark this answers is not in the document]
bigbrett: fixed

src/wh_server_nvm.c
```diff
@@ -217,13 +308,15 @@ int wh_Server_HandleNvmRequest(whServerContext* server,
             rc = WH_SERVER_NVM_LOCK(server);
             if (rc == WH_ERROR_OK) {
                 /* Process the getmetadata action */
-                rc = wh_Nvm_GetMetadata(server->nvm, req.id, &meta);
+                rc = wh_Nvm_GetMetadata(server->nvm,
+                                        _NvmTranslateFromClient(server, req.id),
+                                        &meta);
```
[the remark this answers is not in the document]
bigbrett: fixed

.github/workflows/build-and-test-clientonly.yml
```diff
@@ -49,14 +49,18 @@ jobs:
         config-hash: ${{ matrix.transport }}
         read-only: ${{ github.event_name == 'pull_request' }}
 
-    # Build example server
+    # Build example server. The SHE client tests use the test-only
+    # preprogram/destroy key-management messages, so the CI server must be
+    # built with WOLFHSM_CFG_SHE_ENABLE_TEST_KEY_MGMT (never enable this in
+    # production builds).
```
bigbrett: comment slop

src/wh_keyid.c
```diff
@@ -55,6 +56,53 @@ whKeyId wh_KeyId_TranslateFromClient(uint16_t type, uint16_t clientId,
...
+}
+
+int wh_KeyId_CheckClientObjectId(whKeyId reqId)
+{
+    /* Bits above the id and client-flag fields would be dropped by
+     * translation, remapping the request onto a different object, so a
+     * legacy-style 16-bit id must fail loudly instead */
+    if ((reqId & (whKeyId) ~(WH_KEYID_MASK | WH_CLIENT_KEYID_FLAGS_MASK)) !=
+        0) {
+        return WH_ERROR_BADARGS;
```
billphipps, line 79: Since this is a new behavior for a legacy interface, consider a new error message like DEPRECATED?

src/wh_server.c
```diff
@@ -282,17 +292,16 @@ static int _wh_Server_HandleCommRequest(whServerContext* server,
...
-        if (req.client_id > WH_CLIENT_ID_MAX) {
+        /* client_id 0 (USER=0) names the shared/global and factory-provisioned
+         * namespace, not a client, so it is never a valid connection id. The
+         * client library already refuses it; enforce it here too so a raw
+         * client cannot bind id 0 and reach that namespace through the
+         * per-client key and NVM APIs. */
+        if ((req.client_id == WH_KEYUSER_GLOBAL) ||
+            (req.client_id > WH_CLIENT_ID_MAX)) {
             *out_resp_size = 0;
             return WH_ERROR_BADARGS;
```
billphipps, line 303: Consider ABORTED here.  To me, BADARGS means a coding error, like assert.  In this case, the client somehow passed an invalid client id.

src/wh_server_cert.c
```diff
@@ -878,6 +878,30 @@ int wh_Server_CertVerifyAcert(whServerContext* server, const uint8_t* cert,
...
+ * refuses id 0 and, without global keys, the GLOBAL flag. This confines every
+ * client cert verb to certificate objects, so a client cannot reach a key,
+ * counter, or server-internal object (e.g. the auth user database) through a
+ * cert verb. Server-internal callers of the wh_Server_Cert* API pass full
+ * internal ids directly and are unaffected. */
+static int _CertTranslateFromClient(whServerContext* server, whNvmId reqId,
+                                    int create, whNvmId* out_id)
+{
+    int rc = create ? wh_KeyId_CheckClientObjectIdForCreate(reqId)
+                    : wh_KeyId_CheckClientObjectId(reqId);
```
billphipps, line 897: Nit: consider more verbose if/else instead of the shorthand.  not sure we do this anywhere else in the code.

src/wh_server_cert.c
```diff
@@ -990,18 +1020,22 @@ int wh_Server_HandleCertRequest(whServerContext* server, uint16_t magic,
...
-             * wh_Server_CertReadTrusted() does an unchecked NVM read. */
-            rc = WH_SERVER_NVM_LOCK(server);
+            /* Deny reading a non-exportable or server-only cert object. The id
+             * is confined to the cert namespace above, so this no longer
+             * guards against reaching a key; it enforces per-object export
+             * policy within the cert store. wh_Server_CertReadTrusted() does an
+             * unchecked NVM read, so this is the only gate. */
+            if (rc == WH_ERROR_OK) {
+                rc = WH_SERVER_NVM_LOCK(server);
+            }
```
billphipps, line 1038: Unexpected close } here.  Consider moving this down and spacing the next few sunblocks to improve readability.

src/wh_server_counter.c
```diff
@@ -68,22 +68,29 @@ int wh_Server_HandleCounter(whServerContext* server, uint16_t magic,
...
-            meta->id = WH_MAKE_KEYID(WH_KEYTYPE_COUNTER,
-                                     (uint16_t)server->comm->client_id,
-                                     (uint16_t)req.counterId);
-            /* use the label buffer to hold the counter value */
-            *counter = req.counter;
-
-            ret = WH_SERVER_NVM_LOCK(server);
+            /* Refuse an id that translation would silently alias before
+             * creating anything */
+            ret = wh_KeyId_CheckClientObjectIdForCreate(req.counterId);
```
billphipps, line 73: Is there a concurrency issue with checking the object id without holding the nvm_lock?

src/wh_server_nvm.c
```diff
@@ -48,6 +49,37 @@
 #include "wolfhsm/wh_crypto.h"
 #endif
 
+/* Translate a client-supplied NVM id to the server-internal TYPE/USER/ID
+ * encoding. When WOLFHSM_CFG_LEGACY_CLIENT_NVM is defined, the id is passed
+ * through verbatim (legacy global-flat behavior). */
+static whNvmId _NvmTranslateFromClient(whServerContext* server,
+                                       whNvmId          clientId)
+{
+#ifdef WOLFHSM_CFG_LEGACY_CLIENT_NVM
```
billphipps, line 58: Is this CFG going to be deprecated?  Could this test be put within translateobjectfromclient?

wolfhsm/wh_keyid.h
```diff
@@ -152,6 +152,57 @@ typedef uint16_t whKeyId;
...
+ * have a fixed type: NVM objects, counters, and certificates. Keys use the
+ * plain translator, where the wrapped/hardware flags are meaningful sub-types.
+ *
+ * @param type    Fixed object TYPE to stamp (e.g. WH_KEYTYPE_COUNTER)
+ * @param clientId Connection's client id for the USER field
+ * @param reqId   Requested id from the client (GLOBAL flag honored; wrapped and
+ *  hardware flags ignored)
+ * @return Server-internal keyId with the fixed TYPE, resolved USER, and ID.
+ */
+whKeyId wh_KeyId_TranslateObjectFromClient(uint16_t type, uint16_t clientId,
```
billphipps, line 173: Consider naming ...TranslateObject**Id**From... ?

billphipps (commented):
Wow!!  Looks great!  I only checked headers and src.  Lemme know when you want another review.  Minor comments

wolfhsm/wh_keyid.h
bigbrett: OK can do

src/wh_server_cert.c
bigbrett: will fix

src/wh_server_cert.c
bigbrett: we do it in a lot of places elsewhere, but I agree in this case in particular it would be more readable to have if/else. Will fix

src/wh_server.c
bigbrett: Hmm not sure I agree, that is definitely not how `BADARGS` is used everywhere else in the code. Pretty much everywhere we use it as "caller/client passed invalid parameters to function"

src/wh_keyid.c
bigbrett: IMO this doesn't add much clarity to a user who could have done this accidentally with just too large of an ID val. Legacy seems to muddy the waters. I think badargs is a more straightforward way to convey this to the caller without requiring them to deduce that it is a legacy behavior. The bottom line would just be that the ID is invalid.

src/wh_server_counter.c
bigbrett: nope shoudln't b an issue. The check is a pure function of the request value  and reads nothing but the arg passed by the client.

src/wh_server_nvm.c
bigbrett: yes eventually it would be deprecated I imagine. IMO moving it to `translateObjectFromClient`, we specifically don't want it to apply to other non-NVM APIs so not sure  it has a meaning outside this layer

wolfhsm/wh_keyid.h
bigbrett: OK can do

src/wh_server_nvm.c
bigbrett: yes eventually it would be deprecated I imagine. IMO moving it to translateObjectFromClient, we specifically don't want it to apply to other non-NVM APIs so not sure it has a meaning outside this layer

src/wh_server_counter.c
bigbrett: nope shoudln't b an issue. The check is a pure function of the request value and reads nothing but the arg passed by the client.

src/wh_server_cert.c
bigbrett: will fix

src/wh_server_cert.c
bigbrett: we do it in a lot of places elsewhere, but I agree in this case in particular it would be more readable to have if/else. Will fix

src/wh_server.c
bigbrett: Hmm not sure I agree, that is definitely not how BADARGS is used everywhere else in the code. Pretty much everywhere we use it as "caller/client passed invalid parameters to function"

src/wh_keyid.c
bigbrett: IMO this doesn't add much clarity to a user who could have done this accidentally with just too large of an ID val. Legacy seems to muddy the waters. I think badargs is a more straightforward way to convey this to the caller without requiring them to deduce that it is a legacy behavior. The bottom line would just be that the ID is invalid.

PR #464, opened 2026-07-21
Validate ML-DSA keygen response against received length

yosuke-wolfssl:
## Summary

Two ML-DSA client response handlers consumed server-reported lengths without
checking them against the frame that actually arrived.

`_MlDsaMakeKey` recorded the response length from `wh_Client_RecvResponse` but
never used it. It read `res->keyId` and `res->len` and, on the ephemeral export
path, deserialized `res->len` bytes from `(res + 1)` without confirming the
frame carried that payload. A short but otherwise well-formed OK response
therefore made the client consume whatever trailed the response body in its own
packet buffer.

`wh_Client_MlDsaSignDma` had the same gap: it stored `res->sigLen` into the
caller's `*out_len` without checking that the frame was long enough to contain
the field, or that the reported length fit the caller's buffer.

ECC, RSA and ML-KEM keygen already validate this; ML-DSA was the last keygen
path without it. `_MlDsaMakeKey` backs both `wh_Client_MlDsaMakeCacheKey` and
`wh_Client_MlDsaMakeExportKey`, and the export wrapper is reached from the PQC
signature crypto callback.

Fixes F_5462.

## Changes

**`src/wh_client_crypto.c`**

Keygen — bounds check matching the existing ECC/RSA/ML-KEM condition verbatim:

```c
if (res_len < hdr_sz || res->len > (res_len - hdr_sz)) {
    return WH_ERROR_ABORTED;
}
```

Placed up front, before any response field is consumed, so the `res->keyId` read
on the cache path is covered too. Cache responses carry no payload (`len == 0`,
frame == `hdr_sz`) and still pass.

DMA sign — the response carries no inline payload, so the guard checks the frame
length and the reported signature length against the caller's capacity. That
capacity is stashed in `sig_cap` before the request is sent, because the
response overwrites the request struct it was read from.

**`test-refactor/misc/wh_test_client_malformed_resp.c`** (new) — negative tests.
These APIs send and receive in one blocking call, so they cannot be driven by a
raw comm server on a single thread. The test installs a scripted
`whTransportClientCb` that echoes the request comm header and returns a response
whose reported length is set independently of the frame length it delivers.

- Keygen: a well-formed cache keygen (positive control, also poisons the packet
  buffer), a headers-only frame claiming 256 DER bytes, and a frame shorter than
  the crypto headers.
- DMA sign (`WOLFHSM_CFG_DMA` only): a complete reply that fits, a frame one byte
  short of the response, and a complete frame reporting a signature larger than
  the caller's buffer.

All malformed cases must return `WH_ERROR_ABORTED`.

**`test-refactor/wh_test_list.c`** — registered in the Misc group.

## Verification

Both guards were negative-controlled: with the check neutered the corresponding
subtest fails, and passes once restored.

| build | result |
|---|---|
| `make check` | `38 passed, 25 skipped, 0 failed of 63` |
| `make check DMA=1` | `42 passed, 21 skipped, 0 failed of 63` |

The DMA sign subtest is compiled out without `DMA=1`, so the `DMA=1` run is the
one that exercises it. With the ML-DSA guard forced off, the entry point falls
through to the weak `WH_TEST_DECL` stub and reports SKIPPED rather than an empty
pass.

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -9515,6 +9515,15 @@ static int _MlDsaMakeKey(whClientContext* ctx, int size, int level,
                     /* wolfCrypt allows positive error codes on success in some
                      * scenarios */
                     if (ret >= 0) {
+                        const size_t hdr_sz =
+                            sizeof(whMessageCrypto_GenericResponseHeader) +
+                            sizeof(*res);
+
+                        /* res->len must fit within the received frame */
+                        if (res_len < hdr_sz || res->len > (res_len - hdr_sz)) {
+                            return WH_ERROR_ABORTED;
+                        }
+
                         WH_DEBUG_CLIENT_VERBOSE(
                             "Res recv:keyid:%u, len:%u, ret:%d\n",
                             (unsigned int)res->keyId,
@@ -10106,6 +10115,7 @@ int wh_Client_MlDsaSignDma(whClientContext* ctx, const byte* in, word32 in_len,
     uint8_t*                              dataPtr = NULL;
     uintptr_t                             inAddr  = 0;
     uintptr_t                             outAddr = 0;
+    word32                                sig_cap = 0;
 
     /* Transaction state */
     whKeyId key_id;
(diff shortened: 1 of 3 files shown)


yosuke-wolfssl:
Issue 5462 was fixed by Commit 35bfcf1 in parallel.
Close this.

PR #468, opened 2026-07-21
Translate SHA-2 intermediate state for cross-endian peers

yosuke-wolfssl:
### Problem
wolfHSM supports a client and server on cores with different byte order: every
message header carries an endian marker and the server byte-swaps each field via
`wh_Translate*`. The SHA-2 wire structs declare the resume state as
`uint8_t hash[32]`/`[64]`, and the translation helpers skipped it as "just a byte
array". But that array is a verbatim copy of `wc_Sha256.digest` /
`wc_Sha512.digest`, which wolfCrypt keeps as host-order `word32`/`word64`. A
cross-endian pair therefore exchanged byte-flipped chaining words: every
non-final update compressed from garbage and the digest was silently wrong — no
error, no crash. Final digests are unaffected (wolfCrypt emits canonical
big-endian at finalize), so only multi-message streams broke, and same-endian
pairs never saw it. SHA-3 was already correct: `whMessageCrypto_Sha3State` is
`uint64_t s[25]`, swapped per word.

Addressed by Fenrir 4041.

### Fix (`src/wh_message_crypto.c`)
New `_TranslateSha2State()` swaps the eight chaining words, going through a local
word via `memcpy` (no alignment or aliasing hazard, safe when `src == dest`).

The response direction is ambiguous — `hash[]` holds host-order state on an
update but a canonical digest on the last block, and the struct has no
`isLastBlock` — so the caller now states which it is. The existing response
symbols remain as wrappers passing `0`; no public signature changed.

| Translator | `stateWordSize` |
|---|---|
| `...Sha256Request`, `...Sha256DmaRequest` | `4`, unconditional |
| `...Sha512Request`, `...Sha512DmaRequest` | `8`, unconditional |
| `...Sha2Response_ex`, `...Sha2DmaResponse_ex` (new) | from caller; server passes `req.isLastBlock ? 0 : 4/8` |

`src/wh_client_crypto.c` is untouched: only the server translates, so the
client's copy *defines* the wire format. Same-endian wire bytes are identical to
before, so there is no cross-version interop break. Closes 4041.

### Tests
New `test-refactor/misc/wh_test_message_crypto.c` — the translate layer had no
coverage at all. A loopback test cannot observe peers that disagree on byte
order, so it drives the helpers directly with a foreign-endian magic and checks
per-word swapping, byte-for-byte identity under a native magic, involution,
in-place translation, and that a finalized digest is left alone. It covers the
Keccak state too, which validates the test's own swap detector against
already-correct code.

### Verification
- `test-refactor`: 38 passed / 0 failed of 63; `DMA=1` 42 passed; `NOCRYPTO=1` 12
  passed. Legacy `test/` suite clean.
- Clean under `-std=c90 -Werror -Wall -Wextra`.
- Negative control: reducing the swap to a plain `memcpy` fails the new test
  (`rc=-1`).


src/wh_message_crypto.c
```diff
@@ -639,6 +639,43 @@ int wh_MessageCrypto_TranslateEd25519VerifyResponse(
     return 0;
 }
 
+/* Number of chaining words in a SHA-2 intermediate state, for every variant */
+#define WH_MESSAGE_CRYPTO_SHA2_STATE_WORDS 8
+
+/* Swap the host-order chaining words of a serialized SHA-2 state, as the SHA3
+ * Keccak state does. wordSize is 4 (SHA224/256) or 8 (SHA384/512); any other
+ * value, including 0 for a finalized digest, copies the array verbatim. */
```
AlexLanzano: We probably don't need to mention SHA3 here

AlexLanzano (dismissed):
Small nit. Looks good otherwise

AlexLanzano:
@bigbrett over to you

src/wh_message_crypto.c
yosuke-wolfssl: You are correct, I fixed the comment

bigbrett (dismissed):

PR #471, opened 2026-07-21
Update the SHE test to have the KEK key id configurable. Allow SHE keys to be wrapped via wh_Client_KeyWrap

AlexLanzano:
- Add `WOLFHSM_CFG_TEST_SHE_INTEROP_KEK_ID` to the she key wrap test to allow the KEK id to be overridden
- Allow the server to wrap keys of type `WH_KEYTYPE_SHE`
- Update the SHE keywrap test to use the server to wrap the key blob instead of doing so locally

### DIFF
--- a/src/wh_server_keystore.c
+++ b/src/wh_server_keystore.c
@@ -1838,12 +1838,30 @@ static int _HandleKeyWrapRequest(whServerContext*                  server,
     /* Wrapped key size is only passed back to the client on success */
     resp->wrappedKeySz = 0;
 
-    /* Ensure the keyId in the wrapped metadata has the wrapped flag set */
-    if (!WH_KEYID_ISWRAPPED(metadata.id)) {
+    /* Require a wrappable type: WRAPPED, or SHE so a SHE key can be wrapped and
+     * later primed via unwrap-and-cache. */
+    if (WH_KEYID_TYPE(metadata.id) != WH_KEYTYPE_WRAPPED
+#ifdef WOLFHSM_CFG_SHE_EXTENSION
+        && WH_KEYID_TYPE(metadata.id) != WH_KEYTYPE_SHE
+#endif
+    ) {
         WH_LOG_F(&server->log, WH_LOG_LEVEL_ERROR,
-                 "KeyWrapRequest: keyId:0x%08X is not wrapped", metadata.id);
+                 "KeyWrapRequest: keyId:0x%08X is not a wrappable type",
+                 metadata.id);
         return WH_ERROR_BADARGS;
     }
+
+#ifdef WOLFHSM_CFG_SHE_EXTENSION
+    /* SHE keys are a fixed size; reject a mis-sized blob up front so it can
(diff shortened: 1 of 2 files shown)


bigbrett (changes requested):
looks good, but we need parity in test-refactor/ too. Also, if we end up fully removing `whTest_BuildSheKeyBlob` after that (since I think thats the only other use of it) then we should just kill that function entirely.

test/wh_test_she.c
```diff
@@ -96,9 +96,13 @@ enum {
 
 #if defined(WOLFHSM_CFG_KEYWRAP) && defined(HAVE_AESGCM) && \
     !defined(WOLFHSM_CFG_TEST_CLIENT_ONLY)
-/* Id of the trusted KEK the server task provisions for the SHE<->keywrap
- * interop tests. Defined here so the server task can use it too. */
+/* Trusted KEK id for the SHE<->keywrap interop tests. Override via
+ * WOLFHSM_CFG_TEST_SHE_INTEROP_KEK_ID (e.g. a hardware KEK). */
+#ifdef WOLFHSM_CFG_TEST_SHE_INTEROP_KEK_ID
+#define WH_SHE_INTEROP_KEK_ID WOLFHSM_CFG_TEST_SHE_INTEROP_KEK_ID
```
bigbrett, line 102: I think this override will make it such that tests still try to provision the KEK at this ID unconditionally in `_ProvisionSheServerKek()` (and possibly elsewhere?)

src/wh_server_keystore.c
```diff
@@ -1838,13 +1838,31 @@ static int _HandleKeyWrapRequest(whServerContext*                  server,
     /* Wrapped key size is only passed back to the client on success */
     resp->wrappedKeySz = 0;
 
-    /* Ensure the keyId in the wrapped metadata has the wrapped flag set */
-    if (!WH_KEYID_ISWRAPPED(metadata.id)) {
+    /* Require a wrappable type: WRAPPED, or SHE so a SHE key can be wrapped and
+     * later primed via unwrap-and-cache. */
+    if (WH_KEYID_TYPE(metadata.id) != WH_KEYTYPE_WRAPPED
+#ifdef WOLFHSM_CFG_SHE_EXTENSION
+        && WH_KEYID_TYPE(metadata.id) != WH_KEYTYPE_SHE
```
bigbrett, line 1845: Actually, on a deeper look, we can't do this. It lets a client choose arbitrary bytes and load it into a target SHE slot, instead of requiring it to go through the SHE key update process. SHE keys should only be created/provisioned out-of-band or updated via SHE key update protocol (excluding `wh_Client_ShePreProgramKey()` that is compile time-gated on a different branch and should be fixed soon)

bigbrett (changes requested):

bigbrett:
closing per our discussion. Favor moving the test in question that requires preprovisioned server keys out of the client-only tests and only running in POSIX enviroment

PR #472, opened 2026-07-21
Translate the key wrap metadata trailer

yosuke-wolfssl:
### Problem
The key wrap protocol carries a `whNvmMetadata` as a **raw struct trailer**
after the fixed request/response headers, but the server translated only the
headers. `whNvmMetadata` is four `uint16_t` fields (`id`, `access`, `flags`,
`len`), so every one is byte-swapped for a cross-endian client/server pair:

| Path | Trailer | Symptom for a swapped client |
|---|---|---|
| `WH_KEY_KEYWRAP` request | `whNvmMetadata` + key | `id` `0x4025` (TYPE=WRAPPED) read as `0x2540` (TYPE=SHE) → valid wrap rejected `WH_ERROR_BADARGS` |
| `WH_KEY_KEYUNWRAPEXPORT` response | `whNvmMetadata` + key | `id`, `access`, `flags`, `len` returned byte-swapped |

Every other NVM message flattens the struct into message fields and translates
each one; key wrap is the only place the raw struct crosses the boundary.

Addressed by f_5469.

### Fix (`src/wh_server_keystore.c`)
Server-side only: clients always send native byte order, and the server
translates both directions from the comm-header `magic`.

- **`wh_MessageNvm_TranslateMetadata()`** — new helper in `wh_message_nvm.[ch]`,
  in-place safe, label copied verbatim.
- **`_HandleKeyWrapRequest`** translates on ingress before any field is
  inspected, establishing the invariant that **blobs always store metadata in
  server-native order**.
- **`_HandleKeyUnwrapAndExportRequest`** translates on egress last, after every
  policy check has run on native values.

Both handlers are `static` with one call site each, so they gain a `magic`
parameter. No client, wire-format, or struct change.

**Also fixed, found while auditing the failure paths:** `*out_resp_size` for
`KEYUNWRAPEXPORT` unconditionally counts a `whNvmMetadata` trailer, so a failed
unwrap shipped up to 32 bytes of stale comm-buffer content to the client. The
handler now zeroes the trailer and any decrypted key with `wh_Utils_ForceZero()`
on every failure exit, and the dispatcher scrubs as a backstop. In-switch
`return`s became `ret = X; break;` so one epilogue covers them all.
`_HandleKeyWrapRequest` also sets `resp->cipherType` before its first failure
exit — the client checks `cipherType` ahead of `rc`, so an early return
previously surfaced as `WH_ERROR_ABORTED` instead of the real error.

### Tests
Reworked on review: **1,179 test lines removed, 128 added.** Both original
files were shapes a real peer cannot produce — `wh_test_keywrap_endian.c` (720)
drove `whServerContext*` internals with synthetic magics, and
`wh_test_keywrap_endian_e2e.c` (459) hand-built packets and pushed them at the
transport, because `wh_CommClient` only ever sends native magic. Their two
`wh_test_list.c` entries and the `README.md` coverage-gap bullet go with them.

Replaced by **`test-refactor/misc/wh_test_message_nvm.c`**, following #468: a
loopback pair structurally cannot observe a byte-order bug, so the test calls
`wh_MessageNvm_TranslateMetadata()` directly with `WH_COMM_MAGIC_SWAP`. No
packet crafting, no new entry point registered beyond the one parent function.
It asserts native magic is a no-op, all four scalars swap, the `label` byte
array does not move, translation is its own inverse, in-place (`src == dest`)
matches out-of-place, and NULL args give `WH_ERROR_BADARGS`. Expected values are
written out as literals rather than derived from `wh_Translate16()`, so the
oracle does not share the code under test. The helper is ungated, so the test
runs in every configuration.

The `whNvmMetadata` entry added to `wh_test_check_struct_padding.c` stays — it
is what keeps the raw wire struct padding-free.

Coverage dropped with the harness, none of it reachable from a real client:
foreign-magic acceptance at the comm layer, the `req_packet`/`resp_packet`
aliasing that makes egress translation an in-place swap, wrap/unwrap policy
negatives already covered by `client-server/wh_test_keywrap.c`, and the
malformed-request residue check for the scrub described above. The new
"keep in step with `wh_MessageNvm_TranslateAddObjectRequest`" comment likewise
has no test behind it; the twin is pre-existing code this PR does not touch.

### Verification
On merge base `237bfbc`, clean under `-std=c90 -Werror -Wall -Wextra`, zero
warnings in all four configurations:

| Config | Result |
|---|---|
| `test-refactor` default | 46 passed / 26 skipped / 0 failed of 72 |
| `test-refactor DMA=1` | 50/22/0 |
| `test-refactor SHE=1` | 52/20/0 |
| `test-refactor NOCRYPTO=1` | 17/55/0 |

`whTest_MessageNvmTranslate` passes in all four. Two negative controls, each
rebuilt and re-run: dropping the `access` `WH_T16` fails the test, and dropping
the label `memcpy` fails it; restoring both returns 46/26/0.

### DIFF
--- a/src/wh_server_keystore.c
+++ b/src/wh_server_keystore.c
@@ -40,6 +40,7 @@
 #include "wolfhsm/wh_error.h"
 #include "wolfhsm/wh_message.h"
 #include "wolfhsm/wh_message_keystore.h"
+#include "wolfhsm/wh_message_nvm.h" /* For wh_MessageNvm_TranslateMetadata */
 #include "wolfhsm/wh_utils.h"
 #include "wolfhsm/wh_server.h"
 #include "wolfhsm/wh_log.h"
@@ -1830,6 +1831,7 @@ static int _AesGcmDataUnwrap(whServerContext* server, uint16_t serverKeyId,
 #endif /* !NO_AES */
 
 static int _HandleKeyWrapRequest(whServerContext*                  server,
+                                 uint16_t                          magic,
                                  whMessageKeystore_KeyWrapRequest* req,
                                  uint8_t* reqData, uint32_t reqDataSz,
                                  whMessageKeystore_KeyWrapResponse* resp,
@@ -1849,19 +1851,23 @@ static int _HandleKeyWrapRequest(whServerContext*                  server,
         return WH_ERROR_BADARGS;
     }
 
+    /* Set before any failure exit: the client checks cipherType before rc */
+    resp->cipherType = req->cipherType;
+    resp->wrappedKeySz = 0;
+
     /* Check if the reqData is big enough to hold the metadata and key */
     if (reqDataSz < sizeof(metadata) + req->keySz) {
         return WH_ERROR_BUFFER_SIZE;
     }
 
-    /* Extract the metadata and key from reqData */
+    /* Extract the metadata and key. The metadata trailer arrives in the
+     * client's byte order, so translate before any field is used */
     memcpy(&metadata, reqData, sizeof(metadata));
+    ret = wh_MessageNvm_TranslateMetadata(magic, &metadata, &metadata);
+    if (ret != WH_ERROR_OK) {
+        return ret;
+    }
     memcpy(key, reqData + sizeof(metadata), req->keySz);
-
-    /* Ensure the cipher type in the response matches the request */
-    resp->cipherType = req->cipherType;
-    /* Wrapped key size is only passed back to the client on success */
-    resp->wrappedKeySz = 0;
 
     /* Ensure the keyId in the wrapped metadata has the wrapped flag set */
     if (!WH_KEYID_ISWRAPPED(metadata.id)) {
@@ -2037,16 +2043,18 @@ out:
 }
 
 static int _HandleKeyUnwrapAndExportRequest(
-    whServerContext* server, whMessageKeystore_KeyUnwrapAndExportRequest* req,
-    uint8_t* reqData, uint32_t reqDataSz,
-    whMessageKeystore_KeyUnwrapAndExportResponse* resp, uint8_t* respData,
-    uint32_t respDataSz)
+    whServerContext* server, uint16_t magic,
+    whMessageKeystore_KeyUnwrapAndExportRequest* req, uint8_t* reqData,
+    uint32_t reqDataSz, whMessageKeystore_KeyUnwrapAndExportResponse* resp,
+    uint8_t* respData, uint32_t respDataSz)
 {
-    int            ret;
+    /* Defensive: a case that never assigns ret cannot report success */
+    int            ret = WH_ERROR_BADARGS;
     uint8_t*       wrappedKey;
     whNvmMetadata* metadata;
     uint8_t*       key;
     whKeyId        serverKeyId;
+    uint16_t       keySz = 0;
 
     if (server == NULL || req == NULL || reqData == NULL || resp == NULL ||
         respData == NULL) {
@@ -2080,11 +2088,13 @@ static int _HandleKeyUnwrapAndExportRequest(
 #ifndef NO_AES
 #ifdef HAVE_AESGCM
         case WC_CIPHER_AES_GCM: {
-            uint16_t keySz;
+            uint16_t wrappedKeyUser = 0;
+            uint16_t wrappedKeyType = 0;
 
             if (req->wrappedKeySz < WH_KEYWRAP_AES_GCM_HEADER_SIZE +
                                     sizeof(*metadata)) {
-                return WH_ERROR_BADARGS;
+                ret = WH_ERROR_BADARGS;
+                break;
             }
 
             keySz = req->wrappedKeySz -
@@ -2092,7 +2102,8 @@ static int _HandleKeyUnwrapAndExportRequest(
 
             /* Check if the response data can fit the metadata + key  */
             if (respDataSz < sizeof(*metadata) + keySz) {
-                return WH_ERROR_BUFFER_SIZE;
+                ret = WH_ERROR_BUFFER_SIZE;
+                break;
             }
 
             /* Unwrap the key. The plaintext is handed back to the client, not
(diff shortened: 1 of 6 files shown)


src/wh_server_keystore.c
```diff
@@ -1829,8 +1831,10 @@ static int _HandleKeyWrapRequest(whServerContext*                  server,
         return WH_ERROR_BUFFER_SIZE;
     }
 
-    /* Extract the metadata and key from reqData */
+    /* Extract the metadata and key. The metadata trailer arrives in the
+     * client's byte order, so translate before any field is used */
     memcpy(&metadata, reqData, sizeof(metadata));
+    (void)wh_MessageNvm_TranslateMetadata(magic, &metadata, &metadata);
```
AlexLanzano: Please check the error return of this

src/wh_server_keystore.c
```diff
@@ -2125,6 +2129,10 @@ static int _HandleKeyUnwrapAndExportRequest(
             return WH_ERROR_BADARGS;
     }
 
+    /* The policy checks above run on native metadata, so put the trailer into
+     * the client's byte order only once it is on its way out */
+    (void)wh_MessageNvm_TranslateMetadata(magic, metadata, metadata);
```
AlexLanzano: Please check the error return from this call.

Don't we need to do the translation prior to the policy checks (right after `_AesGcmKeyUnwrap`)? Else the metadata coming in would still be in the native client endianess

test-refactor/server/wh_test_keywrap_endian.c
AlexLanzano, line 1: I would like to see an end-to-end test of this with a client and server if possible

AlexLanzano (changes requested):

src/wh_server_keystore.c
yosuke-wolfssl: Thanks, fixed. Now:

```c
memcpy(&metadata, reqData, sizeof(metadata));
ret = wh_MessageNvm_TranslateMetadata(magic, &metadata, &metadata);
if (ret != WH_ERROR_OK) {
    return ret;
}

File: src/wh_server_keystore.c (hunk 2 above)
Comment (yosuke-wolfssl) [reply src/wh_server_keystore.c]: Error return is now checked, using the fallthrough guard so it can't mask a
prior failure:

```c
if (ret == WH_ERROR_OK) {
    ret = wh_MessageNvm_TranslateMetadata(magic, metadata, metadata);
}
```

On the placement — I kept it on egress, because the metadata here doesn't come
off the wire. It comes out of the decrypted blob, and the blob always holds
metadata in *server* order:

- `_AesGcmKeyWrapWithKek()` `memcpy`s the `whNvmMetadata` struct into the
  plaintext blob verbatim, and `_HandleKeyWrapRequest` translates
  client -> native *before* wrapping. `_HandleKeyWrapExportRequest` wraps
  metadata read straight from the keystore, which is native by definition.
- `_HandleKeyUnwrapAndCacheRequest` does no translation at all and stores blob
  metadata directly into the keystore. That is only correct if blob metadata is
  native, which is the same invariant.

So translating right after `_AesGcmKeyUnwrap` would byte-swap the values that
`WH_KEYID_IS_UNASSIGNED(metadata->id)`, the `WH_KEYTYPE_WRAPPED` check,
`metadata->flags & WH_NVM_FLAGS_NONEXPORTABLE`, and the USER ownership check
all read — a swapped-endian client would then bypass or trip those checks
incorrectly. Egress-last keeps every policy check on native values.

I've expanded the comment at the call site to say this explicitly.

test-refactor/server/wh_test_keywrap_endian.c
yosuke-wolfssl: Added: `test-refactor/client-server/wh_test_keywrap_endian_e2e.c`
(`whTest_KeyWrapEndianE2E`, client group). Wrap -> unwrap-and-export against the
live server over the real transport, once as a same-endian peer and once as a
swapped one.

src/wh_server_keystore.c
yosuke-wolfssl: Correcting my earlier reply, which posted truncated. Final state:

```c
/* Set before any failure exit: the client checks cipherType before rc */
resp->cipherType = req->cipherType;
resp->wrappedKeySz = 0;

memcpy(&metadata, reqData, sizeof(metadata));
ret = wh_MessageNvm_TranslateMetadata(magic, &metadata, &metadata);
if (ret != WH_ERROR_OK) {
    return ret;
}
```

The return is checked. It is dead today (both args are `&metadata`), but
checking it surfaced a live adjacent bug, which is why the `resp->` initializers
moved up: `wh_Client_KeyWrapResponse` tests `cipherType` *before* it reads `rc`,
and `wrapResp` is a zeroed dispatcher local — so any earlier return, including
the pre-existing `WH_ERROR_BUFFER_SIZE` path, reached the client as
`WH_ERROR_ABORTED`. Verified: `cipherType` came back `0x0000` before the move,
`0x0003` after.

src/wh_server_keystore.c
yosuke-wolfssl: Checked now, and it mattered more than expected — the check made a
previously-dead branch live:

```c
if (ret == WH_ERROR_OK) {
    ret = wh_MessageNvm_TranslateMetadata(magic, metadata, metadata);
}
if (ret != WH_ERROR_OK && respDataSz >= sizeof(*metadata)) {
    resp->keySz = 0;
    wh_Utils_ForceZero(metadata, scrubSz);
}
```

I had these as `if` / `else if` first, which fails open: a translation failure
would skip the scrub and leave `resp->keySz` set, shipping the decrypted key
alongside an error rc. Two `if`s keyed off the final `ret` close it.

The scrub came out of the same audit — `*out_resp_size` here unconditionally
counts a trailer, so a failed unwrap shipped up to 32 bytes of stale buffer.

Placement answer unchanged: egress-last, since the blob holds metadata in server
order and every policy check must run on native values.

test-refactor/server/wh_test_keywrap_endian.c
yosuke-wolfssl: Expanded since my last reply. `whTest_KeyWrapEndianE2E` now also covers a denied
export (`NONEXPORTABLE`) and asserts the response trailer comes back zeroed.

That case needs the real harness: the server test uses separate request and
response buffers, but `wh_Server_HandleRequestMessage` passes one pointer as
both. For `KEYUNWRAPEXPORT` both fixed structs are 8 bytes, so the blob and the
response trailer share an offset and `_AesGcmKeyUnwrap` writes the decrypted
trailer there before the scrub.

Byte swaps in both files are hand-rolled rather than reusing the `Translate*`
helpers, so a bug under test cannot cancel itself out — a `flags` bug injected
into `wh_MessageNvm_TranslateMetadata` is now caught by the round trip and the
`NONEXPORTABLE` negative, where it previously passed undetected.

yosuke-wolfssl:
Hello @AlexLanzano ,
I fixed issues you mentioned with some additional stuff I found during rework.
The PR description is also updated accordingly.
Could you review this again ?

AlexLanzano:
@yosuke-wolfssl Looks like you have some failing test cases with your new changes? Could you take a look?

AlexLanzano (approved):

AlexLanzano:
Looks good. Passing off to you @bigbrett

bigbrett (approved):

PR #476, opened 2026-07-22
Bound LMS, XMSS and ML-KEM DMA response frames in the crypto client

yosuke-wolfssl:
### Problem

`wh_Client_RecvResponse` reports the received payload length but nothing validates it against the message being parsed. The comm client reuses one buffer for request and response, so a reply shorter than the response struct leaves the tail holding the client's own outbound request bytes. An unchecked overlay read then returns stale data instead of failing.

Ten DMA response handlers in `wh_client_crypto.c` overlay their response struct with no frame bound:

| Handler | Unbounded read |
|---|---|
| `wh_Client_{Lms,Xmss}MakeKeyDma` | `res->keyId`, stored into the caller's out-param and the key |
| `wh_Client_{Lms,Xmss}SignDma` | `res->sigLen`, bounded against the caller's buffer but not the frame |
| `wh_Client_{Lms,Xmss}VerifyDma` | `resp->res`, the verify verdict itself |
| `wh_Client_{Lms,Xmss}SigsLeftDma` | `res->sigsLeft`, which gates whether a one-time key is reused |
| `_MlKemMakeKeyDma` | `res->keyId`, and `res->keySize` bounded only by the staging buffer |
| `wh_Client_MlDsaSignDma` | `res->sigLen`, bounded by neither the frame nor the caller's buffer |

Follow-up to #457, same family as #474 (CMAC) and #475 (AES DMA).

### Fix (`src/wh_client_crypto.c`)

Each of these responses is fixed size with no trailing payload, so one bound covers every site:

```c
const uint32_t hdr_sz =
if (res_len < hdr_sz) {
    ret = WH_ERROR_ABORTED;
}
```

These handlers previously passed `req_len` as the length out-param of `wh_Client_RecvResponse`; each now receives into its own `res_len`, matching every other handler in the file.

- **Inside the success branch.** The DMA dispatcher sends only the generic header when a handler fails, so an unconditional bound would turn every legitimate error reply into `WH_ERROR_ABORTED` and hide the server's `rc`.
- **Assigns `ret`, never returns early**, so the DMA unmaps and `_MlKemMakeKeyDma`'s staging-buffer zeroize still run.
- **`wh_Client_MlDsaSignDma`** also gains the `res->sigLen > sigCap` caller-capacity check its LMS/XMSS siblings already had, and now unmaps `sigCap` on `WRITE_POST` to match what `WRITE_PRE` mapped.

### Tests

Source-only, no test added. A conforming server always sends the full fixed-size response, so the public `wh_Client_*` API has no path to these guards, and a test reachable only through a fake transport or a hand-built packet exercises the harness rather than the fix.

### Verification

- `test-refactor`: default 46 passed / 29 skipped / 0 failed of 75 · `DMA=1` 50/25/0 · `SHE=1` 52/23/0. Legacy `test/` suite clean.
- Clean under `-std=c90 -Werror -Wall -Wextra`; ASan + UBSan clean on `DMA=1`.
- Mutation-checked: weakening a guard to `hdr_sz * 2` is caught by the LMS, XMSS and ML-DSA DMA tests for 9 of the 10. `_MlKemMakeKeyDma` has no ML-KEM test in either suite, so that guard is compile-verified only.

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -10829,6 +10829,7 @@ int wh_Client_MlDsaSignDma(whClientContext* ctx, const byte* in, word32 in_len,
     uint8_t*                              dataPtr = NULL;
     uintptr_t                             inAddr  = 0;
     uintptr_t                             outAddr = 0;
+    word32                                sigCap  = 0;
 
     /* Transaction state */
     whKeyId key_id;
@@ -10838,6 +10839,9 @@ int wh_Client_MlDsaSignDma(whClientContext* ctx, const byte* in, word32 in_len,
         (out == NULL) || (out_len == NULL)) {
         return WH_ERROR_BADARGS;
     }
+
+    /* Caller's signature buffer capacity, before the response overwrites it */
+    sigCap = *out_len;
 
     key_id = WH_DEVCTX_TO_KEYID(key->devCtx);
 
@@ -10905,7 +10909,7 @@ int wh_Client_MlDsaSignDma(whClientContext* ctx, const byte* in, word32 in_len,
             }
 
             if (ret == WH_ERROR_OK) {
-                req->sig.sz = *out_len;
+                req->sig.sz = sigCap;
                 ret         = wh_Client_DmaProcessClientAddress(
                     ctx, (uintptr_t)out, (void**)&outAddr, req->sig.sz,
                     WH_DMA_OPER_CLIENT_WRITE_PRE, (whDmaFlags){0});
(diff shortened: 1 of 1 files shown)


src/wh_client_crypto.c
```diff
@@ -11552,7 +11570,13 @@ int wh_Client_LmsSignDma(whClientContext* ctx, const byte* msg, word32 msgSz,
             ret = _getCryptoResponse(dataPtr, WC_PK_TYPE_PQC_STATEFUL_SIG_SIGN,
                                      (uint8_t**)&res);
             if (ret >= 0) {
-                if (res->sigLen > sigCap) {
+                const uint32_t hdr_sz =
+                    sizeof(whMessageCrypto_GenericResponseHeader) +
+                    sizeof(*res);
+                if (res_len < hdr_sz / 2) {
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

Frauschi (dismissed):
LGTM, but has conflicts to main that needs fixing.

yosuke-wolfssl:
Hi @Frauschi ,
Thanks for reviewing, I solved the merge conflict

Frauschi (dismissed):

src/wh_client_crypto.c
padelsbach: AI tells me we should validate `out_len` here as well

padelsbach, line 5374: Should also apply the response length check here and in `_CmacKdfMakeKey`

padelsbach (changes requested):

src/wh_client_crypto.c
yosuke-wolfssl: Both of these landed on `main` in the meantime — `3d2e44d` ("Bound KDF response output size against received frame") added the check to `_HkdfMakeKey` and `_CmacKdfMakeKey`. Your comment was against the pre-rebase base, which predated that merge.

Since these two responses carry a trailing payload, the bound there is the variable-length form (`res_len < hdr_sz || res->outSz > (res_len - hdr_sz)`) rather than the fixed-size one this PR uses. Nothing to change here after the rebase onto `67aecaa`.

yosuke-wolfssl: Good catch — fixed in `a2cf1c3`. This one was worse than the nine in the PR: `res->sigLen` was bounded by neither the frame nor the caller's buffer, and it fed the `WRITE_POST` unmap. It now follows the same shape as `wh_Client_LmsSignDma`/`wh_Client_XmssSignDma`:

- `res_len < hdr_sz` → `WH_ERROR_ABORTED`
- `res->sigLen > sigCap` → `WH_ERROR_BADARGS`, the caller-capacity check the LMS/XMSS siblings already had

`sigCap` snapshots `*out_len` before the response overwrites it, and `WRITE_POST` now unmaps `sigCap` to match what `WRITE_PRE` mapped (it previously unmapped the updated `*out_len`).

Verified: `whTest_Crypto_MlDsa` passes under `DMA=1`, and weakening the guard to `hdr_sz * 2` makes it fail with `WH_ERROR_ABORTED`, so the path is genuinely exercised. ASan + UBSan clean.

yosuke-wolfssl:
Hi @padelsbach ,
Could you review this again ?

padelsbach (approved):

PR #482, opened 2026-07-23
CI runtime: add timeouts and concurrency checks

padelsbach:
1. Add timeouts to ensure jobs don't spin endlessly
2. Use `concurrency` to ensure only the latest job runs (eg. for a given `git push` to a PR), and old jobs are cancelled

These are both generally good-citizenship changes, but can make an impact across the WolfSSL org when things go wrong. Approx 1250 runner-minutes would have been saved in the last 6ish weeks (5-10%).

### DIFF
--- a/.github/workflows/build-and-bench.yml
+++ b/.github/workflows/build-and-bench.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ '*' ]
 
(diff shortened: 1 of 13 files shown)


bigbrett (approved):
I was thinking of making this change you must be telepathic

PR #490, opened 2026-07-24
CI: automate coverage comparison for test-refactor

padelsbach:
Changelog:

1. don't run coverage on every PR push. 1x per week to refresh the report, or on-demand (approx 9-10 runner-min saved)
2. automate the comparison between `test/` and `test-refactor/`. informs us on which legacy test modules to port and eventually drop


bigbrett:
@padelsbach merge conflicts. Also, any reason we wouldn't just want this to run nightly? Doesn't take that long.

padelsbach:
@bigbrett I think it can be run mostly on-demand until refactor becomes the norm, but it doesn't hurt to run nightly. Updated.

bigbrett (approved):

PR #491, opened 2026-07-26
Add 32-bit CI job, fix sign-compare warnings in POSIX flash port

yosuke-wolfssl:
### Problem
No CI job builds or runs wolfHSM at 32-bit width, yet every supported port is
32-bit. The AES `needed_size` overflow class #487 only triggers at 32 bits, so it was
invisible to every gate. Adding a 32-bit job also surfaced width bugs in the
POSIX flash port that a 64-bit host silently hides.

### Fix (`port/posix/posix_flash_file.c`)
- **sign-compare:** 5 `off_t`/`ssize_t`-vs-`uint32_t` comparisons that fail
  `-Werror` at 32 bits. The `rc != size` byte-count checks became
  `(rc < 0) || ((size_t)rc != (size_t)size)`.
- **`MAX_OFFSET`:** now computed in `uint64_t` so the doubling can't wrap and the
  `_Init` comparisons stay unsigned at every width — avoids a signed narrowing
  that could take a large `partition_size` into `ftruncate(<negative>)`.
- **`pfill()`:** returns an `int` status (`0`/`-1`) instead of a `size_t` count
  cast to `ssize_t`, which aliased negative above 2 GiB and tripped the new
  `rc < 0` guard.

### CI (`.github/workflows/build-and-test-refactor.yml`)
New ubuntu-only `build-32bit` job (`gcc-multilib`), parallel to the 64-bit matrix:

| Step | Why |
|---|---|
| refactor default / `SHE=1 DMA=1` | size guards + DMA handlers at 32-bit |
| legacy `test/` suite | only harness that runs `POSIX_FLASH_FILE_CB`; refactor uses the RAM sim |
| `NOCRYPTO=1` | distinct source set, skips wolfSSL |

`ARCHFLAGS` is passed to `make run` too, so a stale rebuild can't switch arch mid-step.

### Verification
64-bit and 32-bit (i386/QEMU) match exactly: default 43/26/0, `SHE+DMA` 53/16/0,
`NOCRYPTO` 16/53/0, legacy suite exit 0. Zero warnings, `-Werror` intact —
including the legacy tree's `-Wpointer-arith`, never compiled at 32-bit before.
Built against wolfSSL master tip, so no upstream `-Werror` relaxation is needed.
The legacy run executes `posixFlashFile_*` for the first time at 32-bit width;
a negative control (neutered `pfill()`) aborts that suite, confirming the step gates.

### DIFF
--- a/port/posix/posix_flash_file.c
+++ b/port/posix/posix_flash_file.c
@@ -23,6 +23,7 @@
  */
 
 #include <stddef.h>     /* For NULL */
+#include <stdint.h>     /* For uint64_t, UINT32_MAX */
 #include <fcntl.h>      /* For O_xxxx */
 #include <sys/types.h>  /* For off_t, stat */
 #include <sys/stat.h>   /* For fstat */
@@ -41,24 +42,26 @@ enum {
 };
 
 /** Local declarations */
-#define MAX_OFFSET(_context) ((_context)->partition_size * 2)
+/* Computed in 64 bits so the doubling cannot wrap and the comparisons against
+ * it stay unsigned on platforms with a 32-bit off_t */
+#define MAX_OFFSET(_context) ((uint64_t)(_context)->partition_size * 2)
 
 /* Helper for pwrite like memset.  Write the byte in c to filedes for size
- * bytes starting at offset */
-static ssize_t pfill(int filedes, int c, size_t size, off_t offset);
+ * bytes starting at offset.  Returns 0 on success, -1 on error */
+static int pfill(int filedes, int c, size_t size, off_t offset);
 
 /** Local implementations */
-static ssize_t pfill(int filedes, int c, size_t size, off_t offset)
+static int pfill(int filedes, int c, size_t size, off_t offset)
 {
-    int rc = 0;
+    ssize_t rc = 0;
     uint8_t data = (uint8_t)c;
     size_t count = 0;
     while (count < size) {
         rc = pwrite(filedes, &data, sizeof(data), offset + count);
-        if (rc != sizeof(data)) return rc;
+        if (rc != (ssize_t)sizeof(data)) return -1;
         count += sizeof(data);
     }
-    return size;
+    return 0;
 }
 
 
@@ -69,6 +72,7 @@ int posixFlashFile_Init(   void* c,
     const posixFlashFileConfig* config = cf;
     struct stat st = {0};
     off_t file_size = 0;
+    uint64_t max_offset = 0;
 
     int ret = 0;
     int rc = 0;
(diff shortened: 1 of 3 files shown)


port/posix/posix_flash_file.c
```diff
@@ -87,23 +90,26 @@ int posixFlashFile_Init(   void* c,
         context->partition_size = config->partition_size;
         context->erased_byte = config->erased_byte;
 
+        max_offset = MAX_OFFSET(context);
+
         rc = fstat(context->fd_p1 - 1, &st);
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

port/posix/posix_flash_file.c
```diff
@@ -41,24 +41,26 @@ enum {
 };
 
 /** Local declarations */
-#define MAX_OFFSET(_context) ((_context)->partition_size * 2)
+/* Computed in 64 bits so the doubling cannot wrap and the comparisons against
+ * it stay unsigned on platforms with a 32-bit off_t */
+#define MAX_OFFSET(_context) ((uint64_t)(_context)->partition_size * 2)
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

port/posix/posix_flash_file.c
```diff
@@ -242,7 +256,7 @@ int posixFlashFile_Verify( void* c,
     uint32_t data_offset = 0;
```
yosuke-wolfssl, line 258: Fixed

port/posix/posix_flash_file.c
```diff
@@ -87,23 +90,34 @@ int posixFlashFile_Init(   void* c,
         context->partition_size = config->partition_size;
         context->erased_byte = config->erased_byte;
 
+        max_offset = MAX_OFFSET(context);
+
+        /* The flash interface addresses with uint32_t offsets and seeks with
+         * off_t, so reject a partition whose doubled size exceeds either */
+        if ((max_offset > 0xFFFFFFFFu) ||
```
bigbrett: use constants from `limits.h`

bigbrett (changes requested):

port/posix/posix_flash_file.c
yosuke-wolfssl: Thanks for reviewing, and Done — swapped the hex literal for `UINT32_MAX`:

```c
if ((max_offset > UINT32_MAX) ||
    ((uint64_t)(off_t)max_offset != max_offset)) {
```

Note that `UINT32_MAX` comes from `<stdint.h>`, not `<limits.h>` — `limits.h`
only carries the `CHAR`/`INT`/`LONG_MAX` family, not the fixed-width ones, so
that's the include I added.

I left the second clause as the round-trip cast: POSIX defines no `OFF_MAX`,
and `off_t` isn't guaranteed to be `long`, so `LONG_MAX` would be wrong where
it's `long long` or a 32-bit `int`. Happy to swap it if you have a constant in
mind.

bigbrett (dismissed):

bigbrett:
@yosuke-wolfssl merge conflicts

yosuke-wolfssl:
Hi @bigbrett ,
The merge conflict was solved. Please check this again

bigbrett (approved):

PR #493, opened 2026-07-27
Un-nest Ed25519 cryptocb cases from HAVE_CURVE25519

yosuke-wolfssl:
### Problem

In `wh_Client_CryptoCbStd`, the `#ifdef HAVE_ED25519` block holding the
`WC_PK_TYPE_ED25519_KEYGEN`, `_SIGN` and `_VERIFY` cases was nested inside the
`#ifdef HAVE_CURVE25519` block. These are independent wolfSSL build options, so
the three dispatch cases compiled only when *both* were defined.

With Ed25519 enabled and Curve25519 disabled, every Ed25519 operation on a
wolfHSM devId fell through to `default: ret = CRYPTOCB_UNAVAILABLE;`, which
`wh_Client_CryptoCb` deliberately propagates, so wolfCrypt silently used its
software implementation. For a key whose private scalar lives only in the HSM
(`devCtx` holds a keyId, no local material) that path cannot sign at all — HSM
offload broke with no diagnostic. The DMA sibling `wh_Client_CryptoCbDma`
already had the correct sibling layout, confirming the nesting was accidental.

### Fix (`src/wh_client_cryptocb.c`)

Moved `#endif /* HAVE_CURVE25519 */` up to just after the `WC_PK_TYPE_CURVE25519`
case so the Ed25519 group is a sibling rather than a child:

```diff
         } break;              /* end WC_PK_TYPE_CURVE25519 */
+#endif /* HAVE_CURVE25519 */
 
 #ifdef HAVE_ED25519
         ...
 #endif /* HAVE_ED25519 */
-#endif /* HAVE_CURVE25519 */
```

One line moved, net zero. Every other Curve25519/Ed25519 guard pair in the repo
was audited and is already correct. Closes f-7149.

### Test harness

The existing Ed25519 tests did **not** catch this: they call the
`wh_Client_Ed25519*` APIs directly, bypassing the crypto callback, and the inline
test signs with a key that still holds local private material. The whole suite
passed with the dispatch cases compiled out.

- **`_whTest_CryptoEd25519CryptoCbHsmKey`** signs via `wc_ed25519_sign_msg()` on a
  key reduced to a bare keyId, so the callback must handle it; a software fallback
  returns `BAD_FUNC_ARG`. Pins the std path via `wh_Client_SetDmaMode(ctx, 0)` so
  the guard it protects cannot be masked by ambient DMA mode.
- **`NOCURVE25519=1`** build knob in both harness Makefiles, gating
  `HAVE_CURVE25519` in the shared `test/config/user_settings.h`.
- **One CI row each** in `build-and-test-refactor.yml` and `build-and-test.yml`.

### Verification

Negative control on unfixed source: `whTest_Crypto_Ed25519` fails `rc=-173`
(`BAD_FUNC_ARG`) in both `NOCURVE25519` configs and passes with the fix,
isolating the defect to the guard. All builds clean under
`-std=c90 -Werror -Wall -Wextra`, ASan clean.

| Config | Result |
|---|---|
| refactor `NOCURVE25519=1 ASAN=1` | 44 passed, 0 failed of 71 |
| refactor `DMA=1 ASAN=1 NOCURVE25519=1` | 48 passed, 0 failed of 71 |
| refactor `DMA=1 ASAN=1` | 49 passed, 0 failed of 71 |
| refactor `ASAN=1` | 45 passed, 0 failed of 71 |
| legacy `DMA=1 ASAN=1`, with and without `NOCURVE25519=1` | exit 0 |

Intentionally not in this PR, both tracked as follow-ups: mandatory-dispatch
coverage for `WC_PK_TYPE_ED25519_KEYGEN` (needs `WOLF_CRYPTO_CB_ONLY_ED25519`,
which would also force the inline test's local-material sign through the
callback), and cryptocb-API coverage of the `wh_Client_CryptoCbDma` Ed25519 group.

### DIFF
--- a/test-refactor/client-server/wh_test_crypto_ed25519.c
+++ b/test-refactor/client-server/wh_test_crypto_ed25519.c
@@ -24,6 +24,10 @@
  *   _whTest_CryptoEd25519Inline    - pure wolfCrypt (sign+verify locally,
  *                                    plus negative case for tampered sig)
  *   _whTest_CryptoEd25519ServerKey - server-cached sign and verify keyIds
+ *   _whTest_CryptoEd25519CryptoCbHsmKey
+ *                                  - server-only keyIds driven through the
+ *                                    wolfCrypt API, so the cryptocb must
+ *                                    dispatch the operation
  *   _whTest_CryptoEd25519Dma       - same, via the DMA messaging path
  */
 
@@ -353,6 +357,99 @@ static int _whTest_CryptoEd25519ServerKey(whClientContext* ctx)
 
     if (ret == 0) {
         WH_TEST_PRINT("Ed25519 SERVER KEY DEVID=0x%X SUCCESS\n", devId);
+    }
+
+    wc_ed25519_free(pubKey);
+    wc_ed25519_free(key);
+    (void)wc_FreeRng(rng);
+    return ret;
+}
+
+/* Signs and verifies through the wolfCrypt API on a key whose private material
+ * lives only in the server, so the crypto callback must handle the operation.
+ * A software fallback has no private scalar and returns BAD_FUNC_ARG. */
+static int _whTest_CryptoEd25519CryptoCbHsmKey(whClientContext* ctx)
+{
+    int         devId       = WH_CLIENT_DEVID(ctx);
+    int         ret         = 0;
+    WC_RNG      rng[1];
+    ed25519_key key[1]      = {0};
+    ed25519_key pubKey[1]   = {0};
+    whKeyId     signKeyId   = WH_KEYID_ERASED;
+    whKeyId     verifyKeyId = WH_KEYID_ERASED;
+    byte        msg[]       = "Ed25519 cryptocb dispatch message";
+    byte        sig[ED25519_SIG_SIZE];
+    word32      sigSz    = sizeof(sig);
+    int         verified = 0;
+    uint8_t     label[]  = "Ed25519 CryptoCb Key";
+
+    /* Pin the standard (non-DMA) callback. This test exists to prove the
+     * WC_PK_TYPE_ED25519_* cases in wh_Client_CryptoCbStd are compiled in, so
+     * it must not depend on the ambient DMA mode left by an earlier suite. */
+    (void)wh_Client_SetDmaMode(ctx, 0);
+
+    ret = wc_InitRng_ex(rng, NULL, devId);
+    if (ret != 0) {
+        WH_ERROR_PRINT("Failed to wc_InitRng_ex %d\n", ret);
+        return ret;
+    }
+
+    ret = wc_ed25519_init_ex(key, NULL, devId);
+    if (ret != 0) {
+        WH_ERROR_PRINT("Failed to initialize Ed25519 key: %d\n", ret);
+        (void)wc_FreeRng(rng);
+        return ret;
+    }
+
+    ret = wc_ed25519_init_ex(pubKey, NULL, devId);
+    if (ret != 0) {
+        WH_ERROR_PRINT("Failed to initialize Ed25519 public key: %d\n", ret);
+        wc_ed25519_free(key);
+        (void)wc_FreeRng(rng);
+        return ret;
+    }
+
+    ret = wc_ed25519_make_key(rng, ED25519_KEY_SIZE, key);
+    if (ret != 0) {
+        WH_ERROR_PRINT("Failed to generate Ed25519 key: %d\n", ret);
+    }
+    else {
+        ret = whTest_Ed25519ImportToServer(ctx, devId, key, pubKey, label,
+                                           sizeof(label), &signKeyId,
+                                           &verifyKeyId);
+    }
+
+    if (ret == 0) {
+        sigSz = sizeof(sig);
+        ret   = wc_ed25519_sign_msg(msg, (word32)sizeof(msg), sig, &sigSz, key);
+        if (ret != 0) {
+            WH_ERROR_PRINT("wc_ed25519_sign_msg on server key failed: %d\n",
+                           ret);
+        }
+    }
+
+    if (ret == 0) {
+        ret = wc_ed25519_verify_msg(sig, sigSz, msg, (word32)sizeof(msg),
+                                    &verified, pubKey);
+        if (ret != 0) {
+            WH_ERROR_PRINT("wc_ed25519_verify_msg on server key failed: %d\n",
+                           ret);
+        }
+        else if (verified != 1) {
+            WH_ERROR_PRINT("Server key Ed25519 signature did not verify\n");
+            ret = -1;
+        }
+    }
+
+    if (!WH_KEYID_ISERASED(signKeyId)) {
+        (void)wh_Client_KeyEvict(ctx, signKeyId);
+    }
+    if (!WH_KEYID_ISERASED(verifyKeyId)) {
+        (void)wh_Client_KeyEvict(ctx, verifyKeyId);
+    }
+
+    if (ret == 0) {
+        WH_TEST_PRINT("Ed25519 CRYPTOCB DEVID=0x%X SUCCESS\n", devId);
     }
 
     wc_ed25519_free(pubKey);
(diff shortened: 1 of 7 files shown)


test-refactor/client-server/wh_test_crypto_ed25519.c
```diff
@@ -361,6 +361,94 @@ static int _whTest_CryptoEd25519ServerKey(whClientContext* ctx)
...
+    if (ret != 0) {
+        WH_ERROR_PRINT("Failed to generate Ed25519 key: %d\n", ret);
+    }
+    else {
+        ret = whTest_Ed25519ImportToServer(ctx, devId, key, pubKey, label,
+                                           sizeof(label), &signKeyId,
+                                           &verifyKeyId);
+    }
+
+    if (ret == 0) {
```
Frauschi, line 422: **🟡 [Medium] New regression test's dispatch path is ambient rather than pinned, and the only CI row that runs it is DMA=1**
*💡 SUGGEST*  `test`

The defect fixed in this PR lives in `wh_Client_CryptoCbStd` (`src/wh_client_cryptocb.c:426-491`). `wh_Client_CryptoCb` routes to `wh_Client_CryptoCbDma` first whenever `ctx->dma.preferDma` is set (`src/wh_client_cryptocb.c:103-118`), so which callback the new test exercises depends on ambient client state. The new test never pins that state — unlike sibling suites which explicitly call `wh_Client_SetDmaMode()` or loop `WH_TEST_DMA_MODE_CNT` (e.g. `wh_test_crypto_aes.c:1112`, `wh_test_crypto_sha.c:2451`). By code inspection it happens to work today: `test-refactor/posix/wh_test_posix_client.c` passes no `dmaConfig`, and `wh_Client_Init` only sets `preferDma` when `config->dmaConfig != NULL` (`src/wh_client.c:150-157`), so `preferDma == 0` and the Std callback is hit even in the `DMA=1` CI row. (I could not build/run to confirm — the review sandbox has no access to the wolfSSL checkout outside the worktree.) That is fragile in two ways: if the harness default ever flips to `preferDma=1`, or if an earlier suite leaves DMA mode on after an early return, the guard regression stops being detected with no visible signal. The Ed25519 group in `wh_Client_CryptoCbDma` (`src/wh_client_cryptocb.c:1539-1596`) is also never reached through `wc_ed25519_sign_msg()` by this test.

**Suggestion:**
```suggestion
/* Pin the std cryptocb dispatch: this test exists to prove the
     * WC_PK_TYPE_ED25519_* cases in wh_Client_CryptoCbStd are compiled in. */
    (void)wh_Client_SetDmaMode(ctx, 0);

    if (ret == 0) {
        sigSz = sizeof(sig);
        ret   = wc_ed25519_sign_msg(msg, (word32)sizeof(msg), sig, &sigSz, key);
```

test-refactor/client-server/wh_test_crypto_ed25519.c
```diff
@@ -361,6 +361,94 @@ static int _whTest_CryptoEd25519ServerKey(whClientContext* ctx)
...
+
+    ret = wc_ed25519_init_ex(pubKey, NULL, devId);
+    if (ret != 0) {
+        WH_ERROR_PRINT("Failed to initialize Ed25519 public key: %d\n", ret);
+        wc_ed25519_free(key);
+        (void)wc_FreeRng(rng);
+        return ret;
+    }
+
+    ret = wc_ed25519_make_key(rng, ED25519_KEY_SIZE, key);
```
Frauschi, line 412: **🔵 [Low] WC_PK_TYPE_ED25519_KEYGEN dispatch remains uncovered by the new regression test**
*💡 SUGGEST*

The PR un-nests three dispatch cases — `WC_PK_TYPE_ED25519_KEYGEN`, `_SIGN` and `_VERIFY` — but the new test only pins two of them. The sign case is covered because the key is reduced to a bare keyId (software fallback hits `if (!key->privKeySet) return BAD_FUNC_ARG;` in wolfSSL `ed25519.c:559`), and verify likewise (`pubKeySet == 0` after re-init). The keygen case is not: `wc_ed25519_make_key(rng, ED25519_KEY_SIZE, key)` at line 403 runs against a freshly initialised key with local material available, so when `WC_PK_TYPE_ED25519_KEYGEN` is compiled out the callback returns `CRYPTOCB_UNAVAILABLE`, wolfCrypt generates the key in software, and every subsequent step of the test still succeeds. This is exactly the failure mode described in the PR body — silent software fallback with no diagnostic — surviving for one of the three cases the fix touches.

**Recommendation:** Add a keygen assertion that can only pass through the callback — e.g. generate into a key whose `devCtx` already carries a keyId via `wh_Client_Ed25519MakeCacheKey`/`wh_Client_Ed25519MakeExportKey` semantics, or assert the resulting key is a server handle rather than one holding local private material — so a re-nested `WC_PK_TYPE_ED25519_KEYGEN` guard is caught too.

Frauschi (commented):
## 🐺 Skoll Code Review

**Overall recommendation:** `APPROVE`
**Findings:** 8 total — 2 posted, 6 skipped

### Posted findings
- [Medium] **New regression test's dispatch path is ambient rather than pinned, and the only CI row that runs it is DMA=1** — `test-refactor/client-server/wh_test_crypto_ed25519.c:413-420`
- [Low] **WC_PK_TYPE_ED25519_KEYGEN dispatch remains uncovered by the new regression test** — `test-refactor/client-server/wh_test_crypto_ed25519.c:403-411`

<details><summary>Skipped findings</summary>

- [Medium] Legacy NOCURVE25519 knob has no CI row and no test that would catch the regression
- [Low] File-header test inventory not updated with the new test function
- [Low] New test duplicates ~60 lines of setup/teardown from _whTest_CryptoEd25519ServerKey
- [Low] New NOCURVE25519 knob in test/Makefile is never exercised by CI
- [Low] Regression test does not pin the cryptocb dispatch mode it is guarding
- [Info] File header comment enumerates test functions and is now stale

</details>

*Review generated by Skoll via Claude/Codex*

test-refactor/client-server/wh_test_crypto_ed25519.c
yosuke-wolfssl: Thanks — took both points.

**[Medium] Pinned dispatch mode** — done: `(void)wh_Client_SetDmaMode(ctx, 0)` at the top of the test, so the whole body is pinned. Confirming your inspection empirically, since the sandbox couldn't build: with the guard re-nested the test fails `rc=-173` in both `NOCURVE25519=1 ASAN=1` and `DMA=1 ASAN=1 NOCURVE25519=1`, so the `DMA=1` row was already reaching `wh_Client_CryptoCbStd`. Agreed it was fragile.

**[Low] ED25519_KEYGEN** — real gap, but not coverable as suggested: `wh_Client_Ed25519MakeExportKey` passes `WH_NVM_FLAGS_EPHEMERAL` and exports full local material (`src/wh_client_crypto.c:3892`), so the key is *not* a server handle afterward and that assertion would fail even with the fix. A bad-`keySz` call can't discriminate either — wolfSSL validates size before consulting the cryptocb. All three cases share one `#ifdef HAVE_ED25519`, so the regression class fixed here takes out all three and SIGN catches it; a KEYGEN-only regression needs a newly added separate guard, i.e. a different defect. Mandatory-dispatch coverage needs `WOLF_CRYPTO_CB_ONLY_ED25519`, which would also force the inline test's local-material sign through the callback — follow-up, not this PR.

Also fixed one skipped finding: the legacy `NOCURVE25519` knob had no CI row, so added `DMA=1 ASAN=1 NOCURVE25519=1` to `build-and-test.yml`. Stale file-header inventory updated. Leaving the ~60-line duplication finding — the shared logic is already in `whTest_Ed25519ImportToServer`, the rest is boilerplate every test in the file repeats.

Separate pre-existing gap you surfaced: no test reaches the `wh_Client_CryptoCbDma` Ed25519 group via the wolfCrypt API (`_whTest_CryptoEd25519Dma` calls `wh_Client_Ed25519SignDma`/`VerifyDma` directly). Follow-up issue rather than growing this PR.

test-refactor/client-server/wh_test_crypto_ed25519.c
yosuke-wolfssl: Replied by the above comment

Frauschi (approved):

PR #499, opened 2026-07-28
Implement SHA3 hash benchmarks (224/256/384/512)

kojiws:
## Summary

Enables the SHA3 benchmarks in `benchmark/bench_modules/wh_bench_mod_sha3.c`,
which were stubs returning `WH_ERROR_NOTIMPL` and silently `-> SKIPPED`.
Covers all four variants (224/256/384/512), DMA and non-DMA. SHA3 entries
switched from `BENCH_THROUGHPUT_NONE` to `XBPS` so throughput is reported.
HMAC-SHA3 is out of scope and stays `SKIPPED`.

## Test

`make DMA=1`: all eight SHA3 rows report; previously two were `SKIPPED` and
six did not exist. `make` (non-DMA): the four non-DMA rows report.

### DIFF
--- a/benchmark/bench_modules/wh_bench_mod_sha3.c
+++ b/benchmark/bench_modules/wh_bench_mod_sha3.c
@@ -20,28 +20,181 @@
 #include "wolfhsm/wh_error.h"
 
 #if !defined(WOLFHSM_CFG_NO_CRYPTO) && defined(WOLFHSM_CFG_BENCH_ENABLE)
+#include "wolfssl/wolfcrypt/hash.h"
+#include "wolfssl/wolfcrypt/sha3.h"
 
 #if defined(WOLFSSL_SHA3)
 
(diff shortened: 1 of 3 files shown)


benchmark/bench_modules/wh_bench_mod_sha3.c
```diff
@@ -20,29 +20,182 @@
...
+    /* Only free SHA3 if it was initialized */
+    if (sha3Initialized) {
+        v->freeFn(sha3);
+    }
+
+    return ret;
 }
 
+#define WH_DEFINE_SHA3_BENCH_NON_DMA_FNS(_Bits)                               \
+    int wh_Bench_Mod_Sha3##_Bits(whClientContext* client,                     \
```
bigbrett: I'd rather we skip the macro magic and just use real code here, as it doesn't really reduce the number of code that much. Please refactor to just use normal functions even if slightly more verbose.

bigbrett (changes requested):

bigbrett (approved):

PR #503, opened 2026-08-04
Add CRC integrity check to NVM flash state

bigbrett:
Adds optional CRC16 integrity checking to the nvm_flash backend’s on-flash object state so the directory load, full-object reads, and compaction copies can detect corruption (gated by WOLFHSM_CFG_NVM_FLASH_CRC16), and wires this option through tests, tooling, docs, and CI.

Changes:

- Add wh_Utils_Crc16() utility plus WH_UTILS_CRC16_INIT seed constant.
- Extend nvm_flash on-flash format to embed CRC16 values in object start/count state words and verify metadata/data at key read/copy points.
- Add CRC-enabled test coverage and build knobs for test, test-refactor, and whnvmtool, including CI jobs.


src/wh_nvm_flash.c
```diff
@@ -247,6 +281,18 @@ static int nfMemObject_Read(whNvmFlashContext* context,
             /* Copy the metadata out of the buffer */
             memcpy(&object->metadata, buffer, sizeof(object->metadata));
             clear_metadata = 0;
+#ifdef WOLFHSM_CFG_NVM_FLASH_CRC16
+            /* Verify the metadata against the CRC in the start state word.
+             * DATA_BAD entries are not demoted: their metadata length is
+             * still needed to reserve the partially written data area */
+            if ((object->state.status == NF_STATUS_USED) &&
+                (wh_Utils_Crc16(WH_UTILS_CRC16_INIT, &object->metadata,
```
bigbrett, line 293: duh...

src/wh_nvm_flash.c
```diff
@@ -797,6 +920,28 @@ static int nfMemDirectory_Parse(nfMemDirectory* d)
...
+             * later intact entry re-bounds the reservation with its own start
+             * and count. Reclaimable size is a best effort */
+            d->reclaimable_entries++;
+            if ((context->partition_units - NF_PARTITION_DATA_OFFSET) >
+                d->objects[d->next_free_object].state.start) {
+                d->reclaimable_data +=
+                    (context->partition_units - NF_PARTITION_DATA_OFFSET) -
+                    d->objects[d->next_free_object].state.start;
+            }
+            d->next_free_data = context->partition_units;
```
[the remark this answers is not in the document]
bigbrett: false positive

src/wh_nvm_flash.c
```diff
@@ -47,6 +48,19 @@ enum {
...
+#ifdef WOLFHSM_CFG_NVM_FLASH_CRC16
+/* With CRC16 enabled, the object start and count state words carry a CRC in
+ * bits [47:32], replacing the low half of the magic:
+ *   start word: [63:48]=0x1234 [47:32]=CRC16(metadata) [31:0]=start
+ *   count word: [63:48]=0x1234 [47:32]=CRC16(data)     [31:0]=count
+ * The epoch word and all partition state words keep the full BASE_STATE
+ * magic. The remaining 0x12/0x34 bytes still keep every state word distinct
+ * from erased flash. */
+static const whFlashUnit CRC_BASE_STATE = 0x1234000000000000ULL;
+#define NF_STATE_CRC_PACK(_crc) (((whFlashUnit)(_crc)) << 32)
```
billphipps:
```suggestion
#define NF_STATE_CRC_PACK(_crc) (((whFlashUnit)((uint16_t)(_crc)) << 32)
```

src/wh_nvm_flash.c
billphipps: Grumble.  The flash context should NOT be necessary to parse the directory that was already read in.  I agree with how you are handling this funky edge case, but this change really points out that we are missing some basic sanity checks in the start/count values anyway.

I recommend instead to add the max_count value into the memDirectory structure and set it within the ReadParseMemDirectory function, which needs the context value.  Then, this function won't need the flash context to figure out the max_count (AND can make sure that the next_free_data NEVER exceeds max_count for ALL cases.

There is a minor bug here also in that this loop continues to update the memDirectory values after a failure is detected.  That means that if the next object is good, the next_free_data WILL be set to the correct value, BUT the reclaimable_data that was computed here will be WRONG.  I don't see an easy way around this, but maybe just be aware that reclaimable_data is a measurement and not necessarily gospel.  

Also note that the deduplication loop below will also update reclaimable data after it is tweaked here.

Maybe we should consider a different method to compute/damage reclaimable data?  Seems like we should be able to fully account for the usage of every sector in the data area.  Dunno.

src/wh_nvm_flash.c
```diff
@@ -1027,11 +1188,22 @@ int wh_NvmFlash_GetAvailable(void* c,
...
     }
+    if (context->directory_bad != 0) {
+        return WH_ERROR_ABORTED;
+    }
     nfMemDirectory *d = &context->directory;
     if (out_avail_size != NULL) {
-        *out_avail_size = (context->partition_units -
-                NF_PARTITION_DATA_OFFSET - d->next_free_data) *
-                WHFU_BYTES_PER_UNIT;
+        uint32_t data_units =
```
billphipps: this could be replaced by memDirectory.max_count.  Or a better name.

wolfhsm/wh_nvm_flash.h
```diff
@@ -76,13 +83,16 @@ typedef struct whNvmFlashConfig_t {
...
-    int active;                     /* Which partition (0 or 1) is active */
-    int initialized;
+    const whFlashCb* cb;              /* Flash callbacks */
+    void*            flash;           /* Flash context to use */
+    nfMemState       state;           /* State of active partition */
+    nfMemDirectory   directory;       /* Cache of active objects */
+    uint32_t         partition_units; /* Size of partition in units */
+    int              active;          /* Which partition (0 or 1) is active */
+    int              initialized;
+    int              directory_bad; /* Directory could not be reloaded from
```
billphipps, line 95: Could this be eliminated and simply unset initialized?  Or initialized could become a state instead of boolean?  No fix.

billphipps (changes requested):
Looks great!  Can you modify the memDirectory per comments?

src/wh_nvm_flash.c
bigbrett: fixed

src/wh_nvm_flash.c
bigbrett: Ok refactored slightly such that the directory now carries its own size limit (`max_data`), filled in when the directory is read from flash, so the parse step no longer needs anything else to know how big the data area is. Parsing also caps the next free positon at that limit no matter what was read off the flash, as you suggested. Also fixed latent bug in existing code weher the "out of space" check in AddObject was checking against the whole partiiton not just the data area.

Not addressing the other stuff as it should probably be addressed on its own in a dedicated refactor PR if you have issues with how you originally implemented it?

src/wh_nvm_flash.c
bigbrett: fixed

wolfhsm/wh_nvm_flash.h
bigbrett: maybe, but not sure it really buys us that much to unify them?

billphipps (approved):
Thank you !  This looks great!

PR #505, opened 2026-08-06
Unify client response size checks into _getCryptoResponse

padelsbach:
Moves a handful of repeated hand-rolled size checks into a common function for DRYness.

### DIFF
--- a/src/wh_client_crypto.c
+++ b/src/wh_client_crypto.c
@@ -153,8 +153,10 @@ static uint8_t* _createCryptoRequestWithSubtype(uint8_t* reqBuf, uint16_t type,
                                                 uint16_t algoSubType,
                                                 uint32_t affinity);
 static int      _getCryptoResponse(uint8_t* respBuf, uint16_t type,
-                                   const uint16_t respLen, const uint16_t minRespLen,
-                                   uint8_t** outResponse, uint16_t* outTrailingLen);
+                                   const uint16_t respLen,
+                                   uint8_t**      outResponse,
+                                   const uint16_t minRespLen,
+                                   uint16_t*      outTrailingLen);
 
 
 /* Helper function to prepare a crypto request buffer with generic header */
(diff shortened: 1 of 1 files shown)


src/wh_client_crypto.c
```diff
@@ -5888,7 +5855,8 @@ int wh_Client_CmacUpdateResponse(whClientContext* ctx, Cmac* cmac)
         return ret;
     }
 
-    ret = _getCryptoResponse(dataPtr, WC_ALGO_TYPE_CMAC, (uint8_t**)&res);
+    ret = _getCryptoResponse(dataPtr, WC_ALGO_TYPE_CMAC, res_len, sizeof(*res),
+                             (uint8_t**)&res, NULL);
     if (ret >= 0) {
         /* No trailing payload on update, but the state must be in the frame */
         if (res_len < sizeof(whMessageCrypto_GenericResponseHeader) +
```
padelsbach: fixed

yosuke-wolfssl (commented):
Good cleanup —  Comment is about shape, not correctness.

### The helper's purpose changes here

`_getCryptoResponse()` was a message-translation helper: check `algoType`, return a pointer past the generic header, propagate `rc`. It had no length parameter. This PR makes it the frame-bounds authority for the whole client crypto layer. 

My concern is that it centralizes *both* the generic check (header present, right algorithm, bytes received) and the per-message check. Since the helper can't know the latter, the caller passes it in as `minRespLen` — so `sizeof(*res)` ends up two arguments away from the `(uint8_t**)&res` it describes, with a cast in between that erases the type. Nothing catches a mismatch, and at 44 of the 74 sites no length check is visible locally at all.

### Suggested signature

```c
static int _getCryptoResponse(uint8_t* respBuf, uint16_t type,
                              uint16_t respLen, uint8_t** outResponse,
                              uint16_t* outBodyLen);
```

Helper does the generic half and reports `bodyLen = respLen - sizeof(*header)`; each caller keeps a local `if (bodyLen < sizeof(*res)) return WH_ERROR_ABORTED;`. Same DRY win, but the `sizeof` stays next to the pointer it bounds. The AES sites read better too: `bodyLen < sizeof(*res) + AES_IV_SIZE` locally, instead of a mandatory trailer hidden in `minRespLen`.

It also removes the `const` on the by-value params and the `size_t`→`uint16_t` narrowing at every call site. Either way, the helper now needs a real doc comment — the one-liner plus `TODO` no longer covers it.

PR #508, opened 2026-08-06
CI: Re-enable clang-format-check workflow

padelsbach:
Fixes an issue where the `git clang-format` wrapper was picking up a different version than claimed, and not printing the diff, making it challenging to know what actually failed. 

The guts of the clang-format yaml is moved to a script in `.githooks` so that it can be used locally as a pre-commit hook. Enable it with the following command (also in the readme):
```
    sudo apt-get install -y clang-format-18
    git config core.hooksPath .githooks
```

There are about 13k lines of source code which could be re-formatted, but not touching that for now. 

If this change breaks or causes annoyance, we will disable or drop this workflow.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec "$(git rev-parse --show-toplevel)/.github/scripts/clang-format-check.sh"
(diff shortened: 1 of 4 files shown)


bigbrett (changes requested):
@padelsbach I don't like pre-commit hooks, they end up always being really annoying. I'd rather it just fail in CI vs locally since things like formatting WIP commits intended to be squashed later that might not be syntatically correct can screw things up. 

Can you just scrap the hook but keep the rest?

PR #510, opened 2026-08-12
Fix bug in data unwrap. Remove unnecessary stack buffers in key wrap code

AlexLanzano:
Bug fix
  - _AesGcmDataUnwrapWithKek decrypts into a staging plainData[] instead of dataOut, which overlapped the ciphertext in the shared comm buffer. Fixes silent corruption of any wh_Client_DataUnwrap payload >= 32 bytes.
  - Added encBlobSz > dataSz bound check.

  Data wrap restructure
  - _AesGcmDataWrapWithKek now stages into plainData[] and rejects dataSz > MAX_DATA_SIZE, so the helper owns the no-alias requirement rather than trusting callers.
  - _HandleDataWrapRequest drops its 2 KB data[] and uses reqData directly.
  - Both data helpers ForceZero the staging buffer and return the wolfSSL error instead of discarding it.

  Stack buffers removed (~6 KB)
  - _HandleKeyWrapRequest: key[] becomes a pointer into reqData.
  - _HandleKeyUnwrapAndCacheRequest: key[] becomes a pointer into reqData. Scrub now uses keySz, since sizeof on a pointer would have wiped only 8 bytes.
  - _HandleKeyWrapExportRequest: key[] reads into respData. keySz bounded by min(respDataSz, MAX_KEY_SIZE), and the scrub is failure-only so it cannot clobber the wrapped output.

  Tests
  - Added a data wrap/unwrap round-trip sweep over sizes 1 to 128, spanning the AES block boundary.
  - Enlarged the two legacy data-wrap payloads from 14 to 63 bytes. The old sub-block payload is why the corruption went unnoticed.

### DIFF
--- a/src/wh_client_keywrap.c
+++ b/src/wh_client_keywrap.c
@@ -599,9 +599,10 @@ int wh_Client_DataUnwrapRequest(whClientContext*   ctx,
         return WH_ERROR_BADARGS;
     }
 
-    /* Bound the whole wire length before copying into the comm data buffer */
+    /* Bound the wire length; the blob adds the wrap header to the plaintext */
     if (wrappedDataInSz == 0 ||
-        wrappedDataInSz > WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE ||
+        wrappedDataInSz > (uint32_t)WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE +
+                              WH_KEYWRAP_AES_GCM_HEADER_SIZE ||
         (size_t)sizeof(*req) + wrappedDataInSz > WOLFHSM_CFG_COMM_DATA_LEN) {
         return WH_ERROR_BADARGS;
     }
--- a/src/wh_client_keywrap.c
+++ b/src/wh_client_keywrap.c
@@ -253,7 +253,8 @@ int wh_Client_KeyUnwrapAndExportRequest(whClientContext*   ctx,
     }
 
     /* Bound the whole wire length before copying into the comm data buffer */
-    if (wrappedKeySz == 0 || wrappedKeySz > WOLFHSM_CFG_KEYWRAP_MAX_KEY_SIZE ||
+    if (wrappedKeySz == 0 ||
+        wrappedKeySz > WH_KEYWRAP_AES_GCM_MAX_WRAPPED_KEY_SIZE ||
         (size_t)sizeof(*req) + wrappedKeySz > WOLFHSM_CFG_COMM_DATA_LEN) {
         return WH_ERROR_BADARGS;
     }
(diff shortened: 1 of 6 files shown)


src/wh_server_keystore.c
```diff
@@ -1701,8 +1701,10 @@ static int _AesGcmDataWrapWithKek(whServerContext* server,
     uint8_t  authTag[WH_KEYWRAP_AES_GCM_TAG_SIZE];
     uint8_t  iv[WH_KEYWRAP_AES_GCM_IV_SIZE];
     uint8_t* encBlob;
+    uint8_t  plainData[WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE];
```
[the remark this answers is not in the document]
AlexLanzano: This is not a terrible idea. It would be nice to be able to have this on the heap if that is present. However, I think this is out of scope for the current PR here

src/wh_server_keystore.c
```diff
@@ -1805,6 +1809,10 @@ static int _AesGcmDataUnwrapWithKek(whServerContext* server,
     encBlob   = (uint8_t*)wrappedDataIn + sizeof(iv) + sizeof(authTag);
     encBlobSz = wrappedDataSz - sizeof(iv) - sizeof(authTag);
 
+    if (encBlobSz > dataSz) {
+        return WH_ERROR_BUFFER_SIZE;
+    }
```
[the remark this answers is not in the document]
AlexLanzano: Yes there is...

AlexLanzano:
CI will fail until https://github.com/wolfSSL/wolfHSM/pull/511 is merged

src/wh_server_keystore.c
```diff
@@ -1792,6 +1795,7 @@ static int _AesGcmDataUnwrapWithKek(whServerContext* server,
     uint8_t  iv[WH_KEYWRAP_AES_GCM_IV_SIZE];
     uint8_t* encBlob;
     uint16_t encBlobSz;
+    uint8_t  plainData[WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE];
```
AlexLanzano: This is necessary. We need a staging area in the case where the req and resp buffers are the same

src/wh_client_keywrap.c
```diff
@@ -599,9 +599,10 @@ int wh_Client_DataUnwrapRequest(whClientContext*   ctx,
         return WH_ERROR_BADARGS;
     }
 
-    /* Bound the whole wire length before copying into the comm data buffer */
+    /* Bound the wire length; the blob adds the wrap header to the plaintext */
     if (wrappedDataInSz == 0 ||
-        wrappedDataInSz > WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE ||
+        wrappedDataInSz > (uint32_t)WOLFHSM_CFG_KEYWRAP_MAX_DATA_SIZE +
```
bigbrett, line 606: Looks like key unwrap (untouched by this PR) has the same bug you fix for data wrap in this PR. `KeyUnwrapAndExportRequest`/`KeyUnwrapAndCacheRequest` still bound the *blob* by `WOLFHSM_CFG_KEYWRAP_MAX_KEY_SIZE`, yet a wrapped blob is keySz + header + metadata (+60 bytes).

Scenario (default config, max = 2000): wrap a 2000-byte key → succeeds, returns a 2060-byte blob → pass that blob to `wh_Client_KeyUnwrapAndExport` → client rejects it with BADARGS before sending, since 2060 > 2000. The key round-trips through wrap but can never be unwrapped. Any key over 1940 bytes hits this.

Should use the new `WH_KEYWRAP_AES_GCM_MAX_WRAPPED_KEY_SIZE` macro here (the existing COMM_DATA_LEN check still bounds the wire, so don't tighten the static asserts — that would break small-comm configs). The `WH_TEST_KW_OVER_KEY` oversize tests pin the old bound and need updating too.

AlexLanzano: fixed

bigbrett (approved):

PR #512, opened 2026-08-14
Add crypto callback mode for SLH-DSA

padelsbach:


test-refactor/client-server/wh_test_crypto_slhdsa.c
```diff
@@ -0,0 +1,1007 @@
...
+#include "wolfhsm/wh_common.h"
+#include "wolfhsm/wh_client.h"
+#include "wolfhsm/wh_client_crypto.h"
+
+#include "wh_test_common.h"
+#include "wh_test_list.h"
+
+#ifdef WOLFSSL_HAVE_SLHDSA
+
+#if defined(WOLFSSL_SLHDSA_PARAM_128S) && !defined(WOLFSSL_SLHDSA_VERIFY_ONLY)
```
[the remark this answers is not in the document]
padelsbach: fixed with `WH_TEST_SKIPPED`

src/wh_client_cryptocb.c
```diff
@@ -1313,6 +1371,34 @@ static int _handlePqcSign(whClientContext* ctx, wc_CryptoInfo* info, int useDma)
...
+#ifdef WOLFSSL_HAVE_SLHDSA
+        case WC_PQC_SIG_TYPE_SLHDSA: {
+            const byte* addRnd     = info->pk.pqc_sign.addRnd;
+            byte        addRndSz   = info->pk.pqc_sign.addRndSz;
+            int         randomized = (info->pk.pqc_sign.rng != NULL);
+            int         isMPrime =
+                (info->pk.type == WC_PK_TYPE_PQC_SIG_SIGN_MSG);
+
+            if (!_SlhDsaKeyHasMaterial((const SlhDsaKey*)key)) {
+                addRnd   = NULL;
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1417 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+                        }
+                        else {
+                            /* Update signature length */
+                            *out_len = res->sigLen;
+                        }
+                    }
+                }
+            }
+
+            (void)wh_Client_DmaProcessClientAddress(
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_server_crypto.c
```diff
@@ -7056,282 +8240,230 @@ static int _HandleMlDsaSignDma(whServerContext* ctx, uint16_t magic, int devId,
...
+                           ctx, key, req.options, (const byte*)msgAddr,
+                           (word32)req.msg.sz, req_context, contextSz,
+                           preHashType, req_addRnd, addRndSz, (byte*)sigAddr,
+                           &sigLen);
                 }
 
                 if (sigAddr != NULL) {
-                    /* Post-write processing of signature buffer */
                     (void)wh_Server_DmaProcessClientAddress(
                         ctx, (uintptr_t)req.sig.addr, &sigAddr, sigLen,
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_server_crypto.c
```diff
@@ -6894,159 +8073,164 @@ static int _HandleMlDsaKeyGenDma(whServerContext* ctx, uint16_t magic,
...
-                                    WH_DMA_OPER_CLIENT_WRITE_POST,
-                                    (whServerDmaFlags){0});
-                            }
-                            else {
-                                ret = rc;
-                            }
-                            if (ret != 0) {
-                                _CryptoEvictKeyLocked(ctx, keyId);
-                            }
+                if (ret == 0) {
```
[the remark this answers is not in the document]
padelsbach: fixed

test-refactor/client-server/wh_test_crypto_slhdsa.c
```diff
@@ -0,0 +1,1007 @@
...
+done:
+    wc_SlhDsaKey_Free(key);
+    return ret;
+}
+#endif /* 192S || 256S */
+#endif /* WH_TEST_SLHDSA_COMM_PARAM */
+
+int whTest_Crypto_SlhDsa(whClientContext* ctx)
+{
+#ifdef WH_TEST_SLHDSA_COMM_PARAM
```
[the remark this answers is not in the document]
padelsbach: New fixture is out of scope for this PR

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1417 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+                        }
+                        else {
+                            /* Set verification result */
+                            *out_res = res->verifyResult;
+                        }
+                    }
+                }
+            }
+
+            (void)wh_Client_DmaProcessClientAddress(
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1417 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+    return WH_ERROR_OK;
+}
+
+int wh_Client_SlhDsaImportKey(whClientContext* ctx, SlhDsaKey* key,
+                              whKeyId* inout_keyId, whNvmFlags flags,
+                              uint16_t label_len, uint8_t* label)
+{
+    int      ret    = WH_ERROR_OK;
+    whKeyId  key_id = WH_KEYID_ERASED;
+    byte     buffer[WH_CRYPTO_SLHDSA_MAX_KEY_DER_SIZE];
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_client_crypto.c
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1417 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+        }
+        if (ret == WH_ERROR_OK) {
+            do {
+                ret = wh_Client_RecvResponse(ctx, &group, &action, &res_len,
+                                             WOLFHSM_CFG_COMM_DATA_LEN,
+                                             (uint8_t*)dataPtr);
+            } while (ret == WH_ERROR_NOTREADY);
+        }
+
+        if (seedSz > 0) {
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1472 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+            if (ret == WH_ERROR_OK) {
+                ret = postRet;
+            }
+        }
+        if (keyPre) {
+            int postRet = wh_Client_DmaProcessClientAddress(
+                ctx, (uintptr_t)buffer, (void**)&keyAddr, keyAddrSz,
+                WH_DMA_OPER_CLIENT_WRITE_POST, (whDmaFlags){0});
+            if (ret == WH_ERROR_OK) {
+                ret = postRet;
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_server_crypto.c
```diff
@@ -7056,282 +8245,238 @@ static int _HandleMlDsaSignDma(whServerContext* ctx, uint16_t magic, int devId,
...
                 if (sigAddr != NULL) {
-                    /* Post-write processing of signature buffer */
-                    (void)wh_Server_DmaProcessClientAddress(
-                        ctx, (uintptr_t)req.sig.addr, &sigAddr, sigLen,
-                        WH_DMA_OPER_CLIENT_WRITE_POST,
-                        (whServerDmaFlags){0});
+                    /* Release the range PRE mapped, not the shorter signature
+                     * that came back: sigLen travels in the response. */
+                    int postRet = wh_Server_DmaProcessClientAddress(
+                        ctx, (uintptr_t)req.sig.addr, &sigAddr, req.sig.sz,
```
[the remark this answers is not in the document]
padelsbach: fixed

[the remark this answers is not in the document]
padelsbach: same as above, fixed

src/wh_server_crypto.c
```diff
@@ -6894,159 +8073,169 @@ static int _HandleMlDsaKeyGenDma(whServerContext* ctx, uint16_t magic,
...
                         }
-#endif /* WOLFSSL_MLDSA_PUBLIC_KEY */
-                        if (ret == 0) {
-                            res.keyId   = wh_KeyId_TranslateToClient(keyId);
-                            res.keySize = keySize;
+                        else {
+                            ret = (pub_ret < 0) ? pub_ret : WH_ERROR_ABORTED;
                         }
+                        (void)wh_Server_DmaProcessClientAddress(
+                            ctx, req.key.addr, &clientOutAddr, keySize,
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1472 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+                            if (der_size == 0) {
+                                ret = WH_ERROR_ABORTED;
+                            }
+                            else if ((res_len < hdr_sz) ||
+                                     (res->len > (res_len - hdr_sz))) {
+                                ret = WH_ERROR_ABORTED;
+                            }
+                            else {
+                                uint8_t* key_der = (uint8_t*)(res + 1);
+                                ret = wh_Crypto_SlhDsaDeserializeKeyDer(
```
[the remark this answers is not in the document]
padelsbach: fixed

src/wh_server_crypto.c
```diff
@@ -5581,46 +5813,84 @@ static int _HandleMlKemKeyGen(whServerContext* ctx, uint16_t magic, int devId,
...
+        }
+
         if (ret == 0) {
-            if ((req.flags & WH_NVM_FLAGS_EPHEMERAL) != 0) {
+            if (flags & WH_NVM_FLAGS_EPHEMERAL) {
+                /* Must serialize the key into the response message. */
                 key_id = WH_KEYID_ERASED;
-                ret = wh_Crypto_MlKemSerializeKey(key, max_size, res_out,
-                                                  &res_size);
+                ret    = wh_Crypto_SlhDsaSerializeKeyDer(key, max_size, res_out,
```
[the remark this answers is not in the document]
padelsbach: Added zeroization on failure path, which is the pre-existing pattern

test-refactor/client-server/wh_test_crypto_slhdsa.c
```diff
@@ -0,0 +1,1038 @@
...
+    WH_TEST_RETURN_ON_FAIL(_whTest_CryptoSlhDsaBufferTooSmall(ctx));
+#ifndef NO_SHA256
+    WH_TEST_RETURN_ON_FAIL(_whTest_CryptoSlhDsaBadDigestLen(ctx));
+#endif
+#ifdef WH_TEST_SLHDSA_OVERSIZE_PARAM
+    WH_TEST_RETURN_ON_FAIL(_whTest_CryptoSlhDsaCommBufferLimit(ctx));
+#endif
+#endif /* WH_TEST_SLHDSA_COMM_PARAM */
+
+#ifdef WH_TEST_SLHDSA_FAST_PARAM
```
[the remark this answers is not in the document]
padelsbach: added WH_TEST_SLHDSA_DMA_PARAM, which prefers 128F but falls back to 128S

src/wh_client_cryptocb.c
```diff
@@ -564,8 +565,10 @@ int wh_Client_CryptoCbStd(int devId, wc_CryptoInfo* info, void* inCtx)
 
 #endif /* WOLFSSL_HAVE_LMS || WOLFSSL_HAVE_XMSS */
 
-#if defined(WOLFSSL_HAVE_MLDSA) || defined(HAVE_FALCON)
+#if defined(WOLFSSL_HAVE_MLDSA) || defined(HAVE_FALCON) || \
+    defined(WOLFSSL_HAVE_SLHDSA)
         case WC_PK_TYPE_PQC_SIG_KEYGEN:
+        case WC_PK_TYPE_PQC_SIG_KEYGEN_SEED:
```
Frauschi, line 574: The new `case WC_PK_TYPE_PQC_SIG_KEYGEN_SEED:` here and in `wh_Client_CryptoCbDma` sits inside the broad `MLDSA || FALCON || SLHDSA` guard, while `WC_PK_TYPE_PQC_SIG_SIGN_MSG` and `VERIFY_MSG` twenty lines down are wrapped in `#ifdef WOLFSSL_HAVE_SLHDSA`. Can we make this one consistent with those?

It matters because `_handlePqcSigKeyGen` reads `info->pk.pqc_sig_kg.seed` only in the SLH-DSA branch - the ML-DSA branch ignores it and calls `wh_Client_MlDsaMakeExportKey()`, which is a random keygen. Nothing reaches that today, since ML-DSA and Falcon both go through the non-`Ex` `wc_CryptoCb_MakePqcSignatureKey` and only SLH-DSA emits this pk type, so this is hardening rather than a live bug. But a build can enable both, so I would also return `CRYPTOCB_UNAVAILABLE` from the ML-DSA branch for the SEED type, so it fails closed instead of handing back a key that does not match the caller's seed.

src/wh_client_crypto.c
```diff
@@ -10518,8 +10518,12 @@ int wh_Client_MlDsaVerify(whClientContext* ctx, const byte* sig, word32 sig_len,
         uint16_t action  = WC_ALGO_TYPE_PK;
```
Frauschi, line 10519: `wh_Client_MlDsaVerify` just got its `total_len` widened to 64 bits, with a comment about caller-supplied lengths wrapping past the buffer check. `wh_Client_MlDsaSign` right above still has `uint32_t total_len = sizeof(hdr) + sizeof(*req) + in_len + contextLen;` - same wrap, same guard, 140 lines apart.

`in_len` comes from `info->pk.pqc_sign.inlen`, so this is local API-misuse hardening rather than anything remotely reachable. Since the PR is already fixing that exact class, worth closing the sibling in the same change.

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1485 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+    if ((ctx == NULL) || (key == NULL) ||
+        ((label_len != 0) && (label == NULL))) {
+        return WH_ERROR_BADARGS;
+    }
+
+    if (inout_keyId != NULL) {
+        key_id = *inout_keyId;
+    }
+
+    ret = wh_Crypto_SlhDsaSerializeKeyDer(key, sizeof(buffer), buffer,
```
Frauschi, line 11296: The `wc_ForceZero(buffer, sizeof(buffer))` here only covers half the exposure. `wh_Client_KeyCacheRequest_ex` memcpys the same private-key DER into the shared transport buffer from `wh_CommClient_GetDataPtr`, and nothing clears that. `_SlhDsaMakeKey` in this PR does handle it, with `wc_ForceZero(dataPtr, WOLFHSM_CFG_COMM_DATA_LEN)`, so the two new paths disagree.

It is not reliably overwritten by what follows either: the DER runs to about 320 bytes and a following `CheckPrivKey` writes roughly 88, so the tail survives in a buffer the server side can see under a shared-memory transport. `wh_Client_SlhDsaImportKeyDma` has the same gap through `wh_Client_KeyCacheDma`.

Frauschi (commented):
Went through the SLH-DSA callback support. Big addition, and the structure mirrors the ML-DSA paths closely enough to follow easily.

Comments inline. The ones I would most like to see addressed are the DMA `READ_POST` after a denied `READ_PRE` in `_HandleSlhDsaVerifyDma`, the unvalidated `seedSz / 3` in both keygen handlers, and the seed left resident in the comm packet. The rest are smaller consistency points against the ML-DSA and ML-KEM siblings, and the last few are nits you can take or leave.

Nothing here is a blocker.

src/wh_client_crypto.c
```diff
@@ -11236,6 +11240,1485 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+                         label, pub);
+    if (ret >= 0) {
+        /* Stamp the cached keyId and the client's HSM devId so pub is
+         * immediately usable as a handle to the cached private key. The keyId
+         * is set here as well because the public-key deserialize re-inits
+         * pub and clears it. */
+        wh_Client_SlhDsaSetKeyId(pub, *inout_key_id);
+        pub->devId = WH_CLIENT_DEVID(ctx);
+    }
+    else if (!WH_KEYID_ISERASED(*inout_key_id) &&
```
Frauschi: This rollback condition is copied from `wh_Client_MlDsaMakeCacheKeyAndExportPublic`, whose comment states the premise the `ret == WH_ERROR_ABORTED` clause rests on: `_MlDsaMakeKey` only returns ABORTED after the server has committed. This PR adds a new ABORTED path to `_SlhDsaMakeKey` - the response-frame length check - which fires before `res->keyId` is read and before `*inout_key_id` is updated, so that premise no longer holds. On a short or malformed response with a caller-supplied key id, this evicts a pre-existing key the operation never touched and overwrites the caller's id with `WH_KEYID_ERASED`.

The frame check itself is a good addition. Could we narrow the condition so the ABORTED clause only fires after the commit - latch a flag once `res->keyId` has been read, and test that instead? Separately, `wh_Client_SlhDsaMakeCacheKeyDma` uses a different condition, which is the correct one for that path, but nothing says why; one line there would stop someone later "fixing" the two to match.

test-refactor/client-server/wh_test_crypto_slhdsa.c
```diff
@@ -0,0 +1,1051 @@
...
+#include "wolfhsm/wh_common.h"
+#include "wolfhsm/wh_client.h"
+#include "wolfhsm/wh_client_crypto.h"
+
+#include "wh_test_common.h"
+#include "wh_test_list.h"
+
+#ifdef WOLFSSL_HAVE_SLHDSA
+
+#if defined(WOLFSSL_SLHDSA_PARAM_128S) && !defined(WOLFSSL_SLHDSA_VERIFY_ONLY)
```
Frauschi, line 53: Two gaps in the new suite.

The verify-only job only proves the configuration compiles. `WH_TEST_SLHDSA_COMM_PARAM` and `_FAST_PARAM` are both defined only when `!defined(WOLFSSL_SLHDSA_VERIFY_ONLY)`, so under that config every block in `whTest_Crypto_SlhDsa` compiles out and it returns `WH_TEST_SKIPPED`. Nothing exercises verification against a cached public key, which is the one path a verify-only build actually ships. Could we add a case that caches a known public key and asserts a known-good signature verifies and a tampered one does not, gated on `WOLFSSL_HAVE_SLHDSA` alone rather than the signing parameter macros?

Second, about half the new public API has no direct caller anywhere: `GetKeyId`, `ExportKey`, `ExportPublicKey` (only the Dma variant is covered), `MakeCacheKey`, `MakeCacheKeyFromSeed`, `ImportKeyDma`, `ExportKeyDma`, `MakeExportKeyFromSeedDma`, `CheckPrivKeyDma`. The DER round trip through `wh_Crypto_SlhDsaSerializeKeyDer`/`Deserialize` and `_SlhDsaMakeKeyDma`'s deferred `postErr` handling are reached only indirectly.

src/wh_server_crypto.c
```diff
@@ -7056,282 +8257,250 @@ static int _HandleMlDsaSignDma(whServerContext* ctx, uint16_t magic, int devId,
...
-                                      key);
+    ret = _SlhDsaInitForCachedKey(key, req.param, devId);
     if (ret == 0) {
-        /* Process client signature buffer address */
-        ret = wh_Server_DmaProcessClientAddress(
-            ctx, (uintptr_t)req.sig.addr, &sigAddr, req.sig.sz,
-            WH_DMA_OPER_CLIENT_READ_PRE, (whServerDmaFlags){0});
-
+        ret = _SlhDsaLoadKey(ctx, key_id, WH_NVM_FLAGS_USAGE_VERIFY, key);
         if (ret == 0) {
```
Frauschi, line 8460: The sig `READ_POST` at the bottom is outside the `if (ret == 0)` that gates the sig `READ_PRE`, so a denied PRE still drives a POST against the address the policy just rejected. `wh_Server_DmaProcessClientAddress` assigns `*xformedCliAddr` before it invokes the integrator callback and before the allow-list check, so `sigAddr` is still non-NULL after `WH_ERROR_ACCESS` and the `if (sigAddr != NULL)` guard does not catch it. `msgAddr` has the same shape.

`_HandleSlhDsaSignDma` in this same PR already does this correctly with explicit `msgPre`/`sigPre` flags - can we use that pattern here? `_HandleMlDsaVerifyDma` also nests its sig POST inside the PRE guard, so this is a regression against both.

src/wh_server_crypto.c
```diff
@@ -5581,46 +5813,90 @@ static int _HandleMlKemKeyGen(whServerContext* ctx, uint16_t magic, int devId,
...
-    if (!_IsMlKemLevelSupported((int)req.level)) {
+    if (0 == _IsSlhDsaParamSupported(param)) {
         return WH_ERROR_BADARGS;
     }
 
-    ret = wc_MlKemKey_Init(key, (int)req.level, NULL, devId);
+    ret = wc_SlhDsaKey_Init(key, (enum SlhDsaParam)param, NULL, devId);
     if (ret == 0) {
-        ret = wc_MlKemKey_MakeKey(key, ctx->crypto->rng);
+        if (seedSz > 0) {
```
Frauschi, line 5876: `word32 n = seedSz / 3;` takes the per-component length purely from the client's own total, validated only by `(seedSz % 3) != 0`. `n` then goes in as all three component lengths to `wc_SlhDsaKey_MakeKeyWithRandom()` with no check against `key->params->n`, even though `wc_SlhDsaKey_Init` was given `req.param` two lines up so the correct value is right there. FIPS 205 admits only n = 16, 24 or 32.

Can we compare `n` against `key->params->n` and reject a mismatch? Otherwise the non-DMA path is bounded only by the remaining comm payload. Same shape in the DMA handler.

src/wh_server_crypto.c
```diff
@@ -5538,40 +5635,175 @@ static int _HandleMlDsaCheckPrivKey(whServerContext* ctx, uint16_t magic,
...
+         * request and is propagated so the caller can tell the two apart the
+         * same way a software-only build would. */
+        *out_result = 0;
+        ret         = 0;
+    }
+
+    return ret;
+}
+
+static int _HandleSlhDsaKeyGen(whServerContext* ctx, uint16_t magic, int devId,
```
Frauschi, line 5802: The client-supplied seed is read straight out of the comm packet and never scrubbed. Request and response share the same `whCommServer` buffer inside `whServerContext` - `wh_Server_HandleRequestMessage` passes `data` for both - so this is long-lived process memory, not a stack frame. The 3n-byte blob is `SK.seed || SK.prf || PK.seed`, which reproduces the entire private key.

Several paths return before any response bytes are written over it: the `seedSz % 3` rejection, the `_IsSlhDsaParamSupported` failure, `wc_SlhDsaKey_Init` or `MakeKeyWithRandom` failing, and the NVM-lock / unique-id / cache-import failures. Worth a `wc_ForceZero` over the seed once it has been consumed.

wolfhsm/wh_message_crypto.h
```diff
@@ -1193,6 +1193,142 @@ int wh_MessageCrypto_TranslateMlDsaVerifyResponse(
...
+    uint32_t keyId;
+    uint32_t pubSz; /* PK.seed||PK.root, 2n bytes */
+    /* Data follows:
+     * uint8_t pub[pubSz];
+     */
+} whMessageCrypto_SlhDsaCheckPrivKeyRequest;
+
+/* SLH-DSA Check Private Key Response */
+typedef struct {
+    uint32_t res;
```
Frauschi: `res` is `uint32_t` but carries `WC_KEY_MISMATCH_E`, which is negative: the server does `res.res = (result != 0) ? 0 : WC_KEY_MISMATCH_E;` and the client reads it back as `ret = (int)res->res;`. It round-trips in practice, but converting an out-of-range `uint32_t` back to `int` is implementation-defined. Could this be `int32_t`?

src/wh_server_crypto.c
```diff
@@ -1030,6 +1030,103 @@ static int _MlDsaKeyCacheExportEnforce(whServerContext* ctx, whKeyId keyId,
...
+                       "SLH-DSA key DER");
+
+int wh_Server_SlhDsaKeyCacheImport(whServerContext* ctx, SlhDsaKey* key,
+                                   whKeyId keyId, whNvmFlags flags,
+                                   uint16_t label_len, uint8_t* label)
+{
+    int            ret = WH_ERROR_OK;
+    uint8_t*       cacheBuf;
+    whNvmMetadata* cacheMeta;
+    uint16_t       der_size;
```
Frauschi: `der_size` is uninitialized here. It is only read under `ret == WH_ERROR_OK`, so the serializer will have written it, but initializing to 0 at the declaration saves a reader - and any analyzer - the trace.

src/wh_crypto.c
```diff
@@ -387,6 +387,67 @@ int wh_Crypto_MlDsaDeserializeKeyDer(const uint8_t* buffer, uint16_t size,
 }
 #endif /* WOLFSSL_HAVE_MLDSA */
 
+#ifdef WOLFSSL_HAVE_SLHDSA
+int wh_Crypto_SlhDsaSerializeKeyDer(SlhDsaKey* key, uint16_t max_size,
```
Frauschi, line 391: `wh_Crypto_SlhDsaSerializeKeyDer` hands the caller's buffer to `wc_SlhDsaKey_KeyToDer` and, when the encoder fails, returns the error with the buffer untouched. An encoder that writes a header and then finds it is out of room leaves partial private-key bytes behind. `wh_Crypto_MlKemSerializeKey` in the same file scrubs for exactly this, with `else { wc_ForceZero(buffer, keySize); }`.

Two call sites do not compensate. `_HandleSlhDsaKeyGenDma` serializes into the client-supplied DMA address and proceeds straight to `WRITE_POST` on failure, while its own non-DMA twin `_HandleSlhDsaKeyGen` does `if (ret != 0) { wc_ForceZero(res_out, max_size); res_size = 0; }`. `wh_Server_SlhDsaKeyCacheImport` serializes into a cache slot. Fixing the helper covers both - scrub `max_size` rather than the returned length, since on failure the number of bytes written is unknown.

wolfhsm/wh_message_crypto.h
```diff
@@ -1193,6 +1193,142 @@ int wh_MessageCrypto_TranslateMlDsaVerifyResponse(
...
+
+int wh_MessageCrypto_TranslateSlhDsaSignResponse(
+    uint16_t magic, const whMessageCrypto_SlhDsaSignResponse* src,
+    whMessageCrypto_SlhDsaSignResponse* dest);
+
+/* SLH-DSA Verify Request */
+typedef struct {
+    uint32_t options;
+#define WH_MESSAGE_CRYPTO_SLHDSA_VERIFY_OPTIONS_EVICT (1 << 0)
+#define WH_MESSAGE_CRYPTO_SLHDSA_VERIFY_OPTIONS_EXPORTPUB (1 << 1)
```
Frauschi: `WH_MESSAGE_CRYPTO_SLHDSA_VERIFY_OPTIONS_EXPORTPUB` is defined but never referenced. Leftover, or is a handler meant to be reading it?

src/wh_server_crypto.c
```diff
@@ -5630,53 +5906,42 @@ static int _HandleMlKemKeyGen(whServerContext* ctx, uint16_t magic, int devId,
...
 
     if (ret == WH_ERROR_OK) {
         res.keyId = wh_KeyId_TranslateToClient(key_id);
         res.len   = res_size;
-        (void)wh_MessageCrypto_TranslateMlKemKeyGenResponse(
-            magic, &res, (whMessageCrypto_MlKemKeyGenResponse*)cryptoDataOut);
-        *outSize = sizeof(whMessageCrypto_MlKemKeyGenResponse) + res_size;
-    }
 
+        wh_MessageCrypto_TranslateSlhDsaKeyGenResponse(magic, &res,
```
Frauschi: These translate calls are missing the `(void)` and the explicit struct cast that every sibling handler uses. Worth matching for consistency.

test-refactor/wh_test_list.c
```diff
@@ -75,6 +75,7 @@ WH_TEST_DECL(whTest_Crypto_MlDsa);
 WH_TEST_DECL(whTest_Crypto_Rng);
 WH_TEST_DECL(whTest_Crypto_Rsa);
 WH_TEST_DECL(whTest_Crypto_Sha);
+WH_TEST_DECL(whTest_Crypto_SlhDsa);
```
Frauschi: `whTest_Crypto_SlhDsa` is out of alphabetical order in the list.

src/wh_client_cryptocb.c
padelsbach: fixed

src/wh_client_crypto.c
padelsbach: applied

src/wh_client_crypto.c
padelsbach: fixed

src/wh_client_crypto.c
padelsbach: Good catch. Fixed

src/wh_server_crypto.c
padelsbach: Fixed with new flag

src/wh_server_crypto.c
padelsbach: Added

test-refactor/wh_test_list.c
padelsbach: fixed

wolfhsm/wh_message_crypto.h
padelsbach: removed

test-refactor/client-server/wh_test_crypto_slhdsa.c
padelsbach: Added verify-only test coverage. Normal pattern for WolfHSM does not include all the intermediate APIs (even though many are public)

src/wh_server_crypto.c
padelsbach: Added check

src/wh_server_crypto.c
padelsbach: Added zeroization

wolfhsm/wh_message_crypto.h
padelsbach: Switched to int

src/wh_server_crypto.c
padelsbach: initialized to 0

src/wh_client_crypto.c
Frauschi: In `wh_Client_SlhDsaImportKey()` and the DMA equivalent, the CommBuffer still has the private key DER without a ForceZero.

src/wh_server_crypto.c
```diff
@@ -5581,46 +5813,95 @@ static int _HandleMlKemKeyGen(whServerContext* ctx, uint16_t magic, int devId,
...
+    flags  = req.flags;
+    label  = req.label;
+    seedSz = req.seedSz;
+    seed   = (byte*)cryptoDataIn +
+           sizeof(whMessageCrypto_SlhDsaKeyGenRequest);
 
-    if (!_IsMlKemLevelSupported((int)req.level)) {
+    if (seedSz > (uint32_t)(inSize -
+                            sizeof(whMessageCrypto_SlhDsaKeyGenRequest))) {
         return WH_ERROR_BADARGS;
```
Frauschi, line 5857: The new `wc_ForceZero(seed, seedSz)` covers the unsupported-param and `wc_SlhDsaKey_Init` failures as well as the post-keygen path, but this bounds rejection returns above all of it, so a rejected request still leaves the client's seed bytes in the comm buffer.

`seedSz` is the value being rejected here, so it cannot be the length to scrub - it would run past the request. `inSize - sizeof(whMessageCrypto_SlhDsaKeyGenRequest)` is what is actually there.

src/wh_client_crypto.c
```diff
@@ -11236,6 +11241,1501 @@ int wh_Client_MlDsaCheckPrivKeyDma(whClientContext* ctx, wc_MlDsaKey* key,
...
+    if (ret >= 0) {
+        /* Stamp the cached keyId and the client's HSM devId so pub is
+         * immediately usable as a handle to the cached private key. The keyId
+         * is set here as well because the public-key deserialize re-inits
+         * pub and clears it. */
+        wh_Client_SlhDsaSetKeyId(pub, *inout_key_id);
+        pub->devId = WH_CLIENT_DEVID(ctx);
+    }
+    else if (committed && !WH_KEYID_ISERASED(*inout_key_id)) {
+        /* The server committed a key but the best-effort export returned no
```
Frauschi: This block and its twin in `wh_Client_SlhDsaMakeCacheKeyDma` (line 12337) both run to eight lines, and most of it is arguing the alternative that was not taken. That shape has been getting marked as slop in review lately, and I only asked for one line on the DMA side.

One sentence naming the constraint is enough, e.g. `/* Gated on the commit latch: the response-frame check also reports ABORTED, before any key id has been read. */` - the reasoning behind it reads better in the commit message, where it will not go stale.

Frauschi (commented):
Two small follow-ups from the last round - neither is blocking, and the rest of what I raised looks good now.

src/wh_server_crypto.c
padelsbach: Added

src/wh_client_crypto.c
padelsbach: Cleaned up

Frauschi (approved):

PR #518, opened 2026-08-28
fenrir fixes

bigbrett:
- **F-11419**: SHE `LOAD_KEY` accepted any slot as the authorizing key (e.g. a client-loaded RAM key could overwrite `MASTER_ECU_KEY`). Changed to now enforce the AUTOSAR SHE update-authorization matrix (Table 4.5), documented in `docs/src/5-Features.md`.
- **F-10226**: SHA-2 client wrappers left hash state mutated when the server returned an error, so wolfCrypt's software fallback silently produced a wrong digest. State is now snapshotted on entry and restored on any failure (mirrors the existing SHA-3 handling). Note that this behavior is dumb HOWEVER is required for correctness as long as silent software fallback exists.
- **F-11394**: a failed NVM write during key revocation left the cache marked revoked+committed, so a retry returned success while the NVM record stayed unrevoked. Revocation is now persisted from a temp copy and the cache mutated only on NVM success.
- **F-11393**: `wh_Server_Init` dropped the DMA `memCopyCb` supplied through `whServerConfig`. Fixed, it is now installed (under `WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY`).
- **F-10219**: the SHE PRNG seed's auto-cached copy was not evicted after persisting a new seed, so `_ExtendSeed` re-read a stale seed. Fixed so the cache entry is now evicted after the NVM write.

### DIFF
--- a/src/wh_server.c
+++ b/src/wh_server.c
@@ -156,6 +156,9 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
     if (NULL != config->dmaConfig) {
         server->dma.dmaAddrAllowList = config->dmaConfig->dmaAddrAllowList;
         server->dma.cb               = config->dmaConfig->cb;
+#ifdef WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY
+        server->dma.memCopyCb        = config->dmaConfig->memCopyCb;
+#endif /* WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY */
     }
 #endif /* WOLFHSM_CFG_DMA */
 
--- a/src/wh_server_dma.c
+++ b/src/wh_server_dma.c
@@ -190,8 +190,9 @@ int whServerDma_CopyToClient(struct whServerContext_t* server,
     /* Perform the actual copy */
 #ifdef WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY
     if (server->dma.memCopyCb != NULL) {
-        rc = server->dma.memCopyCb(server, clientAddr, (uintptr_t)serverPtr,
-                                   len, WH_DMA_COPY_OPER_CLIENT_WRITE, flags);
+        rc = server->dma.memCopyCb(server, (uintptr_t)transformedAddr,
+                                   (uintptr_t)serverPtr, len,
+                                   WH_DMA_COPY_OPER_CLIENT_WRITE, flags);
         if (rc != WH_ERROR_OK) {
             return rc;
         }
--- a/src/wh_server_keystore.c
+++ b/src/wh_server_keystore.c
@@ -1407,15 +1407,24 @@ int wh_Server_KeystoreRevokeKey(whServerContext* server, whNvmId keyId)
         return WH_ERROR_OK;
     }
 
-    /* Revoke the key by updating its metadata */
-    _revokeKey(cacheMeta);
-    /* commit the changes */
     if (isInNvm) {
-        ret = wh_Nvm_AddObjectWithReclaim(server->nvm, cacheMeta,
-                                          cacheMeta->len, cacheBuf);
+        /* Persist the revocation from a temporary copy so a failed NVM write
+         * leaves the cached metadata and committed state untouched. Mutating
+         * the cache first would leave it revoked and committed while NVM still
+         * holds a usable key, and a retry would early-return success above
+         * without ever re-persisting. */
+        whNvmMetadata revokedMeta = *cacheMeta;
+        _revokeKey(&revokedMeta);
+        ret = wh_Nvm_AddObjectWithReclaim(server->nvm, &revokedMeta,
+                                          revokedMeta.len, cacheBuf);
         if (ret == WH_ERROR_OK) {
+            _revokeKey(cacheMeta);
             _MarkKeyCommitted(_GetCacheContext(server, keyId), keyId, 1);
         }
+    }
+    else {
+        /* No NVM copy: the cache is the only store, so revoke it directly. */
+        _revokeKey(cacheMeta);
     }
 
     return ret;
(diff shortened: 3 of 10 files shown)


test/wh_test_she.c
```diff
@@ -452,6 +452,77 @@ int whTest_SheClientConfig(whClientConfig* config)
...
+        }
+        if ((ret = wh_Client_SheLoadKey(client, messageOne, messageTwo,
+                                        messageThree, outMessageFour,
+                                        outMessageFive)) != 0) {
+            WH_ERROR_PRINT("SHE LOAD KEY BOOT_MAC_KEY-authorized BOOT_MAC "
+                           "update: expected success, got %d\n",
+                           ret);
+            goto exit;
+        }
+        WH_TEST_PRINT("SHE LOAD KEY authorization matrix SUCCESS\n");
```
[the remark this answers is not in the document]
bigbrett: fixd

src/wh_server.c
```diff
@@ -156,6 +156,9 @@ int wh_Server_Init(whServerContext* server, whServerConfig* config)
     if (NULL != config->dmaConfig) {
         server->dma.dmaAddrAllowList = config->dmaConfig->dmaAddrAllowList;
         server->dma.cb               = config->dmaConfig->cb;
+#ifdef WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY
+        server->dma.memCopyCb        = config->dmaConfig->memCopyCb;
+#endif /* WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY */
```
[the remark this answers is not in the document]
bigbrett: fixed, good catch

src/wh_server_dma.c
```diff
@@ -190,8 +190,9 @@ int whServerDma_CopyToClient(struct whServerContext_t* server,
     /* Perform the actual copy */
 #ifdef WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY
     if (server->dma.memCopyCb != NULL) {
-        rc = server->dma.memCopyCb(server, clientAddr, (uintptr_t)serverPtr,
-                                   len, WH_DMA_COPY_OPER_CLIENT_WRITE, flags);
+        rc = server->dma.memCopyCb(server, (uintptr_t)transformedAddr,
+                                   (uintptr_t)serverPtr, len,
+                                   WH_DMA_COPY_OPER_CLIENT_WRITE, flags);
```
bigbrett, line 195: @rizlik I think your port is the only one that uses these - can you ensure this is the correct fix? It should be since memcpying from the untransformed client address definitely seemed wrong, but want to make sure I don't break your port

rizlik: This was indeed odd and was handled inside the port. I'll update the port to reflect the new contract. Thanks for the heads up.

src/wh_server_keystore.c
```diff
@@ -1407,16 +1407,21 @@ int wh_Server_KeystoreRevokeKey(whServerContext* server, whNvmId keyId)
...
     if (isInNvm) {
-        ret = wh_Nvm_AddObjectWithReclaim(server->nvm, cacheMeta,
-                                          cacheMeta->len, cacheBuf);
+        /* Persist revocation before updating cache to keep state consistent
+         * on failure. */
+        whNvmMetadata revokedMeta = *cacheMeta;
+        _revokeKey(&revokedMeta);
+        ret = wh_Nvm_AddObjectWithReclaim(server->nvm, &revokedMeta,
+                                          revokedMeta.len, cacheBuf);
         if (ret == WH_ERROR_OK) {
```
padelsbach, line 1417: If the NVM write fails, looks like the revoked copy is dropped and the entry lives on unrevoked. Is that what you're going for?

Alternatively, should we mark the entry as uncommitted/dirty and then do the write?

src/wh_server_dma.c
```diff
@@ -190,8 +190,9 @@ int whServerDma_CopyToClient(struct whServerContext_t* server,
     /* Perform the actual copy */
 #ifdef WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY
```
padelsbach, line 191: Should `WOLFHSM_CFG_DMA_CUSTOM_CLIENT_COPY` exist in a test case, or even a Makefile?

src/wh_server_she.c
```diff
@@ -256,6 +258,65 @@ static uint32_t _PopFlags(uint8_t* messageTwo)
...
+{
+    return (sheSlot > WH_SHE_BOOT_MAC) && (sheSlot < WH_SHE_RAM_KEY_ID);
+}
+
+/* Check that the authorizing slot may update the target slot, before any key
+ * material is read. This enforces the SHE memory update policy (AUTOSAR SHE
+ * Table 4.5), with one documented wolfHSM extension: SECRET_KEY, the ROM root
+ * of trust, may authorize any load so it can provision MASTER_ECU_KEY.
+ * Returns 0 when allowed, WH_SHE_ERC_KEY_INVALID otherwise. */
+static int _CheckLoadKeyAuth(uint16_t targetId, uint16_t authId)
```
padelsbach, line 273: According to a quick coverage run, the two SHE fixes here aren't covered with the new tests. Could we add those?

padelsbach (changes requested):

src/wh_server_she.c
bigbrett: added more coverage

src/wh_server_dma.c
bigbrett: maybe, but out of scope for now. Not sure what that would even entail as it is really just pinging callbacks. Not really a priority especially since it is an esoteric feature only used on one hardware platform to work around some crappy silicon

src/wh_server_keystore.c
bigbrett: That was my intent. I suppose the other way is fine too, maybe even preferable, but its kind of bikeshedding at this point. IMO both are okay since at the end of the day if NVM fails, the system is essentially not working and the next POR would return to the same state anyway

padelsbach (approved):

PR #522, opened 2026-09-04
ECC wolfBoot image manager verify methods

bigbrett:
Adds support for wolfBoot images signed with ECC256+SHA256 to support the default verify method on tc4.

Note: AI review agents might complain about returning `ABORTED` vs `NOTVERIFIED` as well as which wolfCrypt error codes are squashed and which are propagated through to the caller. IMO the current implementation is fine for ECC and I think unifying the various algorithms is scope for a future PR.


test-refactor/server/wh_test_server_img_mgr.c
```diff
@@ -508,6 +508,63 @@ static int _whTest_ServerImgMgrRsa2048(whServerContext* server)
     return WH_TEST_SUCCESS;
 }
 
+#if !defined(NO_RSA) || defined(HAVE_ECC)
```
bigbrett: fixed

src/wh_server_img_mgr.c
```diff
@@ -1020,8 +1022,205 @@ int wh_Server_ImgMgrVerifyMethodWolfBootRsa4096WithSha256(
...
+    }
+    if (ret == 0) {
+        ret = wc_ecc_verify_hash(derSig, derSigSz, hash, hashSz, &verifyResult,
+                                 &eccKey);
+    }
+
+    (void)wc_ecc_free(&eccKey);
+
+    if (ret != 0) {
+        return WH_ERROR_ABORTED;
```
bigbrett: resolved

bigbrett: resolved

src/wh_server_img_mgr.c
```diff
@@ -1020,8 +1022,214 @@ int wh_Server_ImgMgrVerifyMethodWolfBootRsa4096WithSha256(
...
+        return WH_ERROR_NOTVERIFIED;
+    }
+
+    ret = wc_ecc_init_ex(&eccKey, NULL, devId);
+    if (ret != 0) {
+        return WH_ERROR_ABORTED;
+    }
+
+    ret = wc_ecc_import_unsigned(&eccKey, pubkey, pubkey + 32, NULL,
+                                 ECC_SECP256R1);
```
padelsbach, line 1089: Why hard code the curve here? Should we at least check the DER for this curve

test-refactor/server/wh_test_server_img_mgr.c
```diff
@@ -632,6 +739,101 @@ static int _whTest_ServerImgMgrWolfBootCertChainRsa4096(whServerContext* server)
...
+    memcpy(corrupt_fw, wolfboot_test_firmware, sizeof(corrupt_fw));
+    corrupt_fw[0] ^= 0x01;
+    altImage.addr = (uintptr_t)corrupt_fw;
+    rc            = wh_Server_ImgMgrVerifyImg(&imgMgr, &altImage, &result);
+    WH_TEST_ASSERT_RETURN(rc != WH_ERROR_OK);
+    WH_TEST_ASSERT_RETURN(rc == result.verifyMethodResult);
+
+    /* Negative: a header signed with RSA4096 is rejected */
+    altImage         = testImage;
+    altImage.hdrAddr = (uintptr_t)wolfboot_test_rsa4096_header;
```
padelsbach, line 814: Could we add the inverse test case: ecc header with rsa verifyMethod?

Frauschi (approved):
LGTM besides the open comments from @padelsbach.

test-refactor/server/wh_test_server_img_mgr.c
bigbrett: I suppose. Kind of is a smoke test though, not sure each algo needs to go through every permutation of mismatches for something the user manually registered (e.g. user error) that would fail later on in runtime due to a sig failure anyway?

src/wh_server_img_mgr.c
bigbrett: wolfBoot uses raw keys so figured it had to hardcode. I think I perhaps could glean that info from the wolfBoot header signature algo field and tie variable length fields to that but could be odd to do statically. Let me think on it for a min

bigbrett: @padelsbach yeah the raw key path has to assume the curve. And wolfBoot already has static dispatch for everything vs dynamic so think we should just keep that paradigm here. Recommend no changes.

test-refactor/server/wh_test_server_img_mgr.c
bigbrett: So after looking at it a bit more I don't think that really adds any meaningful coverage, as it goes through the same failure path since the image falls through the same core verification helper logic. No permutation of this would exercise any new code paths. IMO this wouldn't really add or test for anything new

PR #524, opened 2026-09-18
Support 128-bit flash programming units

aidangarske:
Add configurable 8-byte and 16-byte flash programming units. Keeps the 8 byte as default. STM32H563 can only program internal flash in 128-bit quadwords this is needed wolfTrust integration


wolfhsm/wh_flash_unit.h
```diff
@@ -40,8 +40,28 @@
 
 #include "wolfhsm/wh_flash.h"
 
-/* Smallest programmable unit/size.  Alignment as well */
+/* Smallest programmable unit/size. Alignment as well. Most targets program
+ * 64-bit units. Targets that require 128-bit flash writes may select
+ * WOLFHSM_CFG_FLASH_UNIT_SIZE=16. */
+#ifndef WOLFHSM_CFG_FLASH_UNIT_SIZE
+    #define WOLFHSM_CFG_FLASH_UNIT_SIZE 8
```
padelsbach, line 47: It's not immediately clear if this is bits or bytes. Can you document the units here?

wolfhsm/wh_flash_unit.h
```diff
@@ -40,8 +40,28 @@
...
+#if WOLFHSM_CFG_FLASH_UNIT_SIZE == 8
 typedef uint64_t whFlashUnit;
+    #define WHFU_VALUE(_value) ((whFlashUnit)(_value))
+    #define WHFU_TO_U64(_unit) ((uint64_t)(_unit))
+#elif WOLFHSM_CFG_FLASH_UNIT_SIZE == 16
+typedef struct whFlashUnit_t {
+    uint64_t value;
+    uint64_t padding;
+} whFlashUnit;
+    #define WHFU_VALUE(_value) \
```
padelsbach: Naming of these macros seems inconsistent for such similar logic. When to use `WHFU_TO_` vs `WHFU_`? 

Note there is also `WHFU_BYTES2UNITS` below, which has a different convention.

src/wh_nvm_flash.c
```diff
@@ -1028,14 +1030,30 @@ int wh_NvmFlash_Init(void* c, const void* cf)
...
-            context->partition_units =
-                    context->cb->PartitionSize(context->flash) /
-                    WHFU_BYTES_PER_UNIT;
+        if (context->cb->PartitionSize == NULL) {
+            ret = WH_ERROR_BADARGS;
+        }
+        else {
+            partition_size = context->cb->PartitionSize(context->flash);
+            if (((partition_size % WHFU_BYTES_PER_UNIT) != 0) ||
+                ((partition_size / WHFU_BYTES_PER_UNIT) <
```
padelsbach, line 1075: Should we use `<=` here to allow the full sized partition?

src/wh_nvm_flash.c
```diff
@@ -1051,22 +1069,25 @@ int wh_NvmFlash_Init(void* c, const void* cf)
...
-                    (part_states[1].status == NF_STATUS_USED)) {
+        }
+        else if ((part_states[0].status != NF_STATUS_USED) &&
+                 (part_states[1].status == NF_STATUS_USED)) {
             context->active = 1;
             context->state = part_states[context->active];
-        } else if ( (part_states[0].status == NF_STATUS_USED) &&
-                    (part_states[1].status == NF_STATUS_USED)) {
+        }
+        else if ((part_states[0].status == NF_STATUS_USED) &&
```
padelsbach, line 1111: The both-UNUSED case is not exercised with the new tests. Can you add it? 

Also, the preexisting both corrupted case is not covered if you want to add that too

test/wh_test_nvm_flash.c
```diff
@@ -47,8 +47,8 @@
 #endif
 
 #define FLASH_RAM_SIZE (1024 * 1024) /* 1MB */
-#define FLASH_SECTOR_SIZE (4096) /* 4KB */
-#define FLASH_PAGE_SIZE (8) /* 8B */
+#define FLASH_SECTOR_SIZE (4096)     /* 4KB */
```
padelsbach, line 50: Please add the same test cases to `test-refactor/` too

src/wh_nvm_flash.c
```diff
@@ -1028,14 +1030,30 @@ int wh_NvmFlash_Init(void* c, const void* cf)
...
+                ((partition_size / WHFU_BYTES_PER_UNIT) <
+                 NF_PARTITION_DATA_OFFSET)) {
+                ret = WH_ERROR_BADARGS;
+            }
+            else {
+                context->partition_units = partition_size / WHFU_BYTES_PER_UNIT;
+            }
+        }
+
+        if (ret != WH_ERROR_OK) {
```
padelsbach, line 1084: Nice addition of this cleanup. But it seems to be missing from the failure paths below too. Could you move this to the bottom of the function to handle them all?

And then in the tests, add calls to `wh_NvmFlash_Cleanup` on cases where this function returns error.

padelsbach (changes requested):

wolfhsm/wh_flash_unit.h
```diff
@@ -40,8 +40,28 @@
...
+#ifndef WOLFHSM_CFG_FLASH_UNIT_SIZE
+    #define WOLFHSM_CFG_FLASH_UNIT_SIZE 8
+#endif
+
+#if WOLFHSM_CFG_FLASH_UNIT_SIZE == 8
 typedef uint64_t whFlashUnit;
+    #define WHFU_VALUE(_value) ((whFlashUnit)(_value))
+    #define WHFU_TO_U64(_unit) ((uint64_t)(_unit))
+#elif WOLFHSM_CFG_FLASH_UNIT_SIZE == 16
+typedef struct whFlashUnit_t {
```
billphipps: Grumble.  I had hoped that we would change whFlashUnit into an array of uint32_t's so that the accessors could rely on at least 32-bit alignment.  Initially, I was lazy and avoided adding any alignment attributes to this type by trusting the compiler would always ensure 64-bit was aligned on 64-bit boundaries.  At this point, I think we need to have a positive design that allows flash unit sizes to be variable and ensure that accesses to at least the first 2 32-bit values is easy.

Since the minimum size we want to support is 64-bit, let's construct this struct as a union instead.  Consider something like:

```
/* Flash unit size MUST be a power of 2 */
/* Flash units are aligned to 64-bit */
#ifndef WOLFHSM_CFG_FLASH_UNIT_SIZE
#define WOLFHSM_CFG_FLASH_UNIT_SIZE 8
#endif

#define DIVV_ROUND_UP(v,d) (((v) + (d) - 1) / (d))
#define U64_PER_UNIT DIV_ROUND_UP(WOLFHSM_CFG_FLASH_UNIT_SIZE, 8)
#define U32_PER_UNIT DIV_ROUND_UP(WOLFHSM_CFG_FLASH_UNIT_SIZE, 4)
#define U16_PER_UNIT DIV_ROUND_UP(WOLFHSM_CFG_FLASH_UNIT_SIZE, 2)

typedef union {
    uint64_t u64[U64_PER_UNIT];
    uint32_t u32[U32_PER_UNIT];
    uint16_t u16[U16_PER_UNIT];
} whFlashUnit_t;

static_assert(sizeof(whFlashUnit_t) >= WOLFHSM_CFG_FLASH_UNIT_SIZE);
WHFU_TO_U64(_u) ((_u).u64[0])

/* NVM State accessors now look like */
WHFU_STATE_MAGIC(_u) ((_u).u32[0]);
WHFU_STATE_EPOCH(_u) ((_u).u32[1]);
WHFU_STATE_CRC16(_u) ((_u).u16[1]);
```

This does NOT control the variable alignment any better than u64.  We'll need either some non-portable attribute code or move to C11 where alignas is supported.  Yuck.

billphipps (changes requested):
Consider an alternate construction?

wolfhsm/wh_flash_unit.h
aidangarske: Okay this makes sense I reworked it around the union you suggested with the `u64`, `u32`, and `u16` views and ^2 unit sizes of at least 8 bytes. I kept C90 by adding small compiler specific 8-byte alignment mappings instead of moving to C11.  The endian-specific state indexes preserve the existing on flash layout. I tested the 8-byte and 16-byte configurations, then pinned it into wolfTrust and passed the full host suite and m33mu testing

wolfhsm/wh_flash_unit.h
aidangarske: OKay I fixed this the header and configuration documentation now state that `WOLFHSM_CFG_FLASH_UNIT_SIZE` is measured in bytes.

wolfhsm/wh_flash_unit.h
aidangarske: Ive reworked this with Bills idea as a union with typed `u64`, `u32`, and `u16` views and removed both scalar conversion macros. `WHFU_BYTES2UNITS` remains the byte-count-to-unit-count conversion.

src/wh_nvm_flash.c
aidangarske: I think the current `<` comparison already allows equality. Changing it to `<=` would reject a partition exactly `NF_PARTITION_DATA_OFFSET` units long, so I left it unchanged.

src/wh_nvm_flash.c
aidangarske: Makes sense, added explicit coverage for both free/erased partitions and both corrupted partitions in the legacy and refactored suites, at both 8-byte and 16-byte unit sizes.

test/wh_test_nvm_flash.c
aidangarske: Added matching invalid-geometry and initialization-state cases to `test-refactor/`; the full suite passes at both 8-byte and 16-byte unit sizes.

src/wh_nvm_flash.c
aidangarske: Thanks, I moved failed initialization cleanup to the common exit path and added cleanup-count checks after each error case in both test suites.

PR #525, opened 2026-09-18
SHA-3: zeroize working buffers

padelsbach:
F-8510: wipe the intermediate buffers in SHA3 client side.

Found during SHAKE dev in PR #523


AlexLanzano:
Are there any performance implications from this change? Im not entirely convinced that the client needs to zeroize these buffers.

padelsbach:
> Are there any performance implications from this change? Im not entirely convinced that the client needs to zeroize these buffers.

According to a few benchmark runs, performance impact is < 0.5%, with all the other copying going on.

However, there's not much of a precedent of zeroing hash contents. WolfCrypt `wc_Sha3Free` doesn't clear it, so maybe the AI findings are getting overly picky. I guess we can drop this unless you see otherwise @AlexLanzano

